PR #17, opened 2022-08-30
Load firmware from the module loader

fabiendupont:
It is common for kernel module to come with firmware(s). On immutable operating systems like CoreOS, the standard firmware path `/lib/firmware` is read-only. So, we need to let the kernel know where to lookup firmware file(s).

This change adds the `FirmwarePath` attribute to the `ModprobeSpec` definition and modify the `MakeLoadCommand` and `MakeUnloadCommand` methods to leverage it.

It also adds a `Volume` and a `VolumeMount` to expose the `/var/lib/firmware/<Module.Name>` host directory (created if it doesn't exist) inside the pod. The `/var/lib/firmware` is [the recommended path for firmwares in OpenShift with RHCOS
nodes](https://access.redhat.com/documentation/fr-fr/openshift_container_platform/4.10/html-single/post-installation_configuration/index#rhcos-load-firmware-blobs_post-install-machine-configuration-tasks). Putting the firmware in a `Module.Name` subfolder avoids multiple
Modules to override each others firmware. It also simplifies the removal of the firmware files.

The `MakeLoadCommand` adds a step to copy the `FirmwarePath` to `/var/lib/firmware/<Module.Name>` to the `modprobe` command. The `MakeUnloadCommand` removes the firmware folder previously copied after unloading the module. The commands use `/usr/bin -c` to allow shell control like `&&`.

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
@@ -24,6 +25,8 @@ const (
 	nodeLibModulesVolumeName       = "node-lib-modules"
 	nodeUsrLibModulesPath          = "/usr/lib/modules"
 	nodeUsrLibModulesVolumeName    = "node-usr-lib-modules"
+	nodeVarLibFirmwarePath         = "/var/lib/firmware"
+	nodeVarLibFirmwareVolumeName   = "node-var-lib-firmware"
 	devicePluginKernelVersion      = ""
 )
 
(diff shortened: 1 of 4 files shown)


internal/daemonset/daemonset.go
```diff
@@ -340,42 +364,69 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+	loadCommandShell := []string{
+		"/bin/sh",
+		"-c",
+	}
+
+	loadCommand := "modprobe"
 
 	if ra := spec.RawArgs; ra != nil && len(ra.Load) > 0 {
-		return append(loadCommand, ra.Load...)
+		loadCommand = fmt.Sprintf("%s %s", loadCommand, strings.Join(ra.Load[:], " "))
```
qbarrand, line 375: Why is `[:]` useful after `ra.Load` here?

internal/daemonset/daemonset.go
```diff
@@ -340,42 +364,69 @@ func OverrideLabels(labels, overrides map[string]string) map[string]string {
...
+		return append(loadCommandShell, loadCommand)
+	}
+
+	if fw := spec.FirmwarePath; fw != "" {
+		loadCommand = fmt.Sprintf("cp -r %s %s && %s", fw, nodeVarLibFirmwarePath, loadCommand)
 	}
 
 	if a := spec.Args; a != nil && len(a.Load) > 0 {
-		loadCommand = append(loadCommand, a.Load...)
+		loadCommand = fmt.Sprintf("%s %s", loadCommand, strings.Join(a.Load[:], " "))
```
qbarrand, line 384: I think using a `strings.Builder` would more efficient in `MakeLoadCommand` and `MakeUnloadCommand`.

internal/daemonset/daemonset.go
```diff
@@ -195,6 +200,25 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
...
+				},
+			},
+		}
+		ds.Spec.Template.Spec.Volumes = append(ds.Spec.Template.Spec.Volumes, firmwareVolume)
+
+		firmwareVolumeMount := v1.VolumeMount{
+			Name:      nodeVarLibFirmwareVolumeName,
+			MountPath: nodeVarLibFirmwarePath,
+		}
+		ds.Spec.Template.Spec.Containers[0].VolumeMounts = append(ds.Spec.Template.Spec.Containers[0].VolumeMounts, firmwareVolumeMount)
```
qbarrand, line 219: I think it would probably be safer to have `ds.Spec.Template.Spec.Containers[0]` as a variable, to add the volume mount to it and to ultimately add it to the `TemplateSpec`.

internal/daemonset/daemonset.go
```diff
@@ -195,6 +200,25 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
 		Selector: &metav1.LabelSelector{MatchLabels: standardLabels},
 	}
 
+	if fw := mod.Spec.ModuleLoader.Container.Modprobe.FirmwarePath; fw != "" {
```
qbarrand, line 203: I don't think that new `if` is covered by a unit test.

internal/daemonset/daemonset.go
```diff
@@ -195,6 +200,25 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
...
+		firmwareVolume := v1.Volume{
+			Name: nodeVarLibFirmwareVolumeName,
+			VolumeSource: v1.VolumeSource{
+				HostPath: &v1.HostPathVolumeSource{
+					Path: nodeVarLibFirmwarePath,
```
yevgeny-shnaidman, line 208: do we want to create a path per Module? something like: /var/lib/firnware/<module-name>?
This way we avoid collision between different firmwares with the same name, and also hide the directory from user

qbarrand: If I understand [the docs](https://www.kernel.org/doc/html/v5.19/driver-api/firmware/fw_search_path.html) correctly, it's global.

internal/daemonset/daemonset.go
fabiendupont: Not sure what it would look like. Could you provide an example, please?

internal/daemonset/daemonset.go
fabiendupont: I've also split the volumes to follow the same pattern. Is that what you had in mind?

internal/daemonset/daemonset.go
fabiendupont: Good catch. Test added.

internal/daemonset/daemonset.go
fabiendupont: We have the same understanding. I tried to path a `fw_path` parameter to the module, but it's not recognized. Everytime we add a search path to the kernel, the node is rebooted, so it's probably not what we want.

internal/daemonset/daemonset.go
qbarrand: Yes, although I would have put that work above the assigment of `ds.Spec` line 124.

internal/daemonset/daemonset.go
qbarrand: > Everytime we add a search path to the kernel, the node is rebooted, so it's probably not what we want.

I guess writing to `/sys/module/firmware_class/parameters/path` doesn't reboot the node though, does it?

internal/daemonset/daemonset.go
fabiendupont: Done.

internal/daemonset/daemonset.go
qbarrand: Opened #44 to follow-up on this.

PR #20, opened 2022-08-30
Adding a job that enforce a single commit for each PR.

ybettan:
Adding a job that enforce a single commit for each PR.

This job will make sure that all commits in a PR were squashed to a single
commit.

### DIFF
--- a/ci/prow/check-commits-count
+++ b/ci/prow/check-commits-count
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+number_of_commits=$(git rev-list --count HEAD ^main)
+if [[ ${number_of_commits} != 1 ]]; then
+    echo '
+    All PRs must contain a single commit.
+    Please refer to https://github.com/kubernetes-sigs/kernel-module-management/blob/main/CONTRIBUTING.md
+    '
+    exit 1
+fi


yevgeny-shnaidman:
@ybettan why do we need the PR description and commit message to be identical? why is it a hard requirement?

ybettan:
@yevgeny-shnaidman I think that the commit message should be very descriptive and should include all the info required for understanding the code change.

By enforcing identical commit message and PR description, the reviewer of the PR can just look at the PR description without the need of also checking the commit message inside the PR in order to make sure the commit message wasn't neglected.

In addition, you can add additional info such as reviewers or any other text that is not relevant for describing the change by adding `---` at the end of the PR description and all the text after that will be ignored by the API and doesn't have to match the commit message.

qbarrand:
/hold
Hold while we compare those requirements with what the [Kubernetes contributor cheat sheet](https://github.com/kubernetes/community/tree/master/contributors/guide/contributor-cheatsheet) recommends.

ybettan:
As agreed, this PR is going to be split into 2 PRs:
* One PR to add a job to enforce a single commit per PR
* One PR to enforce the sync between the PR description and the commit message

Both PR will start as optional jobs and the second one might be removed after some time if it causes more issues than benefits.

ci/prow/check-commits-number
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+#FIXME: check if there is a prow github token that can be use to increase the api rate limit from 60 req/h to 5000 req/h
+
+number_of_commits=$(curl -H 'Accept: application/vnd.github+json' \
+    https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/pulls/${PULL_NUMBER} | jq '.commits')
+if [[ ${number_of_commits} != 1 ]]; then
```
qbarrand, line 9: I think we could use `git ` to do that: `git rev-list --count HEAD ^main`

ybettan: Good point.

Fixed.

ci/prow/check-commits-number
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
qbarrand, line 1: I'd rename this to `ci/prow/check-commits-count`. WDYT?

ybettan: Fixed.

PR #24, opened 2022-08-31
Adding build option to disable pushes to registry

yevgeny-shnaidman:
This PR adds additionl flag to the Build APIs
that allows to configure kaniko flag to either
push the image to registry(once it is build),
or skip the push  to registry

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -234,7 +234,7 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
 	logger := log.FromContext(ctx).WithValues("kernel version", kernelVersion, "image", km.ContainerImage)
 	buildCtx := log.IntoContext(ctx, logger)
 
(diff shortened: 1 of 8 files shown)


controllers/module_reconciler.go
```diff
@@ -234,7 +234,7 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
 	logger := log.FromContext(ctx).WithValues("kernel version", kernelVersion, "image", km.ContainerImage)
 	buildCtx := log.IntoContext(ctx, logger)
 
-	buildRes, err := r.buildAPI.Sync(buildCtx, *mod, *km, kernelVersion)
+	buildRes, err := r.buildAPI.Sync(buildCtx, *mod, *km, kernelVersion, true)
```
erusso7, line 237: If this value is always `true`, what's the point of allowing to push or not the image?
The image will always be pushed, unless I'm missing something else :thinking:

PR #28, opened 2022-09-01
Fixing ci/prow/check-commit-count.

ybettan:
Prow is cloning the repo using `git clone --mirror ...`.

In addition to that, the changes are directly tested in the `main`
branch, therefore, the only way to get the diff is to add the `upstream`
remote and check the diff against `upstream/main` instead of `main`.

### DIFF
--- a/ci/prow/check-commits-count
+++ b/ci/prow/check-commits-count
@@ -2,8 +2,13 @@
 
 set -euxo pipefail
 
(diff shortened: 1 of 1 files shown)


ybettan:
Here is another optional [fix](https://github.com/kubernetes-sigs/kernel-module-management/pull/34) which I think is better and lay the infra for writing more complex jobs like checking the commits content before merging.

PR #32, opened 2022-09-04
Separate PullOption for build process and image verification

yevgeny-shnaidman:
Currently we have pull option in the Build section of the Module API. It is used both
for configuring the pull option passed to the build configuration (kaniko)
and also to the function that verifies images existence (pulling its manifest).
We should separate them into 2 different entities:
1) present in ModuleLoader spec - used for image contents inspection and existence verification
2) present in the Build - used for specifying pull options of the base images of the build process


api/v1beta1/module_types.go
```diff
@@ -96,6 +96,11 @@ type KernelMapping struct {
 	// Literal defines a literal target kernel version to be matched exactly against node kernels.
 	Literal string `json:"literal"`
 
+	// +optional
+	// Pull contains settings determining how to check if the ModuleLoader image already exists
+	// and allows overriding of the Module's pull options
```
qbarrand, line 101:
```suggestion
	// and allows overriding of the ModuleLoader's pull options
```

api/v1beta1/module_types.go
```diff
@@ -63,7 +63,7 @@ type Build struct {
 	Dockerfile string `json:"dockerfile"`
 
 	// +optional
-	// Pull contains settings determining how to check if the DriverContainer image already exists.
+	// Pull contains settings determining how to pull the base images of the build process.
 	Pull PullOptions `json:"pull"`
```
qbarrand, line 67: Should that one be made a pointer?

api/v1beta1/module_types.go
yevgeny-shnaidman: fixed

api/v1beta1/module_types.go
yevgeny-shnaidman: I see no reason for it right now. The other 2 are pointers since they can be overriden ( global or in mapping), but the one in Build is not overriden, and the default values are false

PR #33, opened 2022-09-05
Separate in-cluster build and image existence verification

yevgeny-shnaidman:
This PR removes the image existance verification from
Sync API of the build package and moves it into the
handleBuild function of the module reconcilier(prior
to calling the Sync API)

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -21,11 +21,13 @@ import (
 	"fmt"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/auth"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/daemonset"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/metrics"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/statusupdater"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
@@ -52,6 +54,7 @@ type ModuleReconciler struct {
 	daemonAPI        daemonset.DaemonSetCreator
 	kernelAPI        module.KernelMapper
 	metricsAPI       metrics.Metrics
+	registry         registry.Registry
 	filter           *filter.Filter
 	statusUpdaterAPI statusupdater.ModuleStatusUpdater
 }
@@ -63,6 +66,7 @@ func NewModuleReconciler(
 	kernelAPI module.KernelMapper,
 	metricsAPI metrics.Metrics,
 	filter *filter.Filter,
+	registry registry.Registry,
 	statusUpdaterAPI statusupdater.ModuleStatusUpdater) *ModuleReconciler {
 	return &ModuleReconciler{
 		Client:           client,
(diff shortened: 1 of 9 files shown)


controllers/module_reconciler.go
```diff
@@ -229,6 +234,13 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
...
 	}
+	exists, err := r.checkImageExists(ctx, mod, km)
+	if err != nil {
+		return false, fmt.Errorf("failed to check image existence for kernel %s: %w", kernelVersion, err)
+	}
+	if exists {
+		return false, nil
+	}
 
 	// [TODO] check access to the image - execute build only if needed (image is inaccessible)
```
qbarrand, line 245: I think we can drop this comment.

main.go
```diff
@@ -123,14 +123,14 @@ func main() {
 	registryAPI := registry.NewRegistry()
 	helperAPI := build.NewHelper()
 	makerAPI := job.NewMaker(helperAPI, scheme)
-	buildAPI := job.NewBuildManager(client, registryAPI, makerAPI, helperAPI)
+	buildAPI := job.NewBuildManager(client /*registryAPI,*/, makerAPI, helperAPI)
```
qbarrand, line 126: I think we can remove the comment here.

controllers/module_reconciler_test.go
```diff
@@ -27,430 +29,582 @@ const (
 	namespace = "namespace"
 )
 
-var _ = Describe("ModuleReconciler", func() {
-	Describe("Reconcile", func() {
```
qbarrand, line 31: Didn't we agree to wrap all method-level `Describe` inside one `Describe` per type? cc @ybettan

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

main.go
yevgeny-shnaidman: fixed

controllers/module_reconciler_test.go
yevgeny-shnaidman: IIRC we agreed that we can use describe per function being tested. No need to use describe for module, since the file its defines which file/module we are testing.

ybettan: +1 for @yevgeny-shnaidman

controllers/module_reconciler_test.go
```diff
@@ -27,430 +29,582 @@ const (
...
-				clnt.EXPECT().List(ctx, gomock.Any(), gomock.Any()).DoAndReturn(
-					func(_ interface{}, list *v1.NodeList, _ ...interface{}) error {
-						list.Items = nodeList.Items
-						return nil
-					},
-				),
-				clnt.EXPECT().Get(ctx, gomock.Any(), gomock.Any()),
-				clnt.EXPECT().Patch(ctx, gomock.Any(), gomock.Any()),
-			)
+var _ = Describe("ModuleReconciler_HandleBuild", func() {
```
qbarrand, line 460:
```suggestion
var _ = Describe("ModuleReconciler_handleBuild", func() {
```

controllers/node_kernel_reconciler_test.go
```diff
@@ -14,96 +14,94 @@ import (
...
-		BeforeEach(func() {
-			gCtrl = gomock.NewController(GinkgoT())
-			clnt = client.NewMockClient(gCtrl)
-		})
-		const (
-			kernelVersion = "1.2.3"
-			labelName     = "label-name"
-			nodeName      = "node-name"
-		)
+var _ = Describe("Reconcile", func() {
```
qbarrand, line 17:
```suggestion
var _ = Describe("NodeKernelReconciler_Reconcile", func() {
```

yevgeny-shnaidman: fixed

controllers/module_reconciler_test.go
yevgeny-shnaidman: fixed

PR #39, opened 2022-09-06
Streamline GitHub Actions

qbarrand:
Move the build of the manager binary to the unit-tests Prow script.
Remove the linting of the code in favor of the lint Prow script.


ci/prow/lint
```diff
@@ -2,4 +2,13 @@
 
 set -euxo pipefail
 
+# Generate code and compare with the current source.
+# Fail with a diff if some files are outdated.
+make generate
+
+if ! git diff --exit-code; then
+    echo "Changes detected; diff:"
+    git diff
```
ybettan, line 11: I don't think we need to print `diff` as the `-x` of the script will do it for us.

ci/prow/unit-tests
```diff
@@ -3,3 +3,4 @@
 set -euxo pipefail
 
 make unit-test
+make manager
```
ybettan, line 6: Why is this required?

If it is only to test the the code is compiled correctly, then `unit-test` will fail (because `vet` will fail)

ybettan: Shouldn't it be part of the `lint` job in that case?

qbarrand: `lint` runs in a golangci-lint image, not sure it can build our code and with which go version.
Unit tests do not need to build the manager binary to run, so I wouldn't add it as a dependency.
Maybe we should have a separate `build` job then?

ybettan: > Maybe we should have a separate build job then?

I think so if you don't mind.

qbarrand:
/assign @yevgeny-shnaidman
This PR needs a `/lgtm`. Thanks!

PR #40, opened 2022-09-06
Remove /usr/lib references at DaemonSet

enriquebelarte:
In most cases shared libraries directory /lib is a dynamic link to /usr/lib but in some other cases as Ubuntu 18.04 LTS or Debian 11, those directories have completely different content. In any case in-tree modules can be found on /lib/modules/KERNEL_VERSION/ so in order to avoid compatibility issues depending which OS Kubernetes nodes are running we should stick to /lib as a mount point for DaemonSet.

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -261,14 +261,7 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
 }
 
 func (r *ModuleReconciler) checkImageExists(ctx context.Context, mod *kmmv1beta1.Module, km *kmmv1beta1.KernelMapping) (bool, error) {
-	var registryAuthGetter auth.RegistryAuthGetter
-	if irs := mod.Spec.ImageRepoSecret; irs != nil {
-		namespacedName := types.NamespacedName{
-			Name:      irs.Name,
-			Namespace: mod.Namespace,
-		}
-		registryAuthGetter = auth.NewRegistryAuthGetter(r.Client, namespacedName)
-	}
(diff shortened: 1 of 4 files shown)


yevgeny-shnaidman:
@enriquebelarte you need to fix unit-test

internal/daemonset/daemonset.go
```diff
@@ -22,8 +22,6 @@ const (
 	kubeletDevicePluginsPath       = "/var/lib/kubelet/device-plugins"
 	nodeLibModulesPath             = "/lib/modules"
 	nodeLibModulesVolumeName       = "node-lib-modules"
```
yevgeny-shnaidman, line 24: maybe we can restrict it even further to  /lib/modules/<KERNEL_VERSION> ?

ybettan:
@enriquebelarte Can you please make sure you inject all the PR description into the commit message?
Here is a guide for commit messages - https://github.com/kubernetes-sigs/kernel-module-management/blob/main/CONTRIBUTING.md

internal/daemonset/daemonset.go
```diff
@@ -22,8 +22,6 @@ const (
...
-	nodeUsrLibModulesPath          = "/usr/lib/modules"
```
ybettan, line 25: Doesn't it mean we just need to change it to `/lib/modules`?

internal/daemonset/daemonset.go
enriquebelarte: Done.

PR #42, opened 2022-09-06
Moving authenticator creation to auth package

yevgeny-shnaidman:


internal/registry/registry.go
```diff
@@ -40,24 +43,24 @@ type RepoPullConfig struct {
...
-	GetLayersDigests(ctx context.Context, image string, registryAuthGetter auth.RegistryAuthGetter) ([]string, *RepoPullConfig, error)
+	GetLayersDigests(ctx context.Context, image string, secret *corev1.LocalObjectReference, namespace string) ([]string, *RepoPullConfig, error)
 	GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (v1.Layer, error)
 }
 
-type registry struct{}
-
-func NewRegistry() Registry {
-	return &registry{}
+type registry struct {
```
ybettan, line 52: I know we agreed on that but I know see the implications of it - don't you think we should detach the `registry` interface from the k8s cluster itself. 

I am not a fan of making the registry API tied to the k8s client.

WDYT?

internal/registry/registry.go
```diff
@@ -40,24 +43,24 @@ type RepoPullConfig struct {
...
-	if err != nil {
-		return false, fmt.Errorf("failed to get pull options for image %s: %w", image, err)
+func NewRegistry(client client.Client) Registry {
+	return &registry{
+		client: client,
 	}
-	_, err = r.getImageManifest(ctx, image, pullConfig)
+}
+
+func (r *registry) ImageExists(ctx context.Context, image string, po *kmmv1beta1.PullOptions, secret *corev1.LocalObjectReference, namespace string) (bool, error) {
```
ybettan, line 62: That means that if we add another `RegistryAuthGetter` in the future that doesn't relies on a secret we will have to change that function instead of just passing another `RegistryAuthGetter` as an argument.

internal/registry/registry.go
```diff
@@ -247,3 +256,14 @@ func (r *registry) getImageDigestFromMultiImage(manifestListStream []byte) (stri
 	}
 	return "", fmt.Errorf("Failed to find manifest for architecture %s", arch)
 }
+
+func (r *registry) getRegistryAuth(secret *corev1.LocalObjectReference, namespace string) auth.RegistryAuthGetter {
```
ybettan, line 260: What will we do if we add another authenticator (or replace that one) with one that isn't based on a secret?

main.go
```diff
@@ -120,7 +120,7 @@ func main() {
 
 	metricsAPI := metrics.New()
 	metricsAPI.Register()
-	registryAPI := registry.NewRegistry()
+	registryAPI := registry.NewRegistry(client)
```
ybettan, line 123: I would really prefer not to wire those two.
Does it have an additional benefit rather than shorten the code? I also feel it is making the code less comprehensive.

controllers/module_reconciler.go
```diff
@@ -261,14 +261,7 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
...
 func (r *ModuleReconciler) checkImageExists(ctx context.Context, mod *kmmv1beta1.Module, km *kmmv1beta1.KernelMapping) (bool, error) {
-	var registryAuthGetter auth.RegistryAuthGetter
-	if irs := mod.Spec.ImageRepoSecret; irs != nil {
-		namespacedName := types.NamespacedName{
-			Name:      irs.Name,
-			Namespace: mod.Namespace,
-		}
-		registryAuthGetter = auth.NewRegistryAuthGetter(r.Client, namespacedName)
-	}
+	registryAuthGetter := auth.GetRegistryAuth(r.Client, mod)
```
ybettan, line 264: How about `NewRegistryAuthGetterFrom`? It is eventually a New/Construct function

internal/registry/registry.go
```diff
@@ -46,18 +46,15 @@ type Registry interface {
 	GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (v1.Layer, error)
 }
 
-type registry struct{}
+type registry struct {
+}
```
ybettan, line 50: Nit, we can close the diff here.

yevgeny-shnaidman: this change was done by fmt, so i would stick with it

controllers/module_reconciler.go

ybettan:
Powerless LGTM.

PR #52, opened 2022-09-08
Preflight should use PullOptions in image verification

yevgeny-shnaidman:
Since now we have pull options also in kernel mapping and in Spec, those parameters should now be used in the preflight image verification


ybettan:
Powerless LGTM.

PR #65, opened 2022-09-13
adding build verification to preflight

yevgeny-shnaidman:
In case image of the Module is not available or
is does not contain the correct driver version,
preflight will check if Build configuration exists in Module yaml, and if it does, then it will try to build it. In case build is successful, then the module is considered to be verified

### DIFF
--- a/api/v1beta1/preflightvalidation_types.go
+++ b/api/v1beta1/preflightvalidation_types.go
@@ -21,10 +21,12 @@ import (
 )
 
 const (
-	VerificationTrue       string = "True"
-	VerificationFalse      string = "False"
-	VerificationStageImage string = "Image"
-	VerificationStageBuild string = "Build"
+	VerificationTrue          string = "True"
+	VerificationFalse         string = "False"
+	VerificationStageImage    string = "Image"
+	VerificationStageBuild    string = "Build"
+	VerificationStageRequeued string = "Requeued"
+	VerificationStageDone     string = "Done"
 )
 
 // PreflightValidationSpec describes the desired state of the resource, such as the kernel version
@@ -53,7 +55,7 @@ type CRStatus struct {
 	// image (image existence verification), build(build process verification)
 	// +required
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Enum=Image;Build
+	// +kubebuilder:validation:Enum=Image;Build;Requeued;Done
 	VerificationStage string `json:"verificationStage"`
 
 	// LastTransitionTime is the last time the CR status transitioned from one status to another.
--- a/config/crd/bases/kmm.sigs.k8s.io_preflightvalidations.yaml
+++ b/config/crd/bases/kmm.sigs.k8s.io_preflightvalidations.yaml
@@ -72,6 +72,8 @@ spec:
                       enum:
                       - Image
                       - Build
+                      - Requeued
+                      - Done
                       type: string
                     verificationStatus:
                       description: 'Status of Module CR verification: true (verified),
--- a/internal/preflight/preflight.go
+++ b/internal/preflight/preflight.go
@@ -6,8 +6,11 @@ import (
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/auth"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/statusupdater"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 
 	ctrlruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -17,33 +20,37 @@ const (
 	VerificationStatusReasonBuildConfigPresent = "Verification successful, all driver-containers have paired BuildConfigs in the recipe"
 	VerificationStatusReasonNoDaemonSet        = "Verification successful, no driver-container present in the recipe"
 	VerificationStatusReasonUnknown            = "Verification has not started yet"
-	VerificationStatusReasonVerified           = "Verification successful, this Module would be verified again in this Preflight CR"
+	VerificationStatusReasonVerified           = "Verification successful (%s), this Module will not be verified again in this Preflight CR"
 )
 
-//go:generate mockgen -source=preflight.go -package=preflight -destination=mock_preflight_api.go
+//go:generate mockgen -source=preflight.go -package=preflight -destination=mock_preflight_api.go PreflightAPI, preflightHelperAPI
 
 type PreflightAPI interface {
-	PreflightUpgradeCheck(ctx context.Context, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
+	PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
 }
 
 func NewPreflightAPI(
 	client client.Client,
+	buildAPI build.Manager,
 	registryAPI registry.Registry,
+	statusUpdater statusupdater.PreflightStatusUpdater,
 	kernelAPI module.KernelMapper) PreflightAPI {
+	helper := newPreflightHelper(client, buildAPI, registryAPI)
 	return &preflight{
-		registryAPI: registryAPI,
-		kernelAPI:   kernelAPI,
-		client:      client,
+		kernelAPI:     kernelAPI,
+		statusUpdater: statusUpdater,
+		helper:        helper,
 	}
 }
 
 type preflight struct {
-	client      client.Client
-	registryAPI registry.Registry
-	kernelAPI   module.KernelMapper
+	kernelAPI     module.KernelMapper
+	statusUpdater statusupdater.PreflightStatusUpdater
+	helper        preflightHelperAPI
 }
 
-func (p *preflight) PreflightUpgradeCheck(ctx context.Context, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
+func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
+	log := ctrlruntime.LoggerFrom(ctx)
 	mapping, err := p.kernelAPI.FindMappingForKernel(mod.Spec.ModuleLoader.Container.KernelMappings, kernelVersion)
 	if err != nil {
 		return false, fmt.Sprintf("Failed to find kernel mapping in the module %s for kernel version %s", mod.Name, kernelVersion)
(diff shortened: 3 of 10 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management
-  newTag: main
+  newName: quay.io/yshnaidm/oot-operator
+  newTag: preflight
```
ybettan, line 16: Leftover?

api/v1beta1/preflightvalidation_types.go
```diff
@@ -21,10 +21,12 @@ import (
...
-	VerificationTrue       string = "True"
-	VerificationFalse      string = "False"
-	VerificationStageImage string = "Image"
-	VerificationStageBuild string = "Build"
+	VerificationTrue                      string = "True"
+	VerificationFalse                     string = "False"
+	VerificationStageImage                string = "Image"
+	VerificationStageBuild                string = "Build"
+	VerificationStageWaitingNextReconcile string = "WaitingNextReconcile"
+	VerificationStageFinished             string = "Finished"
```
ybettan, line 29:
```suggestion
	VerificationStageRequeued string = "Requeued"
	VerificationStageDone         string = "Done"
```

ybettan, line 29:
```suggestion
	VerificationStageRequeued string = "Requeued"
	VerificationStageDone             string = "Done"
```

api/v1beta1/preflightvalidation_types.go
```diff
@@ -53,7 +55,7 @@ type CRStatus struct {
 	// image (image existence verification), build(build process verification)
 	// +required
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Enum=Image;Build
+	// +kubebuilder:validation:Enum=Image;Build;WaitingNextReconcile;Finished
```
ybettan, line 58: Same

config/crd/bases/kmm.sigs.k8s.io_preflightvalidations.yaml
```diff
@@ -72,6 +72,8 @@ spec:
                       enum:
                       - Image
                       - Build
+                      - WaitingNextReconcile
+                      - Finished
```
ybettan, line 76: Same

internal/preflight/preflight.go
```diff
@@ -55,13 +62,47 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, mod *kmmv1beta1.M
 		return false, fmt.Sprintf("Failed to substitute template in kernel mapping in the module %s for kernel version %s", mod.Name, kernelVersion)
 	}
 
-	return p.verifyImage(ctx, mapping, mod, kernelVersion)
+	err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageImage)
+	if err != nil {
+		log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to image stage"), "module", mod.Name, "error", err)
+	}
+
+	res, msg := p.helper.verifyImage(ctx, mapping, mod, kernelVersion)
```
ybettan, line 70:
```suggestion
	exists, msg := p.helper.verifyImage(ctx, mapping, mod, kernelVersion)
```

internal/preflight/preflight.go
```diff
@@ -55,13 +62,47 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, mod *kmmv1beta1.M
...
+	if res || (mapping.Build == nil && mod.Spec.ModuleLoader.Container.Build == nil) {
+		return res, msg
+	}
```
ybettan, line 73: Took me like 5 min to understand that if so I will suggest:
```suggestion
	if exists, msg := p.helper.verifyImage(ctx, mapping, mod, kernelVersion); exists {
	        return true, msg
	 else if mapping.Build == nil && mod.Spec.ModuleLoader.Container.Build == nil {
		return false, msg
	}
```

internal/preflight/preflight.go
```diff
@@ -55,13 +62,47 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, mod *kmmv1beta1.M
...
+	if err != nil {
+		log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to build stage"), "module", mod.Name, "error", err)
+	}
+
+	return p.helper.verifyBuild(ctx, mapping, mod, kernelVersion)
+}
+
+type preflightHelperAPI interface {
+	verifyImage(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
+	verifyBuild(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
```
ybettan, line 85: See https://github.com/kubernetes-sigs/kernel-module-management/pull/65/files#r969757894

internal/registry/registry.go
```diff
@@ -213,6 +216,8 @@ func (r *registry) getHeaderStreamFromLayer(layer v1.Layer, headerName string) (
 
 	tr := tar.NewReader(gr)
 
+	log.Log.Info("YEV -in getHeaderStreamFromLayer", "fileName", headerName)
```
ybettan, line 219: What does `YEV` mean?

internal/registry/registry.go
```diff
@@ -222,6 +227,7 @@ func (r *registry) getHeaderStreamFromLayer(layer v1.Layer, headerName string) (
 
 			return nil, fmt.Errorf("failed to get next entry from targz: %w", err)
 		}
+		log.Log.Info("YEV -in loop getHeaderStreamFromLayer", "headerName", header.Name)
```
ybettan, line 230: Leftovers I guess.

config/manager/kustomization.yaml
yevgeny-shnaidman: yes, fixed

api/v1beta1/preflightvalidation_types.go
yevgeny-shnaidman: I always agree with everything you suggest :).Fixed

api/v1beta1/preflightvalidation_types.go
yevgeny-shnaidman: Fixed

config/crd/bases/kmm.sigs.k8s.io_preflightvalidations.yaml
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
```diff
@@ -17,33 +20,37 @@ const (
 	VerificationStatusReasonBuildConfigPresent = "Verification successful, all driver-containers have paired BuildConfigs in the recipe"
 	VerificationStatusReasonNoDaemonSet        = "Verification successful, no driver-container present in the recipe"
 	VerificationStatusReasonUnknown            = "Verification has not started yet"
-	VerificationStatusReasonVerified           = "Verification successful, this Module would be verified again in this Preflight CR"
+	VerificationStatusReasonVerified           = "Verification successful (%s), this Module would not be verified again in this Preflight CR"
```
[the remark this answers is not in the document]
yevgeny-shnaidman: upon success, it will contain the stage that allowed verification: Image, Build, or Build and Push etc'

internal/preflight/preflight.go
```diff
@@ -17,33 +20,37 @@ const (
...
+	PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
 }
 
 func NewPreflightAPI(
 	client client.Client,
+	buildAPI build.Manager,
 	registryAPI registry.Registry,
+	statusUpdater statusupdater.PreflightStatusUpdater,
 	kernelAPI module.KernelMapper) PreflightAPI {
+	helper := newPreflightHelper(client, buildAPI, registryAPI)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: i disagree. Originally those functions were private functions of the preflight, i moved them into private interface just for unit-tests. Registry should deal only with accessing the registry and the image contents, the logic of how to check and what to check should be in preflight. The same goes for verifyBuild, and in later PRs it will be epanded more with Push option and also sign verification

internal/preflight/preflight.go
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
yevgeny-shnaidman: answered above

internal/preflight/preflight.go
yevgeny-shnaidman: i am not sure that using if .. else if.. is more readable, what if i change the res to imageVerified?

internal/registry/registry.go
yevgeny-shnaidman: yes, thanks, fixed

internal/registry/registry.go
yevgeny-shnaidman: yes, thanks, fixed

internal/preflight/preflight.go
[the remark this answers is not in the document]
ybettan: > Registry should deal only with accessing the registry and the image contents, the logic of how to check and what to check should be in preflight.

I agree with it. In that case, how about a registry API function that gets a predicate function dictated by the preflight code? That way, you keep the registry logic in the registry package, and you extract the preflight part to be injected as an argument to the registry package.

Regarding `verifyBuild`, it only have 1 API call and it's just checking the return values, meaning that having it as a private method will look exactly the same regarding mocking (instead of mocking `verifyBuild` we will mock `Sync`.

internal/preflight/preflight.go
ybettan: > what if i change the res to imageVerified?

This will be great.

You can keep it if you want and I won't object to it. Just didn't understood that the condition for a few minutes.
This is how I understood it (wan't intuitive for me):
1. if the image was verified then we should return `true` with the correct message
2. if we couldn't verify the image, we want to try to build it but if there is not build option then we have to return `false` and the relevant message.

I found it very unclear the way it was written without the clear separation of the cases but if you prefer it that way I will merge it as you like.

internal/preflight/preflight.go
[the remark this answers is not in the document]
yevgeny-shnaidman: 1) i don't like the idea of predicate function passed as argument. It makes our code much less readable. And it does not save us anything, it is still needs to be implemented in a calling package. 
2) personally i would prefer not to use private interfaces at all and just write more elaborate unit-test for private functions, but i think we agreed that this discussion will wait until we reach a stage where we try to refactor the module reconciler. As for verifyBuiild - this function will probably be expended in later PRs with push configuration and sign verification, in which case it will have more then 1 mock call. And since we already have private interface that includes verifyImage, adding verifyBuild just make code structure more logical

internal/preflight/preflight.go
yevgeny-shnaidman: please take a look how if it is more readable now, with the changed variable name

ybettan: It is better, still a bit confusing to me but let's merge that way.

internal/preflight/preflight.go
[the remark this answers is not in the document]
ybettan: > And it does not save us anything

It is maintaining a better separation between the logical packages IMO.

Ack on `verifyBuild`. I would still move the registry functionality to the `registry` package with some predicate (that contains the logic which isn't related to the registry package, like we do when we pass the authenticator) but seems like we are not going to agree on it so let's merge :)

ybettan:
/lgtm
/hold
Unhold when you are ready to merge.

PR #66, opened 2022-09-15
Removing KMM nodes' labels upon KMMO undeployment

yevgeny-shnaidman:
This is a raw suggestion of how we should cleanup after KMMO undeployment. Since the label are put on the nodes by the KMMO controller, the suggested solution is:
1) define our own SignalHandler, instead of the default one
   defined by controler-runtime
2) signal handler will watch for SIGINT and SIGTERM signals (
   those a signals issued by cri-o to the pod's containers
3) upon signal reception, signal handler will iterate over
   all the nodes in cluster and remove its labels from them
4) it will then block again and wait for another SIGTERM
   upon which it will exit

Notes:
1) i think it is better to use signal handler then lifecycle hook
   since it allows us more flexibality
2) no matter which mecahinizm we use, we need to make sure that
   the undeploy sequnce is exactly opposite to deploy, meaning:
   ClusterRoles and ClusterRolesBinding must be deleted after
   KMMO deployment deletion

### DIFF
--- a/config/crd/bases/kmm.sigs.k8s.io_modules.yaml
+++ b/config/crd/bases/kmm.sigs.k8s.io_modules.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: unapproved, testing-only
-    controller-gen.kubebuilder.io/version: v0.8.0
+    controller-gen.kubebuilder.io/version: v0.9.2
   creationTimestamp: null
   name: modules.kmm.sigs.k8s.io
 spec:
@@ -117,6 +117,7 @@ spec:
                                   required:
                                   - key
                                   type: object
+                                  x-kubernetes-map-type: atomic
                                 fieldRef:
                                   description: 'Selects a field of the pod: supports
                                     metadata.name, metadata.namespace, `metadata.labels[''<KEY>'']`,
@@ -135,6 +136,7 @@ spec:
                                   required:
                                   - fieldPath
                                   type: object
+                                  x-kubernetes-map-type: atomic
                                 resourceFieldRef:
                                   description: 'Selects a resource of the container:
                                     only resources limits and requests (limits.cpu,
@@ -160,6 +162,7 @@ spec:
                                   required:
                                   - resource
                                   type: object
+                                  x-kubernetes-map-type: atomic
                                 secretKeyRef:
                                   description: Selects a key of a secret in the pod's
                                     namespace
@@ -181,6 +184,7 @@ spec:
                                   required:
                                   - key
                                   type: object
+                                  x-kubernetes-map-type: atomic
                               type: object
                           required:
                           - name
@@ -403,6 +407,7 @@ spec:
                                     uid?'
                                   type: string
                               type: object
+                              x-kubernetes-map-type: atomic
                             user:
                               description: 'user is optional: User is the rados user
                                 name, default is admin More info: https://examples.k8s.io/volumes/cephfs/README.md#how-to-use-it'
@@ -435,6 +440,7 @@ spec:
                                     uid?'
                                   type: string
                               type: object
+                              x-kubernetes-map-type: atomic
                             volumeID:
                               description: 'volumeID used to identify the volume in
                                 cinder. More info: https://examples.k8s.io/mysql-cinder-pd/README.md'
@@ -510,6 +516,7 @@ spec:
                                 or its keys must be defined
                               type: boolean
                           type: object
+                          x-kubernetes-map-type: atomic
                         csi:
                           description: csi (Container Storage Interface) represents
                             ephemeral storage that is handled by certain external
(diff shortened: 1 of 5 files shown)


ybettan:
> i think it is better to use signal handler then lifecycle hook
since it allows us more flexibality

Can you elaborate a bit about it?
What can we do with a signal handler that we can't do with a lifecycle hook?

We are already using lifecycle hooks so I am wondering why should we use a different mechanism here.

yevgeny-shnaidman:
> > i think it is better to use signal handler then lifecycle hook
> > since it allows us more flexibality
> 
> Can you elaborate a bit about it? What can we do with a signal handler that we can't do with a lifecycle hook?
> 
> We are already using lifecycle hooks so I am wondering why should we use a different mechanism here. Isn't it sort of bypassing the lifecycle hooks?

Removing labels from all the nodes cannot be done in a one-line in the hook. It means that we need to run our executable that needs to obtain a "kubeconfig" and go over the nodes and remove labels. If we add executable, we need to always maintain and deploy it, in addition to making sure that it stays updated in case we change the labels. Using signal handler allows us to use already obtained client, no need to write additional executable and maintain it, and the code does not need to be updated in case we change the label

ybettan:
> Removing labels from all the nodes cannot be done in a one-line in the hook

`kubectl label node --all kmm.node.kubernetes.io/kernel-version.full-` seems to be working.

If we want to unlable only `masters` or `workers` it might be a bit more tricky though.

yevgeny-shnaidman:
> > Removing labels from all the nodes cannot be done in a one-line in the hook
> 
> `kubectl label node --all kmm.node.kubernetes.io/kernel-version.full-` seems to be working.
> 
> If we want to unlable only `masters` or `workers` it might be a bit more tricky though.

i think that we want to unlabel them all, no need to be picky. the only issue i have now with hook is that in case we change the label, we need to remember to change it in hook, while using signal handler will automatically use the correct label

qbarrand:
Using signal handlers or lifecycle hooks sounds difficult to me, because it means terminating all workloads that use that label whenever the controller restarts. That means unloading the module and I think that is undesirable in many scenarios.

I am wondering if that label removal should be automated at all. A good solution would be OLM uninstallation hooks, but there does not seem to be such a thing.
I think it's reasonable to leave that to the admin, with the documentation including the one-liner @ybettan proposed.

ybettan:
> the only issue i have now with hook is that in case we change the label, we need to remember to change it in hook, while using signal handler will automatically use the correct label

Yes, this is the downside of the hook.

qbarrand:
> Are we sure this is the case? Will the daemonset remove pods if the label selector was removed?

I don't see any reason why the workload would remain if the DaemonSet's selector does not target a given node anymore.

> In addition to that, if the operator restarts, the podprobe post hook will remove the driver so all the workloads will break anyway no @qbarrand?!

Did you mean `modprobe`? The operator does not run `modprobe` - the module loader pods do. If the operator restarts, all its module loader DaemonSets should remain up and running - I think that's what we should aim for, at least.

yevgeny-shnaidman:
> Using signal handlers or lifecycle hooks sounds difficult to me, because it means terminating all workloads that use that label whenever the controller restarts. That means unloading the module and I think that is undesirable in many scenarios.
> 
> I am wondering if that label removal should be automated at all. A good solution would be OLM uninstallation hooks, but there does not seem to be such a thing. I think it's reasonable to leave that to the admin, with the documentation including the one-liner @ybettan proposed.

Yes, that can be an issue. But this can be easily negated by checking if any Module CR exists.

ybettan:
> Did you mean modprobe?

Yes.

> The operator does not run modprobe - the module loader pods do

Right. My bad.

ybettan:
> Yes, that can be an issue. But this can be easily negated by checking if any Module CR exists.

What are the chances of having the operator installed with 0 CRs? I am afraid that in most cases we won't clean (because we will have CRs) so when we do clean it will look like an unexpected behavior. 

I think it will be better to not clean at all rather than cleaning in some cases that will introduce confusion.

yevgeny-shnaidman:
> > Yes, that can be an issue. But this can be easily negated by checking if any Module CR exists.
> 
> What are the chances of having the operator installed with 0 CRs? I am afraid that in most cases we won't clean (because we will have CRs) so when we do clean it will look like an unexpected behavior.
> 
> I think it will be better to not clean at all rather than cleaning in some cases that will introduce confusion.

The scenario that @qbarrand is afraid of is that operator is restarted, CRs are present and it means downtime for running kenrel modules. In case of a restart user is not even aware that operator is restarted and he certainly does not expect the label to be removed. In case operator is being removed, then it does not matter if the CRs have been uninstalled previously or not, they will also be removed, so it is not an issue if the label is removed.

qbarrand:
@yevgeny-shnaidman shall we close this?

yevgeny-shnaidman:
If it is ok to leave the labels - then sure. I remember you said that in the community meeting Intel we not sure if those labels can be left unattended

PR #68, opened 2022-09-16
add signing as part of the build job

chr15p:
add the secureboot signing of kernel modules by converting the build process into a multistage process

This:
- Adds a ...Build.Sign sub section to the CRD for defining the certificate and key secret information for signing.
- It modifies the job manager to loop through a list of jobs in order, waiting for each to finish before it starts the next one, and waiting for all of them to finish before it reports the build as complete.
- It adds a ShouldRun() function to the Maker interface so the job manager loop can determine if a given job is required to run (so the build does not run if Build.Dockerfile is not defined, and signing doesn't run if Build.Sign is not defined in the CR)
- If this is not the last job to run it  changes the name of the image produced by the job by appending the job name, and feeds that into the next job with the final job producing containerImage

- Finally:   It adds a new (signing.go) k8s job definition that actually runs the signing process itself

A similar set of changes will need to be added to buildconfig, shipwright or any other future build processes we consider, but that would be a different PR...


internal/build/job/signer.go
```diff
@@ -0,0 +1,184 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+////go:generate mockgen -source=signer.go -package=signer -destination=mock_signer.go
+
+/*
+type Signer interface {
```
yevgeny-shnaidman, line 20: leftover

api/v1beta1/module_types.go
```diff
@@ -79,6 +96,10 @@ type Build struct {
 	// +optional
 	// KanikoParams is used to customize the building process of the image.
 	KanikoParams *KanikoParams `json:"kanikoParams,omitempty"`
+
+	// +optional
+	// Sign enables in-cluster signing for this mapping
+	Sign *Sign `json:"sign"`
```
yevgeny-shnaidman, line 102: Why is Sign under the Build? What if we want to sign a pre-built image?

internal/build/job/maker.go
```diff
@@ -16,19 +16,43 @@ import (
 //go:generate mockgen -source=maker.go -package=job -destination=mock_maker.go
 
 type Maker interface {
```
yevgeny-shnaidman, line 18: Maker interface is not implemented by 2 entities. Maybe we can move its definition to a manager file?

internal/build/job/maker.go
```diff
@@ -16,19 +16,43 @@ import (
...
+}
+
+func (m *maker) ShouldRun(mod *kmmv1beta1.Module, km *kmmv1beta1.KernelMapping) bool {
+	// if theres no Dockerfile we have no way to build anything anyway
+
+	if (mod.Spec.ModuleLoader.Container.Build == nil || mod.Spec.ModuleLoader.Container.Build.Dockerfile == "") && (km.Build == nil || km.Build.Dockerfile == "") {
+		return false
+	}
+	/*
+		if mod.Spec.ModuleLoader.Container.Build == nil && km.Build == nil {
```
yevgeny-shnaidman, line 45: leftover

internal/build/job/signer.go
```diff
@@ -0,0 +1,184 @@
...
+						Key:  key,
+						Path: path,
+					},
+				},
+			},
+		},
+	}
+}
+
+func (m *signer) makeImageSigningSecretVolumeMount(secretRef *v1.LocalObjectReference, mountpoint string) v1.VolumeMount {
```
yevgeny-shnaidman, line 138: we have makeVolume functions both in signer.go and in maker.go. Can we probably move those functions into some kind of helper file, and uses them with appropriate parameters from both files?

api/v1beta1/module_types.go
chr15p: because the code (i.e. this PR) sits under the build code, so it feels to me that is should do in the CRD as well. I see the "build" process as "create an image we can use" not just "create an image", and if the kmod isn't signed we cant use it.

 It also means I can leverage helper.GetRelevantBuild() to calculate the config to use from the kernelmapping which makes life simpler.  

If we're signing a pre-built image you just need kernelmapping.build.sign e.g.
```
      kernelMappings:
        - regexp: '^.*\.x86_64$'
          containerImage: quay.io/chrisp262/minimal-driver:test1-signed6
          build:
            sign:
              unsignedImage: quay.io/chrisp262/minimal-driver:procfsv1
              keySecret:
                name: securebootkey
              certSecret:
                name: securebootcert
              filesToSign:
                - /modules/simple-kmod.ko
                - /modules/simple-procfs-kmod.ko
```
I am open to moving it to be a peer of build in the crd (so kernelmapping.build and kernelmapping.sign) if people prefer.  (I originally had it this way, with sign being its own package alongside job but it turned into a maintainability nightmare)

yevgeny-shnaidman: i personally think that it is confusing: build is associated with in-cluster build, and signing is a little different. But lets have other people's opinion: @qbarrand @ybettan

internal/build/job/signer.go
chr15p: yes, removed

internal/build/job/maker.go
chr15p: good point, I have separated it out into its own maker.go file, and renamed the maker instance (that implements Maker) to builder because it does the build. So now we have signer.go and builder.go that implement the Maker interface. Hopefully that is both clearer and better structured than before.

internal/build/job/signer.go
```diff
@@ -0,0 +1,175 @@
...
+			Labels:       labels(mod, targetKernel, m.GetName()),
+		},
+		Spec: batchv1.JobSpec{
+			Completions: pointer.Int32(1),
+			Template: v1.PodTemplateSpec{
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{
+						{
+							Name:         "signimage",
+							Image:        "quay.io/chrisp262/kmod-signer:latest",
```
fabiendupont, line 91: Could it become a setting rather than a hard coded value in the Go code?
Shouldn't it point to an image released under the KMM project? And if yes, could you provide the Dockerfile in the PR to make it complete?

qbarrand: > Could it become a setting rather than a hard coded value in the Go code?

Is that to be able to run substitutions when creating the bundle?
Besides that, I hardly see a use case for users to provide their own signer.

chr15p: the source for that image comes from https://github.com/chr15p/kmod_signer  and the idea is to move that code/Dockerfile into the kmm project as a separate PR, at which point this would obviously get updated from pointing at my quay.io repo.
I did consider making it user configurable but as this points to the only image in the world that provides this functionality it didn't seem like a high priority :)

api/v1beta1/module_types.go
chr15p: The (kaniko) build and signing are very similar tasks, most of the differences are in the job pod they run so separating off signing completely leads to a lot of duplicated job-management code, basically a copy/paste of internal/build/job/manager with the field names changed, and that feels like a maintenance mess of its own verses making the build manager a generic job manager and using the same code for both.

I did try using the reconciler to run the jobs (basically running handleBuild() twice, for build then sign), which is in some ways a nicer design, but it involves more and wider changes in the code which is still changing a lot and meant I was spending more time in merge hell trying to rebase onto main and keep up with changes from other PRs then on making this feature work (one step forward two steps back!). So this design is an attempt to balance between the scope of changes it makes and code reuse.

On a narrower point I'm not sure I see signing as separate from building, its a step in the build process. If you're using secure boot then the kaniko output is just an intermediate artifact that isn't usable until the kmods are signed. So the steps are separate in the same sense creating object files and linking are separate processes, technically they are, but in reality they both need to be done to build something useful.

yevgeny-shnaidman: the issue is not the kaniko output, but that customer can use the signing option without kaniko at all: he has pre-built images, but instead of trying to sign them off-cluster ( which can be painful job), he will just lets us sign them in-cluster. In which case it does not make sense for customer to define Build context, when no actual build is being done. Regarding code split: how is it going to be done in downstream?  There we have build config for image building, but we won't use it for signing, correct?

chr15p: but whether the intermediate image comes from kaniko or from outside you are still building it for use in the cluster, so I think it does make sense to define a Build context.  I guess the argument is about whether Build is "create a usable image" or "create an intermediate image we need to do more stuff to before kmmo can use it", I'm (clearly :) ) going for the first option.

As for buildconfig, iirc that code still implements the Maker interface so we can still replace the maker instance with it (or with some cleverness in ShouldRun() have kaniko, buildconfig, and signing in the same binary and check which ones to use at runtime)

chr15p:
superceeded

PR #76, opened 2022-09-19
Adding support for image push during preflight build verification

yevgeny-shnaidman:
1) add a field to the Preflight CRD spec , that signifies if customer
   wants to push the image during preflight verification
2) add support for image push during preflight Build stage 3) add Generator filter to preflight Watch, to skip being scheduled
   on Preflight status changes

### DIFF
--- a/internal/preflight/preflight.go
+++ b/internal/preflight/preflight.go
@@ -26,7 +26,7 @@ const (
 //go:generate mockgen -source=preflight.go -package=preflight -destination=mock_preflight_api.go PreflightAPI, preflightHelperAPI
 
 type PreflightAPI interface {
-	PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
+	PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module) (bool, string)
 }
 
 func NewPreflightAPI(
@@ -49,8 +49,9 @@ type preflight struct {
 	helper        preflightHelperAPI
 }
 
-func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
+func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mod *kmmv1beta1.Module) (bool, string) {
 	log := ctrlruntime.LoggerFrom(ctx)
+	kernelVersion := pv.Spec.KernelVersion
 	mapping, err := p.kernelAPI.FindMappingForKernel(mod.Spec.ModuleLoader.Container.KernelMappings, kernelVersion)
 	if err != nil {
 		return false, fmt.Sprintf("Failed to find kernel mapping in the module %s for kernel version %s", mod.Name, kernelVersion)
@@ -77,12 +78,12 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
 		log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to build stage"), "module", mod.Name, "error", err)
 	}
 
-	return p.helper.verifyBuild(ctx, mapping, mod, kernelVersion)
+	return p.helper.verifyBuild(ctx, pv, mapping, mod)
 }
 
 type preflightHelperAPI interface {
 	verifyImage(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
-	verifyBuild(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
+	verifyBuild(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module) (bool, string)
 }
 
 type preflightHelper struct {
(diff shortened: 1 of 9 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management
-  newTag: main
+  newName: quay.io/yshnaidm/kmmo
```
qbarrand, line 15: Please dismiss those changes.

internal/preflight/preflight.go
```diff
@@ -131,15 +131,23 @@ func (p *preflightHelper) verifyImage(ctx context.Context, mapping *kmmv1beta1.K
...
+	buildRes, err := p.buildAPI.Sync(ctx, *mod, *mapping, kernelVersion, pv.Spec.PushBuiltImage)
 	if err != nil {
 		return false, fmt.Sprintf("Failed to verify build for module %s, kernel version %s, error %s", mod.Name, kernelVersion, err)
 	}
 
 	if buildRes.Status == build.StatusCompleted {
-		return true, fmt.Sprintf(VerificationStatusReasonVerified, "build compiles")
+		msg := "build compiles"
+		if pv.Spec.PushBuiltImage {
+			msg = msg + " and image pushed"
```
qbarrand, line 148:
```suggestion
			msg += " and image pushed"
```

internal/preflight/preflight.go
```diff
@@ -131,15 +131,23 @@ func (p *preflightHelper) verifyImage(ctx context.Context, mapping *kmmv1beta1.K
 	return false, fmt.Sprintf("image %s does not contain kernel module for kernel %s on any layer", image, kernelVersion)
 }
 
-func (p *preflightHelper) verifyBuild(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
+func (p *preflightHelper) verifyBuild(ctx context.Context,
+	pv *kmmv1beta1.PreflightValidation,
```
qbarrand, line 135: How about passing just the boolean instead?

config/manager/kustomization.yaml
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
yevgeny-shnaidman: right now i prefer to pass the pointer to the whole object, since i don't know yet which data i will need in order to implement the sign section. Anyway, once Chris commit is in and we know for sure the how it is going to be implemented, i will probably submit addition PR for a better refactoring

internal/preflight/preflight.go
```diff
@@ -77,12 +77,12 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
 	}
 
-	return p.helper.verifyBuild(ctx, mapping, mod, kernelVersion)
+	return p.helper.verifyBuild(ctx, pv, mapping, mod, kernelVersion)
 }
 
 type preflightHelperAPI interface {
 	verifyImage(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
-	verifyBuild(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
+	verifyBuild(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
```
ybettan, line 85: I am guessing that the `kernelVersion` inside `pv` is the new kernel version and the one passed stand alone is the current kernel version?

Can we rename those? I find it confusing and seems that we are just passing `kernelVersion` twice to function in different places in the code.

yevgeny-shnaidman: actually they are the same. We do have duplicate data here. Since we need the PreflightValidation as part of the API, i will start using (passing) it into the functions, and will use its data inside functions as needed

internal/preflight/preflight.go
```diff
@@ -77,12 +78,12 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
 	}
 
-	return p.helper.verifyBuild(ctx, mapping, mod, kernelVersion)
+	return p.helper.verifyBuild(ctx, pv, mapping, mod)
 }
 
 type preflightHelperAPI interface {
 	verifyImage(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
-	verifyBuild(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string)
+	verifyBuild(ctx context.Context, pv *kmmv1beta1.PreflightValidation, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module) (bool, string)
```
ybettan, line 86: Should we replace the kernelVersion with pv in `verifyImage` as well?

internal/preflight/preflight.go
```diff
@@ -131,15 +132,22 @@ func (p *preflightHelper) verifyImage(ctx context.Context, mapping *kmmv1beta1.K
...
 }
 
-func (p *preflightHelper) verifyBuild(ctx context.Context, mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (bool, string) {
+func (p *preflightHelper) verifyBuild(ctx context.Context,
+	pv *kmmv1beta1.PreflightValidation,
+	mapping *kmmv1beta1.KernelMapping,
+	mod *kmmv1beta1.Module) (bool, string) {
 	// at this stage we know that eiher mapping Build or Container build are defined
-	buildRes, err := p.buildAPI.Sync(ctx, *mod, *mapping, kernelVersion, false)
+	buildRes, err := p.buildAPI.Sync(ctx, *mod, *mapping, pv.Spec.KernelVersion, pv.Spec.PushBuiltImage)
```
ybettan, line 140: Why don't we just pass `pv` or `pv.Spec`?

internal/preflight/preflight.go
yevgeny-shnaidman: for now - no, i don't need pv there. As i wrote above, once Chris PR is in and i know for sure how signing is implemented, i will probably refactor the APIs

internal/preflight/preflight.go
yevgeny-shnaidman: This is build API of build manager. It is also used for Module for in-cluster builts, it can receive only general values and not specific structs

PR #81, opened 2022-09-21
Update of Build configuration in Module should trigger new Job

yevgeny-shnaidman:
Currently once the Build job is created, it cannot be changed. Therefore even if user changes Build configuration in the module, the existing job will not be changed. This fixes it by doing the following 1) when job is created, Template data is hashed and saved as annotation
   in the job
2) when checking job status, hash of the new contents is compared
   to the existing one
3) in case hashes differ, the current job is deleted, which will
   trigger new job creation in the next reconcilation loop
4) in case job is in progress, reconcile should use RequeueAfter
   field with predefined value, otherwise it will cause exponential
   requeing timeouts

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -87,7 +87,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
-//+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch
+//+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch;delete
 
 // Reconcile lists all nodes and looks for kernels that match its mappings.
 // For each mapping that matches at least one node in the cluster, it creates a DaemonSet running the container image
--- a/internal/build/job/maker.go
+++ b/internal/build/job/maker.go
@@ -5,6 +5,7 @@ import (
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
+	"github.com/mitchellh/hashstructure"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -16,7 +17,7 @@ import (
 //go:generate mockgen -source=maker.go -package=job -destination=mock_maker.go
 
 type Maker interface {
-	MakeJob(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string, pushImage bool) (*batchv1.Job, error)
+	MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string, pushImage bool) (*batchv1.Job, error)
 }
 
 type maker struct {
@@ -28,7 +29,7 @@ func NewMaker(helper build.Helper, scheme *runtime.Scheme) Maker {
 	return &maker{helper: helper, scheme: scheme}
 }
 
-func (m *maker) MakeJob(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string, pushImage bool) (*batchv1.Job, error) {
+func (m *maker) MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string, pushImage bool) (*batchv1.Job, error) {
 	args := []string{}
 	if pushImage {
 		args = append(args, "--destination", containerImage)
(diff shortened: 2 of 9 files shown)


controllers/module_reconciler.go
```diff
@@ -47,6 +48,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const buildInProgressTimeout = 10
```
ybettan, line 51: Is it in minutes? If yes can we rename `buildInProgressTimeoutInMinutes`?

controllers/module_reconciler.go
```diff
@@ -131,7 +134,7 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 		if requeue {
 			logger.Info("Build requires a requeue; skipping handling driver container for now", "kernelVersion", kernelVersion, "image", m)
-			res.Requeue = true
+			res.RequeueAfter = time.Second * buildInProgressTimeout
```
ybettan, line 137: Look like this is the interval waiting time. Should we change the const name to `buildInProgressInterval`?

yevgeny-shnaidman: what about buildInProgressRequeueInterva ?l

ybettan:
> in case job is in progress, reconcile should use RequeueAfter
field with predefined value, otherwise it will cause exponential
requeing timeouts

Can you please elaborate on that part please? Why do you mean by exponential timeout?

controllers/module_reconciler.go
ybettan: Sure. Just the `Timeout` word here is confusing - I get the same confusion in the commit message as well BTW.

yevgeny-shnaidman:
> > in case job is in progress, reconcile should use RequeueAfter
> > field with predefined value, otherwise it will cause exponential
> > requeing timeouts
> 
> Can you please elaborate on that part please? Why do you mean by exponential timeout?

It means that every time a Reconcile returns Requeue = true without specifying the precise timeout, controller-runtime will requeue with an increasing interval, something like exponential backoff in case of error ( but not as severe). So after returning Requeue on the same object 5-6 time, then next requeue will come in ~2 minutes.  So, in case job has completed, we might check only after a couple of minutes has passed

internal/build/job/maker.go
```diff
@@ -97,32 +99,39 @@ func (m *maker) MakeJob(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, ta
...
+	if err != nil {
+		return nil, fmt.Errorf("could not hash job's spec template: %v", err)
+	}
+
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: mod.Name + "-build-",
 			Namespace:    mod.Namespace,
 			Labels:       labels(mod, targetKernel),
+			Annotations:  map[string]string{jobHashAnnotation: fmt.Sprintf("%d", specTemplateHash)},
```
ybettan, line 130: Is `fmt.Sprintf("%d", specTemplateHash)` working?
I tried `fmt.Println(string(specTemplateHash))` and I got an empty value.

controllers/module_reconciler.go
yevgeny-shnaidman: fixed. Corrected the commit message also

internal/build/job/manager.go
```diff
@@ -83,6 +87,20 @@ func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1b
 		return build.Result{Status: build.StatusCreated, Requeue: true}, nil
 	}
 
+	changed, err := jbm.isJobChanged(mod, buildConfig, targetKernel, m.ContainerImage, pushImage, job)
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not determine if job has changed: %v", err)
+	}
+
+	if changed {
+		logger.Info("Job spec changed, deleting..", "name", job.Name)
```
ybettan, line 96:
```suggestion
		logger.Info("The module's build spec has been changed, deleting the current job so a new one can be created", "name", job.Name)
```

internal/build/job/maker.go
```diff
@@ -11,6 +11,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	"github.com/mitchellh/hashstructure"
```
qbarrand, line 15:
```suggestion
	"github.com/mitchellh/hashstructure"
```

internal/build/job/manager.go
```diff
@@ -11,8 +11,12 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 )
 
+const jobHashAnnotation = "last-hash"
```
qbarrand, line 18:
```suggestion
const jobHashAnnotation = "kmm.node.kubernetes.io/last-hash"
```

internal/build/job/manager.go
```diff
@@ -83,6 +87,20 @@ func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1b
...
+	changed, err := jbm.isJobChanged(mod, buildConfig, targetKernel, m.ContainerImage, pushImage, job)
+	if err != nil {
+		return build.Result{}, fmt.Errorf("could not determine if job has changed: %v", err)
+	}
+
+	if changed {
+		logger.Info("Job spec changed, deleting..", "name", job.Name)
+		err = jbm.client.Delete(ctx, job)
+		if err != nil {
+			logger.Info(utils.WarnString(fmt.Sprintf("failed to delete build job %s", job.Name)))
```
qbarrand, line 99:
```suggestion
		if err := jbm.client.Delete(ctx, job); err != nil {
			logger.Info(utils.WarnString(fmt.Sprintf("failed to delete build job %s: %v", job.Name, err)))
```

internal/build/job/manager.go
```diff
@@ -83,6 +87,20 @@ func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1b
 		return build.Result{Status: build.StatusCreated, Requeue: true}, nil
 	}
 
+	changed, err := jbm.isJobChanged(mod, buildConfig, targetKernel, m.ContainerImage, pushImage, job)
```
qbarrand, line 90: Should we always generate the new job and compare it to the existing one? That would simplify `isJobChanged` and reduce the number of parameters it receives.

internal/build/job/manager.go
qbarrand: Could we patch the Job instead of deleting it?

internal/build/job/manager.go
yevgeny-shnaidman: fixed

internal/build/job/maker.go
yevgeny-shnaidman: fixed

internal/build/job/manager.go
yevgeny-shnaidman: fixed

internal/build/job/manager.go
yevgeny-shnaidman: no, jobs specs cannot not be patched, kubernetes restriction

controllers/module_reconciler.go
ybettan: TY

internal/build/job/manager.go
ybettan: Why would we want to generate a job even if there is no need for it?

qbarrand:
>  So, in case job has completed, we might check only after a couple of minutes has passed

@yevgeny-shnaidman we watch Jobs and set the owner reference to the `Module`, so whenever the Job status will change, the corresponding `Module` will be reconciled. For that reason, I would not set `RequeueAfter` and remain on the standard behavior hier.

internal/build/job/manager.go
yevgeny-shnaidman: yes, good suggestion it will simplify everything, fixed

yevgeny-shnaidman:
> > So, in case job has completed, we might check only after a couple of minutes has passed
> 
> @yevgeny-shnaidman we watch Jobs and set the owner reference to the `Module`, so whenever the Job status will change, the corresponding `Module` will be reconciled. For that reason, I would not set `RequeueAfter` and remain on the standard behavior here.

yes, you are right, reverting this change

internal/build/job/manager.go
ybettan: I though that @qbarrand meant to always create a new job in k8s and then compare the 2 running jobs.
Only after the changes I understood what you meant :)

ybettan:
lgtm-waiting-for-other-approvers

internal/build/job/manager.go
yevgeny-shnaidman: I am renaming MakeJob -> MakeJobTemplate  to avoid future confusion

PR #85, opened 2022-09-22
Fixing build job deletion, in case its pods are still running

yevgeny-shnaidman:
When we delete build job (due to Spec change) while its pods are still running/failed, those pods are not deleted and will stay there untill deleted manually. This PR adds a Propagate option to the Delete API call that will instaract GC to remove the pods of the deleted job


ybettan:
/lgtm
/hold
Wait for Github actions (tied isn't enforcing them ATM).

PR #86, opened 2022-09-22
Creating TLSOptions in favor of PullOptions and PushOptions.

ybettan:
The `build` object, always push to the same registry that the
`kernelMapping` will pull from, therefore, there is not need to
duplicate the config.

In addition that, I have also renamed some of the fields for better
self-documented code.

Here is how this field is used in the module:

* `kernelMapping[].build.baseImageRegistryTLS` is only meant for pulling the
   base image of the Dockerfile specified in the build.

* `kernelMapping[].registryTLS` is meant for specifying the TLS options for
  pulling the image we want to deploy or for pushing it in case we are building
  it in-cluster (because they will always be the same).

* `moduleLoaderContainerSpec.registryTLS` is just the global value of all
  kernelMapping[]'s entries.


api/v1beta1/module_types.go
```diff
@@ -27,21 +27,10 @@ type BuildArg struct {
 	Value string `json:"value"`
 }
 
-type PullOptions struct {
+type TLSOptions struct {
```
ybettan, line 30:
```suggestion
type RegistryTLS struct {
```

api/v1beta1/module_types.go
```diff
@@ -63,12 +52,8 @@ type Build struct {
...
 
 	// +optional
-	// Pull contains settings determining how to pull the base images of the build process.
-	Pull PullOptions `json:"pull"`
-
-	// +optional
-	// Push contains settings determining how to push a built DriverContainer image.
-	Push PushOptions `json:"push"`
+	// BaseImagePull contains settings determining how to pull the base images of the build-process' Dockerfile.
+	BaseImagePull TLSOptions `json:"baseImagePull"`
```
ybettan, line 56:
```suggestion
	BaseImageRegistryTLS RegistryTLS `json:"baseImageRegistryTLS"`
```

qbarrand:
If we change the CRD now, we probably need to implement a conversion webhook to avoid breaking clients.
https://book.kubebuilder.io/multiversion-tutorial/tutorial.html

ybettan:
@qbarrand @yevgeny-shnaidman Are we OK to merge it without a conversion webhook at this point?

ybettan:
@yevgeny-shnaidman @qbarrand PTAL

qbarrand:
Should we also define `kernelMapping[].sign.baseImageRegistryTLS`?

api/v1beta1/module_types.go
```diff
@@ -63,12 +52,8 @@ type Build struct {
 	Dockerfile string `json:"dockerfile"`
 
 	// +optional
-	// Pull contains settings determining how to pull the base images of the build process.
-	Pull PullOptions `json:"pull"`
-
-	// +optional
-	// Push contains settings determining how to push a built DriverContainer image.
-	Push PushOptions `json:"push"`
+	// BaseImageRegistryTLS contains settings determining how to pull the base images of the build-process' Dockerfile.
```
yevgeny-shnaidman, line 55: why is it used only for pulling image, but not for pushing? You have pushing in the KernelMapping, but not here. Prior to your change it was on the opposite. What am i missing?

internal/build/job/maker.go
```diff
@@ -19,7 +19,8 @@ import (
 //go:generate mockgen -source=maker.go -package=job -destination=mock_maker.go
 
 type Maker interface {
-	MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string, pushImage bool) (*batchv1.Job, error)
+	MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.Build, targetKernel, containerImage string,
```
yevgeny-shnaidman, line 22: If we TLS option fo pushing will be defined in the Build , and not in kernelMapping, then this API needs not to be changed

api/v1beta1/module_types.go
ybettan: This is one of the reason that PR was created. We don't need both.

We are always pushing to the same image registry as the one we are trying to pull the image from to begin with (we are building if the image doesn't exist), therefore, there is no reason to duplicate those configs. We should take them from the kernelMapping instead.

internal/build/job/maker.go
ybettan: It is not. It is defined in the kernelMapping/moduleLoader

ybettan:
> Should we also define kernelMapping[].sign.baseImageRegistryTLS?

I don't think we do. `baseImageRegistryTLS` is only for pulling a base image while building, for pulling a pre-built image or for pushing a just-built image we are using `km.registryTLS`  anyway.

@chr15p Am I missing something here?

api/v1beta1/module_types.go
yevgeny-shnaidman: Why the TLS, which is needed only in build, is put outside of the build? it is not a very clear API.

ybettan: It is not only needed for build, it is needed for access that image registry, if the image doesn't exist then we will build it and push it but we still need those configs for checking if the image exists prior to the build.

chr15p:
> > Should we also define kernelMapping[].sign.baseImageRegistryTLS?
> 
> I don't think we do. `baseImageRegistryTLS` is only for pulling a base image while building, for pulling a pre-built image or for pushing a just-built image we are using `km.registryTLS` anyway.
> 
> @chr15p Am I missing something here?

Signing pulls an image, signs its kmods, then pushes it back again. So it both pulls and pushes, and theoretically they could be to/from different registrys.  We're currently using spec.ImageRepoSecret for the creds to do the pushing and pulling which it doesn't look like you're changing so I think we're ok.

We might want to add per sign pull/push secrets, it would probably be a good thing, but at the moment we dont do that.

ybettan:
> We might want to add per sign pull/push secrets, it would probably be a good thing, but at the moment we dont do that.

This PR isn't about secret but rather about transport (TLS) options.

api/v1beta1/module_types.go
```diff
@@ -199,8 +183,8 @@ type ModuleLoaderContainerSpec struct {
 	Modprobe ModprobeSpec `json:"modprobe"`
 
 	// +optional
-	// Pull contains settings determining how to check if the ModuleLoader image already exists.
-	Pull *PullOptions `json:"pull"`
+	// RegistryTLS set the TLS configs for accessing the registry of the driver-container's image.
```
yevgeny-shnaidman, line 186: driver-container's image -> module loader image

api/v1beta1/module_types.go
```diff
@@ -118,9 +103,8 @@ type KernelMapping struct {
 	Literal string `json:"literal"`
 
 	// +optional
-	// Pull contains settings determining how to check if the ModuleLoader image already exists
-	// and allows overriding of the ModuleLoader's pull options
-	Pull *PullOptions `json:"pull"`
+	// RegistryTLS set the TLS configs for accessing the registry of the driver-container's image.
```
yevgeny-shnaidman, line 106: driver-container's image -> module loader image

ybettan: Fixed.

qbarrand:
> `baseImageRegistryTLS` is only for pulling a base image while building

Signing does essentially the same as building, I don't see any reason why we shouldn't have the same TLS options.

yevgeny-shnaidman:
> > `baseImageRegistryTLS` is only for pulling a base image while building
> 
> Signing does essentially the same as building, I don't see any reason why we shouldn't have the same TLS options.

Signing is accessing the same registry as building. Do you see any scenario where we execute sign without build?

ybettan:
> Signing does essentially the same as building, I don't see any reason why we shouldn't have the same TLS options.

Signing isn't pulling any "base image" as `build` does. It is only pulling existing images, hence, using the `km.registryTLS` (same as build is doing when pushing).

`build.baseImageRegistryTLS` is only used to pull the base image withing a Dockerfile, we don't have such in signing.

ybettan:
As agree with @qbarrand TLS options doesn't exist for signing in this PR, therefore, they will be added separately in a different PR.

PR #88, opened 2022-09-27
centralise the name of the image to build in the controller

chr15p:
The name of the image to build is assumed to be km.ContainerImage in several places in the code, this PR centralises that assumption into a single point in Reconcile(). This should make the code more maintainable in the future when that assumption may no longer be true.  

E.g if we add a signing stage to the image build process, then build itself needs to produce an intermediate image that is consumed by the signing job, and it is the signing job that needs to produce km.ContainerImage. 

(I also notice we have a module.spec.moduleLoader.container.containerImage field, which is presumably meant to be a default, but we cannot use it while the km.ContainerImage assumption is made in lots of different places, )

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -141,7 +141,7 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	for kernelVersion, m := range mappings {
-		requeue, err := r.handleBuild(ctx, mod, m, kernelVersion)
+		requeue, err := r.handleBuild(ctx, mod, m, kernelVersion, m.ContainerImage)
 		if err != nil {
 			return res, fmt.Errorf("failed to handle build for kernel version %s: %w", kernelVersion, err)
 		}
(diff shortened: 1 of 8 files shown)


chr15p:
> 
I like the general idea, and I had thought about it, but at the moment afaict  `module.spec.moduleLoader.container.kernelmapping` and  `module.spec.moduleLoader.container.kernelmapping.containerImage` are both required fields so we would never use a default image name.

To me the answer is to make kernelmapping a non-required field and if its not defined use `module.spec.moduleLoader.container.containerImage` and `module.spec.moduleLoader.container.build`  but that felt like a bigger change and a wider conversation.  

I like your suggestion though, so I will add it into this PR basically as a placeholder and we can sort out the CRD issues later.

qbarrand:
> To me the answer is to make kernelmapping a non-required field and if its not defined use `module.spec.moduleLoader.container.containerImage` and `module.spec.moduleLoader.container.build` but that felt like a bigger change and a wider conversation.

Agreed. #89 should check that at least one of the two is defined - we can implement the fallback function then.

qbarrand:
/approve
/cc @ybettan 
Please review and `/lgtm` if you are happy.

yevgeny-shnaidman:
> > 
> 
> I like the general idea, and I had thought about it, but at the moment afaict `module.spec.moduleLoader.container.kernelmapping` and `module.spec.moduleLoader.container.kernelmapping.containerImage` are both required fields so we would never use a default image name.
> 
> To me the answer is to make kernelmapping a non-required field and if its not defined use `module.spec.moduleLoader.container.containerImage` and `module.spec.moduleLoader.container.build` but that felt like a bigger change and a wider conversation.
> 
> I like your suggestion though, so I will add it into this PR basically as a placeholder and we can sort out the CRD issues later.

@chr15p just one note: kernel mapping itself is required, what you are suggesting is making the containerImage inside the kernel mapping optional, right?

chr15p:
> @chr15p just one note: kernel mapping itself is required, what you are suggesting is making the containerImage inside the kernel mapping optional, right?

Yes, for the purposes of defaulting to a global containerImage it's only the kernelmapping.containerImage that would be optional, not the whole kernelMapping. If for some reason we dont want to do that then we should probably just remove `module.spec.moduleLoader.container.containerImage` entirely (its not actually used at the moment anyway afaict)

tbh if `module.spec.moduleLoader.container.containerImage` and `module.spec.moduleLoader.container.build` are present then thats probably enough information and kernelMapping _could_ be optional (you'd get the same kmod for every kernel version which is probably not what you want, but would still work in a subset of cases) but that is definitely a wider conversation than this PR :)

ybettan:
@chr15p Can you please fill in the commit message?
You can basically just copy in the PR description or follow https://github.com/kubernetes-sigs/kernel-module-management/blob/main/CONTRIBUTING.md#commit-messages.

ybettan:
/lgtm
/hold
Wait for @yevgeny-shnaidman.
@qbarrand Do you know why aren't the Github actions running?

qbarrand:
@ybettan

image

Should be good now.

PR #90, opened 2022-09-27
Add ServiceAccounts for the ModuleLoader and DevicePlugin workloads

mresvanis:
This PR adds ServiceAccounts for the module loader and the device plugin workloads of a `Module`, when no `ServiceAccount`s are specified in the `Module.Spec`. 

This way we avoid using the K8S `default` `ServiceAccount`, which may have different permissions than the module loader and/or the device plugin workloads require.

Specifically, this PR:

- adds the required `ServiceAccount` permissions to the module reconciler
- adds the following `Module` reconciliation steps when the `Module.Spec.ModuleLoader.ServiceAccountName` is empty: 
  - adds a module loader `ServiceAccount`
  - adds the module loader`ServiceAccount` to the module loader DaemonSet
- adds the following `Module` reconciliation steps when the `Module.Spec.DevicePlugin.ServiceAccountName` is empty:
  - adds a device plugin `ServiceAccount`
  - adds the device plugin `ServiceAccount` to the device plugin DaemonSet
  

qbarrand:
This LGTM. Please fix the linting issue and squash into one commit.
/approve

mresvanis:
@ybettan @yevgeny-shnaidman this PR is now ready for review, whenever you have some time it would be great to take a look into it. Thank you!

controllers/module_reconciler.go
```diff
@@ -109,6 +114,17 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	r.setKMMOMetrics(ctx)
 
+	if mod.Spec.ModuleLoader.ServiceAccountName == "" {
+		if err := r.rbacAPI.CreateModuleLoaderRBAC(ctx, *mod); err != nil {
+			return res, fmt.Errorf("could not reconcile module loader RBAC: %w", err)
+		}
+	}
+	if mod.Spec.DevicePlugin != nil && mod.Spec.DevicePlugin.ServiceAccountName == "" {
+		if err := r.rbacAPI.CreateDevicePluginRBAC(ctx, *mod); err != nil {
+			return res, fmt.Errorf("could not reconcile device plugin RBAC: %w", err)
```
ybettan, line 124:
```suggestion
			return res, fmt.Errorf("could not create device-plugin's RBAC: %w", err)
```

controllers/module_reconciler.go
```diff
@@ -109,6 +114,17 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	r.setKMMOMetrics(ctx)
 
+	if mod.Spec.ModuleLoader.ServiceAccountName == "" {
+		if err := r.rbacAPI.CreateModuleLoaderRBAC(ctx, *mod); err != nil {
+			return res, fmt.Errorf("could not reconcile module loader RBAC: %w", err)
```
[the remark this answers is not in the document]
ybettan: BTW, just our of curiosity, why are we failing here? Don't we want to default to the k8s default ServiceAccount (with a warning log) in this case?

controllers/module_reconciler.go
```diff
@@ -359,6 +375,7 @@ func (r *ModuleReconciler) SetupWithManager(mgr ctrl.Manager, kernelLabel string
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&kmmv1beta1.Module{}).
 		Owns(&appsv1.DaemonSet{}).
+		Owns(&v1.ServiceAccount{}).
```
ybettan, line 378: Shouldn't we make the DaemonSet the owner of the serviceAccount herr?

internal/rbac/rbac.go
```diff
@@ -0,0 +1,84 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+)
+
+//go:generate mockgen -source=rbac.go -package=rbac -destination=mock_rbac.go
+
+type RBACCreator interface {
+	CreateModuleLoaderRBAC(ctx context.Context, mod kmmv1beta1.Module) error
+	CreateDevicePluginRBAC(ctx context.Context, mod kmmv1beta1.Module) error
```
ybettan, line 21: What do you think about 
```suggestion
	CreateModuleLoaderServiceAccount(ctx context.Context, mod kmmv1beta1.Module) error
	CreateDevicePluginServiceAccount(ctx context.Context, mod kmmv1beta1.Module) error
```
?

We are basically creating a ServiceAccount here.

controllers/module_reconciler.go
mresvanis: There are two ServiceAccounts per Module CR (one for the module loader and one for the device plugin) and both are owned ([module loader SA](https://github.com/kubernetes-sigs/kernel-module-management/pull/90/files#diff-98d75845082e0ae72cd6181df75be68f64cca9558e74b7122da4aadcfe0b0d0cR47) and [device plugin SA](https://github.com/kubernetes-sigs/kernel-module-management/pull/90/files#diff-98d75845082e0ae72cd6181df75be68f64cca9558e74b7122da4aadcfe0b0d0cR68)) by the Module CR and managed by the controller. 

Here, the controller-owned ServiceAccounts are setup to be watched, in order for the controller to respond to create / delete / update events on the former by reconciling the owner Module CR (and subsequently the ServiceAccounts themselves).

Does this make sense? Did I understand your question correctly?

internal/rbac/rbac.go
```diff
@@ -0,0 +1,84 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("cound not reconcile device plugin ServiceAccount: %w", err)
+	}
+	logger.Info("Reconciled device plugin ServiceAccount", "name", sa.Name, "result", opRes)
+
+	return nil
+}
+
+func GetModuleLoaderServiceAccountName(mod kmmv1beta1.Module) string {
```
ybettan, line 78:
```suggestion
func GenerateModuleLoaderServiceAccountName(mod kmmv1beta1.Module) string {
```

internal/rbac/rbac.go
```diff
@@ -0,0 +1,84 @@
...
+	return mod.Name + "-module-loader"
+}
+
+func GetDevicePluginServiceAccountName(mod kmmv1beta1.Module) string {
```
ybettan, line 82:
```suggestion
func GenerateDevicePluginServiceAccountName(mod kmmv1beta1.Module) string {
```

internal/rbac/rbac.go
```diff
@@ -0,0 +1,84 @@
...
+			Name:      GetModuleLoaderServiceAccountName(mod),
+			Namespace: mod.Namespace,
+		},
+	}
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, rc.client, sa, func() error {
+		return controllerutil.SetControllerReference(&mod, sa, rc.scheme)
+	})
+	if err != nil {
+		return fmt.Errorf("cound not reconcile module loader ServiceAccount: %w", err)
```
ybettan, line 50:
```suggestion
		return fmt.Errorf("cound not create module-loader's ServiceAccount: %w", err)
```

internal/rbac/rbac.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	logger.Info("Reconciled module loader ServiceAccount", "name", sa.Name, "result", opRes)
```
ybettan, line 52:
```suggestion
	logger.Info("Created module-loader's ServiceAccount", "name", sa.Name, "result", opRes)
```

internal/rbac/rbac.go
```diff
@@ -0,0 +1,84 @@
...
+			Name:      GetDevicePluginServiceAccountName(mod),
+			Namespace: mod.Namespace,
+		},
+	}
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, rc.client, sa, func() error {
+		return controllerutil.SetControllerReference(&mod, sa, rc.scheme)
+	})
+	if err != nil {
+		return fmt.Errorf("cound not reconcile device plugin ServiceAccount: %w", err)
```
ybettan, line 71:
```suggestion
		return fmt.Errorf("cound not create device-plugin's ServiceAccount: %w", err)
```

controllers/module_reconciler.go
ybettan: Thanks for the explanation. It does make a lot of sense :)

controllers/module_reconciler.go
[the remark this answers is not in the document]
qbarrand: > Don't we want to default to the k8s default ServiceAccount (with a warning log) in this case?

That's not what the user asked, so I would definitely fail here.

internal/rbac/rbac.go
mresvanis: Makes total sense, I updated the PR accordingly :+1:

PR #91, opened 2022-09-28
add the fields for signing to crd

chr15p:
Add fields to the CRD to support kmod signing

This PR just adds the fields, the functionality to actually use them will come in a later PR. This is mostly intended for discussion, to ensure the actual code is built on foundations we agree on, and to try and keep the size of that future PR down to a reviewable and mergeable size (which my previous attempts have not been great at!) 

Edit: 
updated commit message to say

this add four fields to the crd in their own subsection:

FileList - a list of kmod files within the container to sign

KeySecret and CertSecret - these are the names of secrets containing the private and public keys respectivly that
will be used to sign the kmods. The public key should be the one added into the uefi moklist for secureboot,
 and KeySecret is its private component. (See the kernel docs e.g.
https://www.kernel.org/doc/html/v4.15/admin-guide/module-signing.html#manually-signing-modules
for details of this process)

unsignedImage - the optional name of a dtk image containing the unsigned knods, used only if their is no build stage.

The expected workflow will be (not implemented in this commit):
if there is only a kernelmapping.build section it will produce the image kernelmapping.containerImage
(for the non-secureboot case)

if there is both km.build and km.sign then build will produce an intermediate image and sign will
consume that to produce km.containerImage

if there is only km.sign signing will consumse km.sign.unsignedImage and produce km.containerImage
(this is for the case where a vendor supplies prebuilt images that need signing)

the generation of the intermediate image name will be handled automatically within the controller
(again in a different commit) but will probably be something like km.containerImage + "-unsigned"


ybettan:
@qbarrand We need to trigger the Github actions here as well. I don't think I have enough privileges to do it (can't see the `workflow` tab).

ybettan:
Can you please copy the content of of the PR description into the commit message?

api/v1beta1/module_types.go
```diff
@@ -81,6 +81,22 @@ type Build struct {
...
+type Sign struct {
+	// +optional
+	// Image to sign, ignored if a Build is present, required otherwise
+	UnsignedImage string `json:"unsignedImage,omitempty"`
+
+	// a secret containing the private key used to sign kernel modules for secureboot
+	KeySecret *v1.LocalObjectReference `json:"keySecret"`
+
+	// a secret containing the public key used to sign kernel modules for secureboot
+	CertSecret *v1.LocalObjectReference `json:"certSecret"`
```
ybettan, line 93: Why do we need a public key for signing purposes?

ybettan:
Few questions:
What is the expected flow here? Can we have it in the commit message (and PR description as well)? Are we pulling from the `unsignedImage`, signing, and then what? pushing somewhere? are we overriding the unsigned image? pushing a new tag? Do we need to mention the new tag name in the `Sign` object as we do for the `unsignedImage`?

chr15p:
written a better commit message that hopefully explains my thinking  (I've been staring at this stuff so long I forget it isn't obvious to other people! )

The signing process itself will use the sign-file binary distributed with the kernel source (see https://www.kernel.org/doc/html/v4.15/admin-guide/module-signing.html#manually-signing-modules) and that requires the public and private key, so the simple answer is we need both keys/certs as input parameters to that code.  We could write our own signing code but that seems fragile.

ybettan:
Thanks for updating the commit message.

> unsignedImage - the optional name of a dtk image containing the unsigned knods, used only if their is no build stage.

How DTK is related here? We don't have DTK u/s. I guess you meant "driver container" instead of "dtk"?!

Besides that, LGTM.

api/v1beta1/module_types.go
```diff
@@ -89,6 +105,10 @@ type KernelMapping struct {
 	// Build enables in-cluster builds for this mapping and allows overriding the Module's build settings.
 	Build *Build `json:"build"`
 
+	// +optional
+	// Sign enables in-cluster signing for this mapping
+	Sign *Sign `json:"sign"`
```
yevgeny-shnaidman, line 110: as said in the previous PR, i think that the SIGN section should be outside of Build

chr15p: I'm not sure I understand what you mean. In this PR it's alongside build, not in it, so we have `kernelmapping.build` and `kernelmapping.sign` not `kernelmapping.build.sign`.  That was how I interpreted your previosu comments but do you mean it should be outside of `kernelmapping` entirely and we only have one signing section for all kernel versions?

api/v1beta1/module_types.go
```diff
@@ -81,6 +81,22 @@ type Build struct {
...
+
+	// a secret containing the private key used to sign kernel modules for secureboot
+	KeySecret *v1.LocalObjectReference `json:"keySecret"`
+
+	// a secret containing the public key used to sign kernel modules for secureboot
+	CertSecret *v1.LocalObjectReference `json:"certSecret"`
+
+	// +optional
+	// paths inside the image for the kernel modules to sign (if ommited all kmods are signed)
+	FileList []string `json:"filesToSign,omitempty"`
```
qbarrand, line 97:
```suggestion
	FilesToSign []string `json:"filesToSign,omitempty"`
```
For consistency? or maybe the other way around?

api/v1beta1/module_types.go
qbarrand: I think this is much better. Shall we add `,omitempty` to the `json` tag?

qbarrand (commented):
Two small comments, lgtm otherwise. Thanks!

api/v1beta1/module_types.go
yevgeny-shnaidman: Do we really need this list? Can't we just sign all the .ko file under the appropriate directory (/usr/lib/modiles/<kernel>)?

chr15p: FilesToSign is clearer in the yaml than FileList so lets go with that.

For some reason I cant reply directly to your comment about having adding omitempty, I hadn't added it so it would match the attributes for the km.Build field. It makes sense though so I've force pushed a change with this and the FileList change.

qbarrand: > Do we really need this list? Can't we just sign all the .ko files under the appropriate directory (/usr/lib/modiles/)?

Follow-on issue: #94. Let's merge this for now.

PR #92, opened 2022-09-28
Include .git in image builds

qbarrand:
Explicitly include .git in Google Cloud Builds

To include VCS data in the controller binary, our image build process
requires .git to be present in the working directory.
Google Cloud Build, the system used by the Kubernetes project to build
container images, ignores .git by default.
Explicitly include .git in the GCB environment through a .gcloudignore
file.

Reference: https://github.com/GoogleCloudPlatform/cloud-builders/issues/236#issuecomment-374629200

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -1,0 +1,2 @@
+# Explicitly include .git in Google Cloud Builds as it is needed by Dockerfile.
+!.git


ybettan:
Can we use `create a .gcloudignore that explicitly includes .git (echo '!.git' > .gcloudignore)` instead? Seems more comprehensive to me.

Also, can you describe in the commit message (and PR description) why this change is required (we are using .git for getting the commit sha in our logs bla bla, prow need to clone `.git` as part of its process in order for the code to compile etc)?

It would be nice to get it directly instead of going to the link and trying to figure out what the author needed.

qbarrand:
@ybettan PTAL

ybettan:
Thank you!
/lgtm

PR #93, opened 2022-09-29
Draft: Troubleshoot image builds

qbarrand:
Debug only: do not merge.
/hold


ybettan:
Isn't the job will be triggered on commit merges? Do you expect this PR to trigger the build?

qbarrand:
I wanted to verify if `.git` is indeed present in Prow jobs. We can close this now.

PR #97, opened 2022-10-04
Add an extra label to the build job to label the stage of the build

chr15p:
This PR adds a third label to kubernetes job objects created by kmmo for future proofing.

Currently jobs are labelled the module.name and target-kernel, and these are used to search for the job to ensure its creation, completion etc.

The manger searching for these jobs has no other way of determining which job it should be managing.

If we add multiple jobs into the build pipeline (e.g. a signing job) the build manager has no way of distinguishing between the different jobs, so when it searches for completed jobs it will detect finished signing jobs as if they were finished build (kaniko) jobs, and vice versa. This would lead to much confusion and the pipeline not running through its stages correctly.

This commit adds a third "kmm.node.kubernetes.io/build-stage" label that is set to "build" to make it explicit that this is a build job. For signing jobs it should be set to "sign" and for any other future jobs it would be set to a descriptive name. The build job manager then searches for all 3 labels (and future job managers should also search for their own build-stage label) and ignores any jobs with a different build-stage set.

### DIFF
--- a/internal/build/job/manager.go
+++ b/internal/build/job/manager.go
@@ -37,6 +37,7 @@ func labels(mod kmmv1beta1.Module, targetKernel string) map[string]string {
 	return map[string]string{
 		constants.ModuleNameLabel:    mod.Name,
 		constants.TargetKernelTarget: targetKernel,
+		constants.JobType:            "build",
 	}
 }
 
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -5,4 +5,5 @@ const (
 	NodeLabelerFinalizer = "kmm.node.kubernetes.io/node-labeler"
 	TargetKernelTarget   = "kmm.node.kubernetes.io/target-kernel"
 	DaemonSetRole        = "kmm.node.kubernetes.io/role"
+	JobType              = "kmm.node.kubernetes.io/job-type"
 )
--- a/internal/build/job/maker_test.go
+++ b/internal/build/job/maker_test.go
@@ -73,6 +73,7 @@ var _ = Describe("MakeJobTemplate", func() {
 				Labels: map[string]string{
 					constants.ModuleNameLabel:    moduleName,
 					constants.TargetKernelTarget: kernelVersion,
+					constants.JobType:            "build",
 				},
 				OwnerReferences: []metav1.OwnerReference{
 					{


internal/constants/constants.go
```diff
@@ -5,4 +5,5 @@ const (
 	NodeLabelerFinalizer = "kmm.node.kubernetes.io/node-labeler"
 	TargetKernelTarget   = "kmm.node.kubernetes.io/target-kernel"
 	DaemonSetRole        = "kmm.node.kubernetes.io/role"
+	BuildStage           = "kmm.node.kubernetes.io/build-stage"
```
qbarrand, line 8: How about `kmm.node.kubernetes.io/job-type`?

chr15p: actually BuildStage is used as a variable name in the metrics package (it doesn't conflict, but will confuse), so I've changed everything to constants.JobType and `kmm.node.kubernetes.io/job-type`

ybettan:
@chr15p Can we please update the commit message: `build-stage` --> `build-type`?

internal/build/job/manager.go
```diff
@@ -37,6 +37,7 @@ func labels(mod kmmv1beta1.Module, targetKernel string) map[string]string {
 	return map[string]string{
 		constants.ModuleNameLabel:    mod.Name,
 		constants.TargetKernelTarget: targetKernel,
+		constants.JobType:            "build",
```
ybettan, line 40: Just because I am intrigued. How does it differ from https://github.com/kubernetes-sigs/kernel-module-management/blob/0c4df8dba30113323825ba1c12292880f88f6958/internal/build/job/maker.go#L126? Will it be easier to fetch jobs based on labels rather than on job's names?

ybettan: Seems like it is indeed easier to filter jobs based on labels cause it is supported by the client already.

chr15p: tldr: its easier because it's what we're already doing, and adding an extra label to match on is simpler than having to change the existing code :)

Longer answer: 
We only have the module name and the job-stage in GenerateName so we get the same GenerateName for different jobs building for different kernel versions. The jobs themselves are named  with GenerateName and a random 5 character string.

The job manager only manages a single job for a single `kernelmapping` Getting the job we want to manage would then need getting all the jobs that start with GenerateName and then somehow figuring out which one is the one we actually want to maange. Not impossible, but complicated and k8s has strong support for adding and searching by labels so I don't think searching by name would give us any advantages.

The only possible issue I can think of is you have two jobs running for the same module on the same kernel version when you'd have two jobs with the same labels returned, but they'd both have the same base name as well so a name search wouldn't improve this. (its an unlikely edge case that implies a problem anyway so which we currently handle getting multiple jobs back by throwing an error)

Issue #94, opened 2022-09-30
Consider signing all files in the module directory


qbarrand:
Do we really need this list? Can't we just sign all the .ko files under the appropriate directory (/usr/lib/modiles/<kernel>)?

_Originally posted by @yevgeny-shnaidman in https://github.com/kubernetes-sigs/kernel-module-management/pull/91#discussion_r983599080_

PR #101, opened 2022-10-05
add code/dockerfile for building the secureboot signing job image

chr15p:
This commit adds a stand alone binary (signimage.go) and an associated Dockerfile that together will provide the image for use in the signing job.

The dockerfile compiles signimage, and retrieves the sign-file binary from the kernel-devel package. sign-file is the binary provided by the kernel developers for signing kmods, and while its operation is not super-complex and could be handled within signimage.go, we use it to ensure maximum compatibility. 

The signimage.go binary takes an image name, pulls it down, loops through each file in each layer and if its a kernel modules extracts it to a temporary directory, and runs sign-file on it. All signed kmods are then added back as a (single) new layer, and the new image pushed back to the registry.

It is intended that this output image is the driver container named in km.containerImage

### DIFF
--- a/cmd/signimage/signimage.go
+++ b/cmd/signimage/signimage.go
@@ -1,0 +1,297 @@
+package main
+
+import (
+	"archive/tar"
+	"bytes"
+	"flag"
+	"fmt"
+	"github.com/docker/cli/cli/config"
+	dockertypes "github.com/docker/cli/cli/config/types"
+	"github.com/go-logr/logr"
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
+	"io"
+	"k8s.io/klog/v2/klogr"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func checkArg(arg *string, varname string, fallback string) {
+	if *arg == "" {
+		if fallback != "" {
+			*arg = fallback
+		} else {
+			fmt.Printf("%s not found:\n", varname)
+			flag.PrintDefaults()
+			os.Exit(0)
+		}
+	}
+}
+
+/*
+** Convert a relative path to an absolute path
+** filenames in image manifests are normally relative but can be "path/file" or "./path/file"
+** or occasionaly be absolute "/path/file" (in baselayers) depending on how they were created.
+** and we need to turn them into absolute paths for easy string comparisons against the
+** filesToSign list we've been passed from the CR via the cli
+** the easiest way to do this is force it to be abs, then clean it up
+ */
+func canonicalisePath(path string) string {
+	return filepath.Clean("/" + path)
+}
+
+func signFile(filename string, publickey string, privatekey string) error {
+	logger.Info("running /sign-file", "algo", "sha256", "privatekey", privatekey, "publickey", publickey, "filename", filepath.Base(filename))
+	out, err := exec.Command("/sign-file", "sha256", privatekey, publickey, filename).Output()
+	if err != nil {
+		return fmt.Errorf("signing %s returned: %s\n error: %v\n", filename, out, err)
+	}
+	return nil
+}
+
+func getAuthFromFile(configfile string, repo string) (authn.Authenticator, error) {
+
+	if configfile == "" {
+		logger.Info("no pull secret defined, default to Anonymous")
+		return authn.Anonymous, nil
+	}
+
+	f, err := os.Open(configfile)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+	cf, err := config.LoadFromReader(f)
+	if err != nil {
+		return nil, err
+	}
+
+	var cfg dockertypes.AuthConfig
+
+	cfg, err = cf.GetAuthConfig(repo)
+	if err != nil {
+		return nil, err
+	}
+
+	return authn.FromConfig(authn.AuthConfig{
+		Username:      cfg.Username,
+		Password:      cfg.Password,
+		Auth:          cfg.Auth,
+		IdentityToken: cfg.IdentityToken,
+		RegistryToken: cfg.RegistryToken,
+	}), nil
+
+}
+
+func die(exitval int, message string, err error) {
+	fmt.Fprintf(os.Stderr, "\n%s\n", message)
+	logger.Info("ERROR "+message, "err", err)
+	logger.Error(err, message)
+	os.Exit(exitval)
+}
+
+func processFile(filename string, header *tar.Header, tarreader io.Reader, data []interface{}) error {
+
+	registryObj := data[0].(registry.Registry)
+	extractionDir := data[1].(string)
+	filesList := data[2].(string)
+	privKeyFile := data[3].(string)
+	pubKeyFile := data[4].(string)
+	kmodsToSign := data[5].(map[string]string)
+
+	canonfilename := canonicalisePath(filename)
+
+	//either the kmod has not yet been found, or we didn't define a list to search for
+	if kmodsToSign[canonfilename] == "not found" ||
+		(filesList == "" &&
+			kmodsToSign[canonfilename] == "" &&
+			canonfilename[len(canonfilename)-3:] == ".ko") {
+
+		logger.Info("Found kmod", "kmod", canonfilename, "matches kmod in image", header.Name)
+		//its a file we wanted and haven't already seen
+		//extract to the local filesystem
+		err := registryObj.ExtractFileToFile(extractionDir+"/"+header.Name, header, tarreader)
+		if err != nil {
+			return err
+		}
+		kmodsToSign[canonfilename] = extractionDir + "/" + header.Name
+		logger.Info("Signing kmod", "kmod", canonfilename)
+
+		//sign it
+		err = signFile(kmodsToSign[canonfilename], pubKeyFile, privKeyFile)
+		if err != nil {
+			return fmt.Errorf("error signing file %s", canonfilename)
+		}
+		logger.Info("Signed successfully", "kmod", canonfilename)
+		return nil
+
+	}
+	return nil
+}
+
+func addFileToTarball(sourcename string, filename string, tarwriter *tar.Writer) error {
+	finfo, err := os.Stat(sourcename)
+	if err != nil {
+		return fmt.Errorf("failed to stat %s: %w", filename, err)
+	}
+
+	if finfo.IsDir() {
+		return nil
+	}
+	hdr := &tar.Header{
+		Name:     filename,
+		Mode:     int64(finfo.Mode()),
+		Typeflag: 0,
+		Size:     finfo.Size(),
+	}
+
+	if err := tarwriter.WriteHeader(hdr); err != nil {
+		return fmt.Errorf("failed to write tar header: %w", err)
+	}
+
+	f, err := os.Open(sourcename)
+	if err != nil {
+		return fmt.Errorf("failed to open file to add to the tar: %w", err)
+	}
+	defer f.Close()
+
+	if _, err := io.Copy(tarwriter, f); err != nil {
+		return fmt.Errorf("failed to read file into the tar: %w", err)
+	}
+
+	return nil
+}
+
+var logger logr.Logger
+
+func main() {
+	// get the env vars we are using for setup, or set some sensible defaults
+	var err error
+	var unsignedImageName string
+	var signedImageName string
+	var pullSecret string
+	var pushSecret string
+	var extractionDir string
+	var filesList string
+	var privKeyFile string
+	var pubKeyFile string
+	var nopush bool
+
+	logger = klogr.New()
+
+	flag.StringVar(&unsignedImageName, "unsignedimage", "", "name of the image to sign")
+	flag.StringVar(&signedImageName, "signedimage", "", "name of the signed image to produce")
+	flag.StringVar(&filesList, "filestosign", "", "colon seperated list of kmods to sign")
+	flag.StringVar(&privKeyFile, "key", "", "path to file containing private key for signing")
+	flag.StringVar(&pubKeyFile, "cert", "", "path to file containing public key for signing")
+	flag.StringVar(&pullSecret, "pullsecret", "", "path to file containing credentials for pulling images")
+	flag.StringVar(&pullSecret, "pushsecret", "", "path to file containing credentials for pushing images")
+	flag.BoolVar(&nopush, "no-push", false, "do not push the resulting image")
+
+	flag.Parse()
+
+	checkArg(&unsignedImageName, "unsignedimage", "")
+	checkArg(&signedImageName, "signedimage", unsignedImageName+"signed")
+	checkArg(&filesList, "filestosign", "")
+	checkArg(&privKeyFile, "key", "")
+	checkArg(&pubKeyFile, "cert", "")
+	checkArg(&pullSecret, "pullsecret", "")
+	checkArg(&pushSecret, "pushsecret", pullSecret)
+	// if we've made it this far the arguments are sane
+
+	// get a temp dir to copy kmods into for signing
+	extractionDir, err = os.MkdirTemp("/tmp/", "kmod_signer")
+	if err != nil {
+		die(1, "could not create temp dir", err)
+	}
+	defer os.RemoveAll(extractionDir)
+
+	// sets up a tar archive we will use for a new layer
+	var b bytes.Buffer
+	tarwriter := tar.NewWriter(&b)
+
+	//make a map of the files to sign so we can track what we want to sign
+	kmodsToSign := make(map[string]string)
+	for _, x := range strings.Split(filesList, ":") {
+		if canonicalisePath(x) != x {
+			err = fmt.Errorf("%s not an sbsolute path", x)
+			die(9, "paths for files to sign must be absolute", err)
+		}
+		kmodsToSign[x] = "not found"
+	}
+
+	a, err := getAuthFromFile(pullSecret, strings.Split(unsignedImageName, "/")[0])
+	if err != nil {
+		die(2, "failed to get auth", err)
+	}
+
+	r := registry.NewRegistry()
+
+	img, err := r.GetImageByName(unsignedImageName, a)
+	if err != nil {
+		die(3, "could not Image()", err)
+	}
+
+	logger.Info("Successfully pulled image", "image", unsignedImageName)
+	logger.Info("Looking for files", "filelist", strings.Replace(filesList, ":", " ", -1))
+
+	/*
+	** loop through all the layers in the image from the top down
+	 */
+	err = r.WalkFilesInImage(img, processFile, r, extractionDir, filesList, privKeyFile, pubKeyFile, kmodsToSign)
+	if err != nil {
+		die(9, "failed to search image", err)
+	}
+
+	/*
+	** check if we found everything, if not then explode
+	 */
+	missingKmods := 0
+	for k, v := range kmodsToSign {
+		if v == "not found" {
+			missingKmods = 1
+			logger.Info("Failed to find expected kmod", "kmod", k)
+		} else {
+			err := addFileToTarball(v, k, tarwriter)
+			if err != nil {
+				die(1, "failed to add signed kmods to tarball", nil)
+			}
+
+		}
+	}
+	if missingKmods != 0 {
+		die(4, "Failed to find all expected kmods", fmt.Errorf("Failed to find all expected kmods"))
+	}
+
+	outputTarFile := extractionDir + "/layerfile.tar"
+	err = os.WriteFile(outputTarFile, b.Bytes(), 0700)
+	if err != nil {
+		die(5, "failed to write layer to tarball", err)
+	}
+
+	//create a new image from our old image with our tarball as a new layer
+	signedImage, err := r.AddLayerToImage(outputTarFile, img)
+	if err != nil {
+		die(6, "failed to add layer to image", err)
+	}
+
+	logger.Info("Appended new layer to image", "image", signedImageName)
+
+	if !nopush {
+		a, err = getAuthFromFile(pushSecret, strings.Split(signedImageName, "/")[0])
+		if err != nil {
+			die(7, "failed to get push auth", err)
+		}
+
+		// write the image back to the name:tag set via the args
+		err := r.WriteImageByName(signedImageName, signedImage, a)
+		if err != nil {
+			die(8, "failed to write signed image", err)
+		}
+		// we're done successfully, so we need a nice friendly message to say that
+		logger.Info("Pushed image back to repo", "image", signedImageName)
+	}
+	os.Exit(0)
+}
--- a/internal/registry/registry.go
+++ b/internal/registry/registry.go
@@ -9,16 +9,25 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+
 	"github.com/google/go-containerregistry/pkg/v1/remote/transport"
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/auth"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/google/go-containerregistry/pkg/v1/types"
 )
 
 const (
@@ -44,6 +53,14 @@ type Registry interface {
 	VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool
 	GetLayersDigests(ctx context.Context, image string, po *kmmv1beta1.PullOptions, registryAuthGetter auth.RegistryAuthGetter) ([]string, *RepoPullConfig, error)
 	GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (v1.Layer, error)
+	WriteImageByName(imageName string, image v1.Image, auth authn.Authenticator) error
+	WalkFilesInImage(image v1.Image, fn func(filename string, header *tar.Header, tarreader io.Reader, data []interface{}) error, data ...interface{}) error
+	GetLayerMediaType(image v1.Image) (types.MediaType, error)
+	AddLayerToImage(tarfile string, image v1.Image) (v1.Image, error)
+	GetImageByName(imageName string, auth authn.Authenticator) (v1.Image, error)
+	ParseReference(imageName string) (name.Reference, error)
+	ExtractBytesFromTar(size int64, tarreader io.Reader) ([]byte, error)
+	ExtractFileToFile(destination string, header *tar.Header, tarreader io.Reader) error
 }
 
 type registry struct {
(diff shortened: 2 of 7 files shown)


signimage/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM registry.fedoraproject.org/fedora as ksource
+RUN yum install -y kernel-devel
+RUN cp /usr/src/kernels/*/scripts/sign-file /sign-file
```
ybettan, line 3: This layer seems redundant.

In addition to that, are you assuming anything regarding `/usr/src/kernels/*/scripts/sign-file`?
What happens if there are multiple kernels? are we going to copy multiple files? (When I run that container on my fedora host I don't have any kernel in `/usr/src/kernels/`.

signimage/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY go.sum go.sum
+RUN go mod download
+COPY signimage.go signimage.go
+# Build
+RUN go build signimage.go
+
+FROM registry.fedoraproject.org/fedora
+
+COPY --from=builder /workspace/signimage /
+COPY --from=ksource /sign-file /sign-file
```
ybettan, line 19: Any reason we are not using the binary from that layer instead of coping it from the previous identical layer?

signimage/Makefile
```diff
@@ -0,0 +1,34 @@
+IMAGE_TAG_BASE ?= quay.io/chrisp262/kmod-signer
```
ybettan, line 1: Do we plan to refer to your repo?

signimage/Makefile
```diff
@@ -0,0 +1,34 @@
...
+IMAGE_TAG ?= $(shell  git log --format="%H" -n 1)
```
ybettan, line 2: Why are we using a commit sha as the tag instead of always using `latest` for example?

signimage/Makefile
```diff
@@ -0,0 +1,34 @@
...
+
+# Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
+ifeq (,$(shell go env GOBIN))
+GOBIN=$(shell go env GOPATH)/bin
+else
+GOBIN=$(shell go env GOBIN)
+endif
+
+.PHONY: all
+all: signimage
```
ybettan, line 14:
```suggestion
all: sign-image
```

signimage/Makefile
```diff
@@ -0,0 +1,34 @@
...
+.PHONY: fmt
+fmt: ## Run go fmt against code.
+	go fmt ./...
+
+.PHONY: vet
+vet: ## Run go vet against code.
+	go vet ./...
+
+.PHONY:  signimage
+sign-image: ## Build sign-image binary.
```
ybettan, line 25:
```suggestion
signimage: ## Build sign-image binary.
```
Are you OK for having the same name in the target and in the PHONY?

signimage/README.md
```diff
@@ -0,0 +1,50 @@
+A utility to pull down an image, extract named kernel modules from it, sign them with the provided keys, and add them back in as a new layer, then upload that new image under a new tag.
+
+It operates as a wrapper around the sign-file binary provided as part of the kernel-devel package (a wrapper rather than a reimplementation to ensure its bug-for-bug compatabile, rather then having awhole new set of bugs of its own). sign-file is distributed as part of the kernel source so in theory is kernel version specific but in reality it hasn't changed to 5+ years, and for kernel modules to be whitelisted across major RHEL versions the signing format also has to be compatable so this is not a major concer.
+
+Configuration is done via command line switches or failing that via environment variables
```
ybettan, line 5: Why do we need both switches and env variables? Assuming that KMM's code is the only one consuming that binary, won't it be easier to maintain a single method of passing arguments?

signimage/README.md
```diff
@@ -0,0 +1,50 @@
...
+- PUSHSECRET  path to file containing pull credentials
+
+
+
+## Examples
+An example of its use as a Kubernetes job can be found in the ```kmod_signer_job.yaml``` file
+
+Or the from the command line
+```
+./signimage -unsignedimage quay.io/chrisp262/minimal-driver:procfsv1 \
```
ybettan, line 43: Should we use a generic image, for example `quay.io/<org>/<image>:<tag>`

signimage/go.mod
```diff
@@ -0,0 +1,9 @@
+module readkmod
```
ybettan, line 1: Any special reason this is not called `signimage`?

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	checkarg(&signedimagename, "SIGNEDIMAGE", unsignedimagename+"signed")
+	checkarg(&fileslist, "FILESTOSIGN", "")
+	checkarg(&privkeyfile, "KEYSECRET", "")
+	checkarg(&pubkeyfile, "CERTSECRET", "")
+	checkarg(&pullsecret, "PULLSECRET", "")
+	checkarg(&pushsecret, "PUSHSECRET", pullsecret)
+	// if we've made it this far the arguements are sane
+
+	// get a temp dir to copy kmods into for signing
+	extractiondir, err := os.MkdirTemp("/tmp/", "kmod_signer")
```
ybettan, line 209: From the method's doc:
```
It is the caller's responsibility to remove the directory when it is no longer needed.
```

Should we add a `defer` closing function?

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	if err != nil {
+		fmt.Errorf("could not create temp dir: %v\n", err)
+		panic(err)
```
ybettan, line 212:
```suggestion
		panic(fmt.Errorf("could not create temp dir: %w", err))
```

signimage/README.md
```diff
@@ -0,0 +1,50 @@
...
+  -key string
+        path to file containing private key for signing
+  -pullsecret string
+        path to file containing credentials for pulling images
+  -pushsecret string
+        path to file containing credentials for pushing images (defaults to the pushsecret)
+  -signedimage string
+        name of the signed image to produce (defaults to "${unsignedimage}-signed")
+  -unsignedimage string
+        name of the image to sign
```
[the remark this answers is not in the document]
qbarrand: Those work as well with `flag`.

signimage/Makefile
qbarrand: I don't think we need new Makefile targets. [`cloudbuild.yaml`](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/cloudbuild.yaml), where we will be adding instructions for this new image, already doesn't use them.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	var b bytes.Buffer
+	tarwriter := tar.NewWriter(&b)
+
+	//make a map of the files to sign so we can track what we want to sign
+	kmodstosign := make(map[string]string)
+	for _, x := range strings.Split(fileslist, ":") {
+		kmodstosign[x] = "not found"
+	}
+
+	// set up image download otions
```
ybettan, line 225:
```suggestion
	// set up image download options
```

signimage/Makefile
qbarrand: That will also be automated in the postsubmit job. We get both `latest` and the SHA.

signimage/README.md
qbarrand: > Why do we need both switches and env variables? Assuming that KMM's code is the only one consuming that binary, won't it be easier to maintain a single method of passing arguments?

Agreed.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	opts := make([]name.Option, 0)
+	ref, err := name.ParseReference(unsignedimagename, opts...)
```
ybettan, line 227:
```suggestion
	ref, err := name.ParseReference(unsignedimagename, []name.Option{})
```

I think it is clearer to make sure no one else is using `opts`.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	signedimage, err := mutate.AppendLayers(img, signedlayer)
+	if err != nil {
+		fmt.Errorf("failed to append layer: %v", err)
+	}
+
+	fmt.Printf("\n== Appended new layer to image\n")
+
+	if nopush == false {
+		// write the image back to the name:tag set via the envvars
+		signedref, err := name.ParseReference(signedimagename, opts...)
```
ybettan, line 337:
```suggestion
		signedref, err := name.ParseReference(signedimagename, []name.Option{})
```

I think it is clearer to make sure no one else is using `opts`.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
+package main
```
qbarrand, line 1: I think we need unit tests for this software.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	for _, x := range strings.Split(fileslist, ":") {
+		kmodstosign[x] = "not found"
+	}
+
+	// set up image download otions
+	opts := make([]name.Option, 0)
+	ref, err := name.ParseReference(unsignedimagename, opts...)
+	if err != nil {
+		fmt.Errorf("could not parse the container image name: %v\n", err)
+		panic(err)
```
ybettan, line 230: What does that line do? it is just returning a new error but no one is using it. Am I missing something?
```suggestion
		panic(fmt.Errorf("could not parse the container image name: %w", err))
```

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	a, err := getAuthFromFile(pullsecret, strings.Split(unsignedimagename, "/")[0])
+	if err != nil {
+		fmt.Printf("failed to get auth: %v\n", err)
+		panic(err)
```
ybettan, line 236: Same with the error.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	descriptor, err := remote.Get(ref, remote.WithAuth(a))
+	if err != nil {
+		te := &transport.Error{}
+
+		if errors.As(err, &te) && te.StatusCode == http.StatusNotFound {
+			fmt.Errorf("could not get image: http.StatusNotFound")
```
ybettan, line 244: Who uses the return value of that?

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+
+		fmt.Errorf("could not get image: %v", err)
```
ybettan, line 247: Who uses the return value of that?

signimage/signimage.go
qbarrand, line 230: We should not use `panic` and return errors instead.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	img, err := descriptor.Image()
+	if err != nil {
+		fmt.Errorf("could not Image(): %v", err)
+		panic(err)
+	}
+	fmt.Printf("\n== Successfully pulled image: %s\n", unsignedimagename)
+	fmt.Printf("\n== Looking for files: %s\n", strings.Replace(fileslist, ":", " ", -1))
+
+	/*
+	** loop through all the layers in the image from the top down
```
ybettan, line 260: Why do we need to loop over all the layers? Isn't inspecting the last layer enough?

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	 */
+	layers, _ := img.Layers()
+	for i := len(layers) - 1; i >= 0; i-- {
+		fmt.Printf("== Searching layer %d\n", i)
```
ybettan, line 264: Should we use logs?

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+		currentlayer := layers[i]
+		layerreader, err := currentlayer.Uncompressed()
+		if err != nil {
+			fmt.Errorf("could not get layer: %v", err)
```
ybettan, line 268: Who uses the return value?

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+		panic(err)
+	}
+
+	img, err := descriptor.Image()
+	if err != nil {
+		fmt.Errorf("could not Image(): %v", err)
+		panic(err)
+	}
+	fmt.Printf("\n== Successfully pulled image: %s\n", unsignedimagename)
+	fmt.Printf("\n== Looking for files: %s\n", strings.Replace(fileslist, ":", " ", -1))
```
qbarrand, line 257: Could we use a logging framework here? [logr](https://github.com/go-logr/logr) as the facade and [klog](https://github.com/kubernetes/klog) as the implementation work well in the community.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+			fmt.Errorf("could not get layer: %v", err)
+			panic(err)
+		}
+
+		/*
+		** loop through all the files in the layer
+		 */
+		tarreader := tar.NewReader(layerreader)
+		for {
+			header, _ := tarreader.Next()
```
qbarrand, line 277: We should not ignore the error here.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+			fmt.Printf("Failed to find expected kmod %s \n", k)
+		}
+	}
+	if missingkmods != 0 {
+		fmt.Printf("Failed to find all expected kmods\n")
+		os.Exit(1)
+	}
+
+	//turn our tar archive into a layer
+	mediatype, _ := layers[len(layers)-1].MediaType()
```
qbarrand, line 321: We should not ignore the error here.

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+		panic(err)
+	}
+
+	a, err := getAuthFromFile(pullsecret, strings.Split(unsignedimagename, "/")[0])
+	if err != nil {
+		fmt.Printf("failed to get auth: %v\n", err)
+		panic(err)
+	}
+
+	descriptor, err := remote.Get(ref, remote.WithAuth(a))
```
yevgeny-shnaidman, line 239: We already have a registry package that has API for accessing the image, get layers of the image and checking if a file is present in the image. Maybe we can use it ( with some additional APIs). This will probably save a lot of code lines that you have here

signimage/Dockerfile
chr15p: we could install kernel-devel in this layer, but its huge (80MB plus dependencies) and we only need sign-file (15760 bytes) from it, so we do it this way to try and keep the resultant image as small as possible.

signimage/Makefile
chr15p: Again the Makefile is about making debugging easier so it uses the git commit because you probably want latest to reference the latest production version. As Quentin says kmmo itself should never reference it but should use the official image.

signimage/Makefile
chr15p: yes but I'm not sure that rule should be PHONY at all, it does produce a file.... I'll have a think and fix it either way..

signimage/Dockerfile
yevgeny-shnaidman: can we just use any image, and not fedora? for example alpine?

signimage/README.md
chr15p: --help works and produces a usage message, its provided automagically by the flag package.

I'm not assuming this is only used by KMM I think it should at least be possible to use it standalone in some non-KMM build pipeline and supporting environment variables only really adds a couple of lines and actually lets you create defaults taht you can then override with args, so I think the flexability is worth it.

signimage/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Examples
+An example of its use as a Kubernetes job can be found in the ```kmod_signer_job.yaml``` file
+
+Or the from the command line
+```
+./signimage -unsignedimage quay.io/chrisp262/minimal-driver:procfsv1 \
+	-pullsecret pullsecret \
+	-key mykey.priv \
+	-cert mykey.der \
+	-filestosign /modules/simple-kmod.ko:/modules/simple-procfs-kmod.ko
```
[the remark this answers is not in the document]
chr15p: if the image name is generic these should probably be generic as well rather than simple-kmod or kmm-mod e.g </var/lib/kmod1.ko>[:</var/lib/kmod2.ko>]...

signimage/signimage.go
chr15p: its running in a containers ephemeral layer so it will get destroyed when that does, but it's always good to tidy up...  I will add it

signimage/signimage.go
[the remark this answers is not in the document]
chr15p: no, its running in its own container so stdout is being captured by k8s automatically so these Printf statements are available via kubectl logs <pod>, messages from this don't end up in the KMM controller logs so dont need the same format

signimage/signimage.go
chr15p: that depends on if you've put all your kmods in the last layer :)
kvc-simple-kmod for example uses two seperate COPY statements so puts its two kmods in two different layers

signimage/signimage.go
chr15p: We could but I don't see what benefit it would add, this is a pretty simple batch job, that runs in its own container, and then exits. It's doing one thing and (hopefully) doing it well.

The complexity of using a full blown logging framework just so I can dump a handful of progress lines to stdout feels excessive to me, great for long running stuff like controllers, overkill for this.

signimage/signimage.go
chr15p: I disagree.

Firstly:  This is a standalone piece of code intended to run in a job that (may or may not be) run from KMMO. It is really only is in this repo at all because it need to be somewhere, and it not being in my personal github repo makes some external political things easier. The idea of closely coupling this code with that of the controller seems unnecessary. 

Secondly: After a quick look, while Registry superficially works in the same space, it would need significant changes to it make it usable, anything that references kmmv1beta1 would need to be removed, signimage does not have access to the crd, nor does it need it, but Registry refers to it extensively. 

Thirdly: Checking if a single image is present in the top layer is not enough here, we need to find all kmods (issue #94), in all layers (the code to loop through layers isn't in Registry), change them (not in Registry), add them to a new layer (not in Registry), write the image back (not in Registry). So there's a lot of functionality registry is missing we'd need to add.

So as it stands there is not much overlap between my code and Registry so it would need to be pretty much rewritten from scratch.

yevgeny-shnaidman: actually, i think that the work to be done is not that big:
1) since your code is in this repo, why not the already available package? If, sometime in the future,  we decide that your code needs to be removed to a different repo, we can simply make registry a public package
2) registry is only using the PullOption struct of the kmmv1beta, it does not receive a CRD as a parameter anywhere. You will need to configure your access to the registry, so why not use kmmv1beta structs, as opposed to using structs defined in other packages?
3) we will deal with finding all kmod later ( your code right now also does not handle it). You main function can be describe as:

-  get the image
-  get all layers of image
-  loop over each layer and check if a file exist
-  in case file exists - get it, sign and add it
So we can add extraction function to the package and also add function, and use the existing function. I think is will significantly reduce your code and enhance our package, which is a good thing. The loop can stay in the main, just as it is not in preflight package, for example

chr15p: > we will deal with finding all kmod later ( your code right now also does not handle it).

yes it does, somewhere around line 284 
`if kmodstosign[canonfilename] == "not found" || (fileslist == "" && kmodstosign[canonfilename] == "") {`

this checks if the file has been passed in via filelist (that then populates kmodstosign),  or if filelist is empty and we haven't seen this kmod before, then sign anyway. 

(this is why Sign.FilesToSign in the api was set as optional in the crd :) )

I'll have a look at the rest of it tomorrow

signimage/signimage.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+	defer os.Remove(extractiondir)
+
+	// sets up a tar archive we will use for a new layer
+	var b bytes.Buffer
+	tarwriter := tar.NewWriter(&b)
+
+	//make a map of the files to sign so we can track what we want to sign
+	kmodstosign := make(map[string]string)
+	for _, x := range strings.Split(fileslist, ":") {
```
yevgeny-shnaidman, line 223: how do we expect to get the list of .ko files to sign? Will we expect only the files' names (e.i. simple-kmod.ko) or the whole path ( /lib/modules/kernel/simple-kmod.ko)?

signimage/signimage.go
```diff
@@ -0,0 +1,352 @@
...
+		for {
+			header, err := tarreader.Next()
+			if err == io.EOF || header == nil {
+				break // End of archive
+			}
+
+			// paths in a layer are relative, and supplied paths are absolute so canonicalise
+			canonfilename := canonicalisePath(header.Name)
+			//either the kmod has not yet been found, or we didn't define a list to search for
+			if kmodstosign[canonfilename] == "not found" || (fileslist == "" && kmodstosign[canonfilename] == "") {
```
yevgeny-shnaidman, line 284: i might be missing something, but it seems that in case of an empty filelist, this code will be signing every file/directory in the layer,no?

signimage/signimage.go
```diff
@@ -0,0 +1,352 @@
...
+				flag.PrintDefaults()
+				os.Exit(0)
+			}
+		}
+	}
+	return
+}
+
+func canonicalisePath(path string) string {
+	canonical := strings.Replace("/"+path, "/./", "/", -1)
```
yevgeny-shnaidman, line 54: since the files in the layer are in relative path, isn't it enough just to add a "/" to the start of the path of the file?

signimage/signimage.go
qbarrand: Yes, it might be useful during development where `/tmp` has a slightly longer lifetime.

signimage/signimage.go
qbarrand: The complexity of setting up `klog` together with `logr` is moderate: https://github.com/kubernetes-sigs/kernel-module-management/blob/main/main.go#L78-L82
It would give us fields, levelled logging, the ability to mock the logger in unit tests, and more.
This is also the [preferred approach in Kubernetes](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#logging-in-kubernetes):
> Kubernetes project uses [klog](https://github.com/kubernetes/klog) for logging.

signimage/signimage.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+		offset += rc
+		if err == io.EOF {
+			break
+		}
+	}
+	dirname := filepath.Dir(header.Name)
+	err := os.MkdirAll(tmpdir+"/"+dirname, 0750)
+	if err != nil {
+		die(2, "could not create tempdir for kmod", err)
```
qbarrand, line 75: I would not call `os.Exit` from anywhere except the `main` function. The rest of the code should return errors to be processed by `main`.

chr15p:
I've done some major restructuring based on comments.
- moved to using klogr for printing to stdout
- moved a lot of the mainloop functionality to helper functions
- moved those helper functions to a library

I tried to put the helper functions into internal/registry/registry.go but that has a bunch of practical issues, libraries in internal/ are only available to things in directories above them so binaries in signimge/ are prohibited from seeing them. You can work around this for compilation with a bit of Makefile cleverness, but controller-gen (as used in `make generate`) really doesn't like it   (best option I could find to get it to ignore the directory ran into https://github.com/kubernetes-sigs/controller-tools/issues/684 )

So for the moment I have my own registry.go, if we can work around the controller-gen issue  we should be able to merge signimage/registry.go and internal/registry/registry.go easily.

qbarrand:
@chr15p having these files as separate Go module make them unable to import the KMM libraries indeed.
Moving `signimage.go` to a directory under `cmd` (such as `cmd/sign/signimage.go` for instance) would eliminate that problem by including the new code into the existing module that we have for KMM. Is there any downside in doing that?

chr15p:
I have no issue with moving it somewhere but I still cant get it to work

> [cprocter@toolbox kernel-module-management-upstream]$ ls cmd/signimage/signimage.go                                                                                                                              
cmd/signimage/signimage.go
> [cprocter@toolbox kernel-module-management-upstream]$ go build cmd/signimage/signimage.go                                                                                                                        
> [cprocter@toolbox kernel-module-management-upstream]$ make generate
> go install github.com/golang/mock/mockgen@v1.6.0
> /home/cprocter/engineering/kernel-module-management-upstream/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./..."                                                                         
> signimage.go:18:2: use of internal package github.com/kubernetes-sigs/kernel-module-management/internal/registry not allowed                                                                                      
> Error: not all generators ran successfully
> run `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -w` to see all available markers, or `controller-gen object:headerFile=hack/boilerplate.go.txt paths=./... -h` for usage                
>make: *** [Makefile:99: generate] Error 1

This is similar to the error when I compile signimage.go from within cmd/signimage/   so it feels like controller-gen is recursing into the directory and doing its thing.

I've tried moving it to other places as well and same result :(

signimage/Dockerfile
ybettan: Thanks for the explanation.

Regarding the first part, I mean that you have 2 identical layers, so we can combine them (2 layers instead of 3 in total).
You override it so I don't have the original file to see.

signimage/README.md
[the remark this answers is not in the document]
ybettan: Indeed.

signimage/Dockerfile
chr15p: you're right, that `COPY  cp /usr/src/kernels/*/scripts/sign-file /sign-file` is redundant, I'd not spotted that. I have removed it.

signimage/Dockerfile
chr15p: its a little trickier than that, in that sign-file needs libcrypto, glibc etc, whiich is why the first and third layers are the same base image. In theory we could move to alpine for both,  I'm using fedora because it avoid a bunch of repo and entitlement related issues I had when using ubi-minimal. I dont have enough knowledge of alpine to get sign-image and its deps installed reliably, but theres nothing to stop it being changed in the future.

signimage/signimage.go
chr15p: we're expecting the full paths with colon separators, e.g.  
`-filestosign /opt/lib/modules/4.18.0-348.2.1.el8_5.x86_64/simple-kmod.ko:/opt/lib/modules/4.18.0-348.2.1.el8_5.x86_64/simple-procfs-kmod.ko`
we could change it to be just the filename but I'm not sure that really adds anything, and prevents the (admittedly edge) use case of a single signed-image supporting multiple versions of the same kmod each signed for a different key.

signimage/signimage.go
chr15p: yep, that was a bug that crept in while I was debugging something else :)  should be fixed now

internal/registry/registry.go
```diff
@@ -245,3 +259,124 @@ func (r *registry) getImageDigestFromMultiImage(manifestListStream []byte) (stri
...
+
+/*
+** generic function to loop through all the files in an image and run
+** a function on them, based loosly on ftw() or filepath.Walk().
+** image  - the image to examine
+** fn	  - the function to call on each file
+** data[] - an array of strings to pass through to the helper function
+**        to provide any additional data needed.
+ */
+func (r *registry) WalkFilesInImage(image v1.Image, fn func(filename string, header *tar.Header, tarreader io.Reader, data []string) error, data []string) error {
```
yevgeny-shnaidman, line 351: This function walk on all the files in all the layers. We are only interested in the file under specific directory. So maybe we should add an option to this function or to processing function. WDYT?

signimage/README.md
chr15p: If we're only going for one then switches makes more sense, I just like the envars as fallback options. However this is not red line for me, and the discussion is taking more effort than they're worth so I've removed them.  If anyone cares they can always be added back in later.

internal/registry/registry.go
```diff
@@ -44,6 +52,12 @@ type Registry interface {
 	VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool
 	GetLayersDigests(ctx context.Context, image string, po *kmmv1beta1.PullOptions, registryAuthGetter auth.RegistryAuthGetter) ([]string, *RepoPullConfig, error)
 	GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (v1.Layer, error)
+	WriteImageByName(imageName string, image v1.Image, auth authn.Authenticator) error
+	WalkFilesInImage(image v1.Image, fn func(filename string, header *tar.Header, tarreader io.Reader, data []string) error, data []string) error
```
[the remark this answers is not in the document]
chr15p: Done.
The array of strings was an attempt to make this a general purpose function but your way is much cleaner.

internal/registry/registry.go
chr15p: KMMO is only interested in the files under specific directory, but I am trying to keep this image as generic as possible so it is still usable outside KMMO. That means we can't assume that all the kmods are in a single directory, so it would have to be an array (or similar) of directories and that adds a lot more complication.  Plus then future users may have different requirements.

The signimage job currently runs through in a few tens of seconds, so I don't really think there's much need to optomise this function for performance at the moment.

So, maybe in the future, but I dont think we need to do that in this PR.

signimage/go.mod
chr15p: no just a left over from an older name, but this go.mod hass been removed so fixed by default :)

signimage/signimage.go
```diff
@@ -0,0 +1,357 @@
...
+	}), nil
+
+}
+
+func main() {
+	// get the env vars we are using for setup, or set some sensible defaults
+
+	var unsignedimagename string
+	var signedimagename string
+	var fileslist string
```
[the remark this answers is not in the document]
chr15p: think I have caught all of these let me know if you spot one I've missed...

signimage/signimage.go
chr15p: Lots of the generic code has been moved into internal/registry/registry.go and the binary has moved into cmd/signimage so tI think that resolves (or at least outdates) this

signimage/signimage.go
chr15p: The files in a layer are normally relative but I think their exact format depends how exactly they were created. They are sometimes stored as  'path/` and sometimes `./path/` and sometimes aren't relative at all but are absolute (I suspect this is in baselayers) which become `//path` with a "/" appended. 

In the CRD we are asking for absolute paths, so this function is just doing some search and replace to clean up those patterns which then allows us to do string comparison on the paths to easily check if the file we have in the image is one to sign

signimage/signimage.go
chr15p: I'm leaving the die() helper function calls in main() because to unroll them would result in lots of repetitive logging/exit code, but you're right calling it from many layers down the stack in the image walking code is pretty nasty. Fixed.

signimage/Makefile
chr15p: The Makefile has been merged into the main Makefile so marking as resolved

cmd/signimage/signimage.go
```diff
@@ -0,0 +1,309 @@
...
+		} else {
+			fmt.Printf("%s not found:\n", varname)
+			flag.PrintDefaults()
+			os.Exit(0)
+		}
+	}
+	return
+}
+
+func canonicalisePath(path string) string {
```
yevgeny-shnaidman, line 34:
```suggestion
func canonicalisePath(path string) string {
       filepath.IsAbs(path) {
           return path
       }
       return "/" + filepath.Clean(path)
}
       
```

yevgeny-shnaidman: Also, lets add a comment that explains all the various filepaths scenarios that can be defined by the layet

cmd/signimage/signimage.go
```diff
@@ -0,0 +1,309 @@
...
+	canonical := strings.Replace("/"+path, "/./", "/", -1)
+	return strings.Replace(canonical, "//", "/", -1)
+}
+
+func extractFile(tmpdir string, header *tar.Header, tarreader io.Reader) (string, error) {
```
yevgeny-shnaidman, line 39: why not make this function part of registry package?

cmd/signimage/signimage.go
chr15p: I've pushed a variant of this (appending the "/" needs to be inside the clean, also I think its good to canonicalise absolutepaths just in case they have "./" etc in them), I've also changed it so the filestosign list gets canonicalised as well, just in case someone doesn't read the instructions and tries to sign something with a relative path, or adds "../" type things to the path.

yevgeny-shnaidman: Kudos for your version, it is shorter :). Regarding filestosign: do we want to try to correct users' mistakes? What if by trying to correct their mistake we will actually sign a file in an incorrect path? I think that in case user has provided incorrect path, we can just print a warning message , and that's it. WDYT?

cmd/signimage/signimage.go
chr15p: We'd end up with signimage calling into registry which calls back into signimage, which in turn calls back into registry, which feels complicated.

But then additionally extractFile is pretty much tailored for the signimage use case (it expects io.Reader to be lined up ready to read the first file, then extracts to a specific place in /tmp that maintains the path information so it can be added to the new layer, and creates things with specific permissions), some of this could be factored out into registry but its not generic enough to all go in there, 

So the code is simple if its in signimage but moving those few lines to registry makes it a lot more complex.

cmd/signimage/signimage.go
chr15p: I used to write perl as a job, shorter is not always better!

I agree about chance for signing the wrong file, I think its very unlikely to ever happen but its cleaner just to put out an error message and let the user fix their problem rather than take the risk. I have updated the code.

cmd/signimage/signimage.go
yevgeny-shnaidman: besides persmissions, which can be passed as a parameter, i don't see anything sign-related code in this function. io.Reader can be set to whatever offset user want prior to calling function, and extracting file to user-defined directory also seems ok for general use case

chr15p: Ok, I've moved that functionality to the registry library. I think it makes the signimage code worse because it has to pass the registry object into itself, so it can be passed back to signimage, so it can call functions in registry, but I guess if anyone else ever uses the registry library for extracting files it will help them.

PR #118, opened 2022-10-19
Modify Makefile in kmm-kmod

enriquebelarte:
Modify Makefile to use running kernel version if KERNEL_SRC_DIR is not set.

### DIFF
--- a/ci/kmm-kmod/Makefile
+++ b/ci/kmm-kmod/Makefile
@@ -1,4 +1,7 @@
-KERNEL_SRC_DIR =
+ifndef KERNEL_SRC_DIR
+KERNEL_SRC_DIR =/lib/modules/$(shell uname -r)/build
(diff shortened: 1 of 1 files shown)


ci/kmm-kmod/Makefile
```diff
@@ -1,4 +1,11 @@
-KERNEL_SRC_DIR =
+ifndef KVER
+KVER=$(shell uname -r)
+endif
+
+ifndef KERNEL_SRC_DIR
+KERNEL_SRC_DIR =/lib/modules/$(KVER)/build
+endif
+
```
qbarrand, line 8:
```suggestion
ifndef KERNEL_SRC_DIR
KERNEL_SRC_DIR =/lib/modules/$(shell uname -r)/build
endif

```

PR #121, opened 2022-10-20
add internal/sign library packages for signer job

chr15p:
This PR adds library support for creating and managing k8s jobs that will back the Sign API. It adds the internal/sign directory structure including three types
signer - creates the job template
signJobManager - creates, manages, and reports the completion of, the job.
helper - calculates the effective Sign configuration from the kerrnelmapping and defaults.

This closely follows the existing internal/build (kaniko) implementation, mainly because it essentially does the same thing (run a job external to the controller) and is expected to be run by the reconcile loop in a very similar fashion. It's separate so kaniko support can be replaced or changed without consequences.

This commit only adds libraries, it does not add controller support, or any code to the reconcile loop.  km.Sign will still not work after this commit, but this brings us another step closer.
Controller support is the final piece needed and will be added in a separate commit.


internal/sign/helper.go
```diff
@@ -0,0 +1,47 @@
...
+type helper struct{}
+
+func NewSignerHelper() Helper {
+	return &helper{}
+}
+
+func (m *helper) GetRelevantSign(mod kmmv1beta1.Module, km kmmv1beta1.KernelMapping) *kmmv1beta1.Sign {
+
+	if mod.Spec.ModuleLoader.Container.Sign == nil {
+		// km.Build cannot be nil in case mod.Build is nil, checked above
```
qbarrand, line 22:
```suggestion
		// km.Sign cannot be nil in case mod.Sign is nil, checked above
```

internal/sign/job/manager.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("expected 0 or 1 job, got %d", n)
+	}
+
+	return &jobList.Items[0], nil
+}
+
+func (jbm *signJobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, previousImage string, targetImage string, pushImage bool) (sign.Result, error) {
+	logger := log.FromContext(ctx)
+
+	logger.Info("Building in-cluster")
```
qbarrand, line 68: Do we want to remove that line?

internal/sign/job/manager.go
```diff
@@ -0,0 +1,133 @@
...
+		return sign.Result{Status: sign.StatusCreated, Requeue: true}, nil
+	}
+
+	changed, err := jbm.isJobChanged(job, jobTemplate)
+	if err != nil {
+		return sign.Result{}, fmt.Errorf("could not determine if job has changed: %v", err)
+	}
+
+	if changed {
+		logger.Info("The module's build spec has been changed, deleting the current job so a new one can be created", "name", job.Name)
```
qbarrand, line 98:
```suggestion
		logger.Info("The module's sign spec has been changed, deleting the current job so a new one can be created", "name", job.Name)
```

internal/sign/job/manager.go
```diff
@@ -0,0 +1,133 @@
...
+		opts := []client.DeleteOption{
+			client.PropagationPolicy(metav1.DeletePropagationBackground),
+		}
+		err = jbm.client.Delete(ctx, job, opts...)
+		if err != nil {
+			logger.Info(utils.WarnString(fmt.Sprintf("failed to delete build job %s: %v", job.Name, err)))
```
qbarrand, line 104:
```suggestion
			logger.Info(utils.WarnString(fmt.Sprintf("failed to delete signing job %s: %v", job.Name, err)))
```

internal/sign/manager.go
```diff
@@ -0,0 +1,26 @@
...
+
+type Result struct {
+	Requeue bool
+	Status  Status
+}
+
+//go:generate mockgen -source=manager.go -package=sign -destination=mock_manager.go
+
+type SignManager interface {
+	Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, previousImage string, targetImage string, pushImage bool) (Result, error)
```
qbarrand, line 25: What is `previousImage`?

internal/sign/job/signer.go
```diff
@@ -0,0 +1,138 @@
...
+		args = append(args, "-filestosign", strings.Join(signConfig.FilesToSign, ":"))
+	}
+
+	volumes := []v1.Volume{}
+	volumeMounts := []v1.VolumeMount{}
+	volumes = append(volumes, m.makeImageSigningSecretVolume(signConfig.KeySecret, "key", "key.priv"))
+	volumes = append(volumes, m.makeImageSigningSecretVolume(signConfig.CertSecret, "cert", "public.der"))
+
+	volumeMounts = append(volumeMounts, m.makeImageSecretVolumeMount(signConfig.CertSecret, "/signingcert"))
+	volumeMounts = append(volumeMounts, m.makeImageSecretVolumeMount(signConfig.KeySecret, "/signingkey"))
```
qbarrand, line 61: Nit:
```suggestion
	volumes := []v1.Volume{
		m.makeImageSigningSecretVolume(signConfig.KeySecret, "key", "key.priv"),
		m.makeImageSigningSecretVolume(signConfig.CertSecret, "cert", "public.der"),
	}

	volumeMounts := []v1.VolumeMount{
		m.makeImageSecretVolumeMount(signConfig.CertSecret, "/signingcert"),
		m.makeImageSecretVolumeMount(signConfig.KeySecret, "/signingkey"),
	}
```

internal/sign/job/signer.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	if err := controllerutil.SetControllerReference(&mod, job, m.scheme); err != nil {
+		return nil, fmt.Errorf("could not set the owner reference: %v", err)
+	}
+
+	return job, nil
+}
+
+func (m *signer) makeImageSigningSecretVolume(secretRef *v1.LocalObjectReference, key string, path string) v1.Volume {
```
qbarrand, line 102: Do `makeImageSigningSecretVolume`, `makeImageSecretVolumeMount` and `volumeNameFromSecretRef` need to be `Signer` private methods? I think they could just be standalone functions?

internal/sign/suite_test.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"testing"
+
+	"github.com/kubernetes-sigs/kernel-module-management/internal/test"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+var scheme *runtime.Scheme
```
qbarrand, line 12: I don't think this variable is used in this package.

internal/sign/helper_test.go
```diff
@@ -0,0 +1,174 @@
...
+	)
+
+	BeforeEach(func() {
+		ctrl = gomock.NewController(GinkgoT())
+		h = NewSignerHelper()
+	})
+
+	AfterEach(func() {
+		ctrl.Finish()
+	})
```
qbarrand, line 34: I think the gomock controller does that automatically - no need to call `AfterEach` manually.

internal/sign/job/manager.go
yevgeny-shnaidman: I think it is better to keep it - signifies start of building flow, it will help us during debugs

internal/sign/job/manager.go
```diff
@@ -0,0 +1,133 @@
...
+	if n := len(jobList.Items); n == 0 {
+		return nil, errNoMatchingBuild
+	} else if n > 1 {
+		return nil, fmt.Errorf("expected 0 or 1 job, got %d", n)
+	}
+
+	return &jobList.Items[0], nil
+}
+
+func (jbm *signJobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, previousImage string, targetImage string, pushImage bool) (sign.Result, error) {
```
yevgeny-shnaidman, line 65: This function is almost exact copy of Sync function in internal/build/job/manager.go. Maybe we can move the bulk of the code to a common library, and both Sync functions will call it

internal/sign/job/manager.go
```diff
@@ -0,0 +1,133 @@
...
+	case job.Status.Active == 1:
+		return sign.Result{Status: sign.StatusInProgress, Requeue: true}, nil
+	case job.Status.Failed == 1:
+		return sign.Result{}, fmt.Errorf("job failed: %v", err)
+	default:
+		return sign.Result{}, fmt.Errorf("unknown status: %v", job.Status)
+	}
+}
+
+func (jbm *signJobManager) isJobChanged(existingJob *batchv1.Job, newJob *batchv1.Job) (bool, error) {
```
yevgeny-shnaidman, line 123: This function is an exact copy of the on in the internal/build/job/manager.go. We can also move to a common library

internal/sign/job/signer.go
yevgeny-shnaidman: I suggest we move all the makeVolume functions to some common package, since they can be used as is ( with changes in parameters) both in job makers for build and sign

internal/sign/manager.go
yevgeny-shnaidman: i think it should be renamed to "imageToSign"

internal/sign/job/manager.go
chr15p: changed to "Signing in cluster"

ybettan:
How does this PR relate to https://github.com/kubernetes-sigs/kernel-module-management/pull/101/files?

chr15p:
The basic architecture here is that kmm runs a k8s job to create the image with the kmods in it (kaniko), then it will run a separate k8s job that takes that image, signs the kmods it contains, and outputs a new image. Then it will run this second, signed, image that runs modprobe.  So the relationships between the PRs are:

- [PR101](https://github.com/kubernetes-sigs/kernel-module-management/pull/101/files ) adds the code to sign kmods, and the dockerfile etc that create the image that gets run for the signing k8s job.

- This PR adds the functionality to start and control that signing job.  So this code will run the job from [PR101](https://github.com/kubernetes-sigs/kernel-module-management/pull/101/files ) .

- There will then be a third PR that adds the hooks into the reconcile loop that call this code.

- Also PRs 88, 91 and 97 which have already been merged and added the Sign section to the CRD and made some other changes needed in other modules.

Its broken up this way because if I do it all as one PR that's a huge block of code that is hard to code review properly, hard to fix issues with, and even harder to keep rebased onto HEAD while its code reviewed (the merge conflicts were horrible!). Hopefully when these two PRs are merged the reconcile PR will be small and much easier to rebase as needed.

Hopefully that all makes sense to other people, not just in my head!!

internal/sign/job/manager.go
chr15p: I've created a new library internal/jobmanager/jobmanager.go  and moved most of the supporting functions used by sync over to it, let me know what you think

chr15p: internal/jobmanager/jobmanager.go now includes the functions like makeImageSigningSecretVolume() that are used to create the job.
I'm not going to update the build job in this PR, it would make the scope of this change much wider (requiring changes not jsut to build but also to main.go and the controller) but its all ready for a future change.

PR #123, opened 2022-10-23
Change the condition for node label deletion when ModuleLoader is del…

yevgeny-shnaidman:
…eted

When DaemonSet of ModulLoader is deleted, its pods start the process of pod lifecycle termination. The pre-stop hook is run immediatly and unloads the kernel module. The pods DeletionTimestamp is set, but the state stays Ready, due to a probable signal blocking by bash/sleep. This commit changes the condition for label deletion to look at DeletionTimestamp also, in order to remove the label in parallel to kernel module removal

In addition  this commit also changes CRD generated yaml file with sign section,
which were forgotten during commit that introduce sign section into CRD object

### DIFF
--- a/config/crd/bases/kmm.sigs.k8s.io_modules.yaml
+++ b/config/crd/bases/kmm.sigs.k8s.io_modules.yaml
@@ -2077,6 +2077,48 @@ spec:
                               description: Regexp is a regular expression to be match
                                 against node kernels.
                               type: string
+                            sign:
+                              description: Sign enables in-cluster signing for this
+                                mapping
+                              properties:
+                                certSecret:
+                                  description: a secret containing the public key
+                                    used to sign kernel modules for secureboot
+                                  properties:
+                                    name:
+                                      description: 'Name of the referent. More info:
+                                        https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                        TODO: Add other useful fields. apiVersion,
+                                        kind, uid?'
+                                      type: string
+                                  type: object
+                                  x-kubernetes-map-type: atomic
+                                filesToSign:
+                                  description: paths inside the image for the kernel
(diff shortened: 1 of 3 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
```
ybettan, line 15: Forgotten.

config/crd/bases/kmm.sigs.k8s.io_modules.yaml
```diff
@@ -2077,6 +2077,48 @@ spec:
                               description: Regexp is a regular expression to be match
                                 against node kernels.
                               type: string
+                            sign:
```
ybettan, line 2080: I guess all of this is related to the fact you are still using your personal image in the PR?!

config/manager/kustomization.yaml
yevgeny-shnaidman: fixed

config/crd/bases/kmm.sigs.k8s.io_modules.yaml
yevgeny-shnaidman: no, i think it is due to the fact that Chris probably did not push the generated files, when he pushed the sign change into CRD

ybettan:
Here is a [task](https://issues.redhat.com/browse/MGMT-11470) to make sure we don't forget to merge the generated code together with the PR.

ybettan:
/lgtm
/hold
Unhold once CI is green (tide isn't enforcing GH actions correctly).

PR #127, opened 2022-10-26
Add EXTRA_MODULE variable to kmm-kmod Makefile

enriquebelarte:
Add EXTRA_MODULE variable to Makefile so extra module name

can be set in args.

### DIFF
--- a/ci/kmm-kmod/Makefile
+++ b/ci/kmm-kmod/Makefile
@@ -2,10 +2,13 @@ ifndef KERNEL_SRC_DIR
 KERNEL_SRC_DIR =/lib/modules/$(shell uname -r)/build
 endif
 
+ifndef MY_MODULE
+EXTRA_MODULE=kmm_ci_b.o
+endif
 
 ccflags-y += -I$(KERNEL_SRC_DIR)/include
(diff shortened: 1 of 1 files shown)


qbarrand:
Could this be something more explicit, such as `EXTRA_MODULE`?

ybettan:
Can you please elaborate on the use-case for that change?

enriquebelarte:
> Can you please elaborate on the use-case for that change? Why do we need it? An example on the previous usage VS the new usage can help as well.

It is something that @cvultur thought as something that would help on separating test-cases so name of the "EXTRA" module would not be always a fixed name and could be set as an extra argument during build step, ie:

...  
 `build:`  
 `buildArgs:`  
 `- name: EXTRA_MODULE`  
 `value: multi-stage.o`  
 `dockerfile: |`  
 `FROM gcr.io/myrepo/base:0.1 as builder`  
 `ARG KERNEL_VERSION`  
 `ARG EXTRA_MODULE`

...

If no EXTRA\_MODULE argument is set then behavior is same as before.

cdvultur:
Indeed @enriquebelarte, that is the purpose for this change

/lgtm

Issue #109, opened 2022-10-13
FirmwarePath extra directory level causing FW initialization issue.


hershpa:
**Issue Summary:**
FirmwarePath extra directory level is causing FW initialization to fail.

**Suggested Priority(P1-P3) & Urgent(Urgent, medium, Low):**
P1 and Urgent

**Issue Detail:**
We have added `firmware_class.path=/var/lib/firmware` as a kernel parameter using MCO to tell the kernel to look for firmware in /var/lib/firmware since the /lib/firmware is read only on RHCOS. 

KMMO copies the firmware to `/var/lib/firmware/Module.Name/FirmwarePath/`. Due to this extra directory level, FirmwarePath, the FW initialization fails as it cannot find the firmware binaries. When I manually copy the firmware to `/var/lib/firmware/Module.Name/`, then the FW initialization works.

**Solution Proposal:**
Is it possible for KMMO to copy firmware to  `/var/lib/firmware/FirmwarePath/` instead of `/var/lib/firmware/Module.Name/FirmwarePath/`. This removes Module.Name from the path and thus can be customized based on what value the user sets for FirmwarePath. 

One other related question, can the FirmwarePath be referenced as a variable in the dockerfile?

The use case would be user can simply reference FirmwarePath as a variable when they are copying the firmware from builder image. It would be a improvement from a user point of view so that firmware is always in the right place and user does not have to explicitly list path again in dockerfile.

hershpa:
Another question related to this issue, when the path is created for firmware by KMMO (/var/lib/firmware/<Module.Name>), it appends the name of the module (metadata.name). Wouldn't it make more sense for KMMO to use spec.moduleLoader.container.modprobe.moduleName when it creates the firmware path?

uMartinXu:
In the GA release installed from OCP-4.12 operator Hub. This issue is still there.

qbarrand:
Verified today with @hershpa that it is working as expected.

PR #133, opened 2022-11-03
Fix failing firmware requests

mresvanis:
This PR fixes failing firmware requests, which are caused by the additional firmware path directory, named after the Module CR, which is created by KMM when copying the firmware files from the module-loader container to the host directory (and default firmware search path) `/var/lib/firmware`.

Modules request firmware files using a relative path, e.g. `vendor/device/firmware.bin`, which is appended to the default firmware search path. The additional firmware path directory makes the firmware requests unresolvable.

Specifically, this change:

- removes the additional firmware path directory, named after the Module CR
- ensures that only the copied firmware files are removed after the module is unloaded and non-empty firmware directories are skipped (this was not needed before, because each Module CR had its own firmware subpath under `/var/lib/firmware` and the whole subpath could be removed)

> It should be noted that this change assumes that `find` and `xargs` are available in the module-loader container. The `ubi-minimal` base image does not included them by default and the `findutils` package needs to be installed. 

Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/109

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -170,13 +170,11 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
 	}
 
 	if fw := mod.Spec.ModuleLoader.Container.Modprobe.FirmwarePath; fw != "" {
-		moduleFirmwarePath := fmt.Sprintf("%s/%s", nodeVarLibFirmwarePath, mod.Name)
-
 		firmwareVolume := v1.Volume{
 			Name: nodeVarLibFirmwareVolumeName,
 			VolumeSource: v1.VolumeSource{
 				HostPath: &v1.HostPathVolumeSource{
-					Path: moduleFirmwarePath,
+					Path: nodeVarLibFirmwarePath,
 					Type: &hostPathDirectoryOrCreate,
 				},
 			},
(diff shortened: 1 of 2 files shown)


mresvanis:
@fabiendupont

internal/daemonset/daemonset.go
```diff
@@ -421,7 +419,7 @@ func MakeUnloadCommand(spec kmmv1beta1.ModprobeSpec, modName string) []string {
 
 	fwUnloadCommand := ""
 	if fw := spec.FirmwarePath; fw != "" {
-		fwUnloadCommand = fmt.Sprintf(" && rm -rf %s/%s", nodeVarLibFirmwarePath, modName)
+		fwUnloadCommand = fmt.Sprintf(" && cd %s && find |sort -r |xargs -I{} rm -d %s/$(basename %s)/{}", fw, nodeVarLibFirmwarePath, fw)
```
yevgeny-shnaidman, line 422: why cannot we just remove the file with rm -f <fileName> command? We have the directory (nodeVarLibFirmwarePath) and the file name/subpath (spec.FirmwarePath)...

mresvanis: If we try to remove the whole directory directly, it could be that we remove files that were not copied for the Module being unloaded. The latter would lead to another Module being broken. For example:

- `Module A` includes in its module-loader image this `/opt/lib/firmware/vendor/firmware_a.bin` fw and the Module's `FirmwarePath: /opt/lib/firmware/vendor` 
- `Module B`  includes in its module-loader image this `/opt/lib/firmware/vendor/firmware_b.bin` fw and the Module's `FirmwarePath: /opt/lib/firmware/vendor`
- After loading both Modules, the host firmware directory would be `/var/lib/firmware/vendor/{firmware_a.bin,firmware_b.bin}`
- By unloading `Module A` and `rm`ing the `/var/lib/firmware/vendor/` dir, we would also remove Module B's firmware

What do you think? Does this make sense?

PR #134, opened 2022-11-03
Fix OLM bundle failing because of incorrect CSV filename

mresvanis:


ybettan:
Thanks for that PR @mresvanis 

/ok-to-test
/lgtm
/approve

PR #139, opened 2022-11-08
Add check-api-changes script to CI

enriquebelarte:
When API code changes, we need to regenerate CRD configs using make generate manifests.
This script checks that regenerated files are pushed in the same PR which changes the API
 code so CRD configs are not merged in next unrelated PRs.

### DIFF
--- a/ci/prow/check-api-changes
+++ b/ci/prow/check-api-changes
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+make generate manifests
+
+
+if [[ $(git diff) ]]; then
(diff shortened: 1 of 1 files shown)


ci/prow/check-api-changes
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+make generate manifests
+
+
+if [[ $(git diff) ]]; then
+   echo 'Please run `make generate manifests` and repush'
```
qbarrand, line 9: Could we print the diff here?

enriquebelarte: Done

PR #149, opened 2022-11-11
Garbage collection for successfull build jobs

yevgeny-shnaidman:
This PR contains the following changes:
1) impement garbage collection for successfull build jobs and pods 2) refactoring jobhelper package and moving build code to use it 3) refactoring/adding unit-test

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -163,14 +163,10 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return res, fmt.Errorf("could handle device plugin: %w", err)
 	}
 
-	logger.Info("Garbage-collecting DaemonSets")
-
-	// Garbage collect old DaemonSets for which there are no nodes.
-	validKernels := sets.StringKeySet(mappings)
-
-	deleted, err := r.daemonAPI.GarbageCollect(ctx, dsByKernelVersion, validKernels)
+	logger.Info("Run garbage collection")
+	err = r.garbageCollect(ctx, mod, mappings, dsByKernelVersion)
 	if err != nil {
-		return res, fmt.Errorf("could not garbage collect DaemonSets: %v", err)
+		return res, fmt.Errorf("failed to run garbage collection: %v", err)
 	}
 
 	err = r.statusUpdaterAPI.ModuleUpdateStatus(ctx, mod, nodesWithMapping, targetedNodes, dsByKernelVersion)
@@ -178,7 +174,7 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return res, fmt.Errorf("failed to update status of the module: %w", err)
 	}
 
-	logger.Info("Garbage-collected DaemonSets", "names", deleted)
+	logger.Info("Reconcile loop finished successfully")
 
 	return res, nil
 }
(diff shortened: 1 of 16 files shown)


controllers/module_reconciler.go
```diff
@@ -357,6 +353,32 @@ func (r *ModuleReconciler) handleDevicePlugin(ctx context.Context, mod *kmmv1bet
 	return err
 }
 
+func (r *ModuleReconciler) garbageCollect(ctx context.Context,
```
qbarrand, line 356: Do we want to collect signing jobs as well?

internal/utils/jobhelper.go
```diff
@@ -16,14 +17,17 @@ import (
 type Status string
 
 const (
-	jobHashAnnotation = "kmm.node.kubernetes.io/last-hash"
+	BuildType = "build"
+	SignType  = "sign"
```
qbarrand, line 21: `utils.BuildType` is not very explicit IMHO.
```suggestion
	JobTypeBuild = "build"
	JobTypeSign  = "sign"
```
Also wondering if the Job helper should have its own package.

internal/utils/jobhelper.go
```diff
@@ -127,3 +129,31 @@ func (jh *jobHelper) GetJobStatus(job *batchv1.Job) (Status, bool, error) {
...
+
+	return jobList.Items, nil
+}
+
+func moduleKernelLabels(moduleName, targetKernel, jobType string) map[string]string {
+	return map[string]string{
+		constants.ModuleNameLabel:    moduleName,
+		constants.TargetKernelTarget: targetKernel,
+		constants.JobType:            jobType,
+	}
```
qbarrand, line 151:
```suggestion
	m := moduleLabels(moduleName, jobType)
	m[constants.TargetKernelTarget] = targetKernel
	
	return m
```

internal/build/job/manager.go
```diff
@@ -7,59 +7,48 @@ import (
...
-		client.MatchingLabels(labels(mod, targetKernel)),
-		client.InNamespace(mod.Namespace),
-	}
-
-	if err := jbm.client.List(ctx, &jobList, opts...); err != nil {
-		return nil, fmt.Errorf("could not list jobs: %v", err)
+func (jbm *jobManager) GarbageCollect(ctx context.Context, mod kmmv1beta1.Module) ([]string, error) {
+	jobs, err := jbm.jobHelper.GetModuleJobs(ctx, mod, utils.BuildType)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get build jobs for module %s", mod.Name)
```
qbarrand, line 35:
```suggestion
		return nil, fmt.Errorf("failed to get build jobs for module %s: %v", mod.Name, err)
```

internal/build/job/manager.go
```diff
@@ -74,32 +63,29 @@ func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1b
...
 		return build.Result{}, fmt.Errorf("could not determine if job has changed: %v", err)
 	}
 
 	if changed {
 		logger.Info("The module's build spec has been changed, deleting the current job so a new one can be created", "name", job.Name)
-		opts := []client.DeleteOption{
-			client.PropagationPolicy(metav1.DeletePropagationBackground),
-		}
-		err = jbm.client.Delete(ctx, job, opts...)
+		err = jbm.jobHelper.DeleteJob(ctx, job)
```
qbarrand, line 88: Why remote the propagation option here?

controllers/module_reconciler.go
yevgeny-shnaidman: yes, but it will be implemented after all Chris' PRs are completed

internal/utils/jobhelper.go
yevgeny-shnaidman: yes, it can go both ways, but eventually i thought that its main purpose is to assist other package with limited functionalities, not as a standalone package, that's why i asked Chris to put it in utils. Both jobhelper and volume helper contain command functions for build and sign packages, that is additional reason why  it seems to me more appropriate to put them in utils

yevgeny-shnaidman: fixed the rename of types

internal/utils/jobhelper.go
yevgeny-shnaidman: unless you insist, i prefer to keep it as is: i think it is slightly more readable, since you see immediately all the labels, without need to check what moduleLabels function does

internal/build/job/manager.go
yevgeny-shnaidman: fixed

internal/build/job/manager.go
yevgeny-shnaidman: i am guessing you meant "remove" :). propagation policy is now part of DeleteJob function

internal/utils/jobhelper.go
qbarrand: Aren't you afraid that at some point we update one but forget to update the other?

internal/build/job/manager.go
```diff
@@ -7,59 +7,48 @@ import (
...
-		return nil, errNoMatchingBuild
-	} else if n > 1 {
-		return nil, fmt.Errorf("expected 0 or 1 job, got %d", n)
+	deleteNames := make([]string, 0, len(jobs))
+	for _, job := range jobs {
+		if job.Status.Succeeded == 1 {
+			opts := []client.DeleteOption{
+				client.PropagationPolicy(metav1.DeletePropagationBackground),
+			}
+			err = jbm.client.Delete(ctx, &job, opts...)
```
qbarrand, line 44: Why isn't this using `DeleteJob` from the helper?

yevgeny-shnaidman: fixed

qbarrand (approved):

PR #151, opened 2022-11-14
Change back nodeLibModulesPath to /lib/modules

enriquebelarte:
Mounting nodeLibModulesPath as '/lib/modules/'+ kernelVersion
makes GKE fail when Module loader is run, as it tries to mount 
host '/lib/modules' at '/opt/lib/modules/<kernelVersion>+'
which results in depmod error not finding directory.
```
INFO[1403] Running: [/bin/sh -c depmod -b /opt]         
depmod: ERROR: could not open directory /opt/lib/modules/5.10.133+: No such file or directory
depmod: FATAL: could not search modules: No such file or directory
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 1
```


enriquebelarte:
I'm closing this PR as it is something specifically set for COS running nodes at GKE in Google Cloud and still need to investigate other related issues and probably find a better solution to make kmm work there.

PR #152, opened 2022-11-14
Add the kernel - ClusterClaim reconciler

qbarrand:
This reconciler collects all kernel versions running in the cluster and populates a ClusterClaim with them.
To minimize updates to the ClusterClaim (which triggers a sync back to the ManagedCluster), kernel versions are sorted alphabetically and separated by a new line.
This controller is not started by any manager yet because we still have to determine if we add it to the mainline KMM or if we create a dedicated image for Spoke clusters.

/cc @mresvanis @yevgeny-shnaidman @ybettan

### DIFF
--- a/controllers/node_kernel_clusterclaim.go
+++ b/controllers/node_kernel_clusterclaim.go
@@ -1,0 +1,112 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+
+	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"open-cluster-management.io/api/cluster/v1alpha1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;patch;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=create;delete;get;list;patch;update;watch
+
+const clusterClaimName = "kernel-versions.kmm.node.kubernetes.io"
+
+type NodeKernelClusterClaimReconciler struct {
+	client client.Client
+}
+
+func NewNodeKernelClusterClaimReconciler(client client.Client) *NodeKernelClusterClaimReconciler {
+	return &NodeKernelClusterClaimReconciler{client: client}
+}
+
+func (r *NodeKernelClusterClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	nodeList := v1.NodeList{}
+
+	logger.Info("Listing all nodes")
+
+	if err := r.client.List(ctx, &nodeList); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not list nodes: %v", err)
+	}
+
+	kernelsMap := make(map[string]struct{})
+
+	for _, n := range nodeList.Items {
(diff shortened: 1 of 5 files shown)


controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (r *NodeKernelClusterClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	nodeList := v1.NodeList{}
+
+	logger.Info("Listing all nodes")
+
+	if err := r.client.List(ctx, &nodeList); err != nil {
```
yevgeny-shnaidman, line 43: do we want to add a check for Node status? working only on nodes with Ready status?

controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,112 @@
...
+	for _, n := range nodeList.Items {
+		kernelsMap[n.Status.NodeInfo.KernelVersion] = struct{}{}
+	}
+
+	// The map's order is unpredictable; we do not want to update the ClusterClaim if the value only
+	// changed because of a different order in the kernel versions.
+	// Copy the kernels into a slice and sort it alphabetically.
+	kernelsSlice := make([]string, 0, len(kernelsMap))
+
+	for k := range kernelsMap {
```
yevgeny-shnaidman, line 58:
```suggestion
	for k := range kernelsMap {
	   kernelsSlice = append(kernelsSlice, k)
	}
```

controllers/node_kernel_clusterclaim.go
qbarrand: If anything, I would filter out the unschedulable (tainted) nodes like you did in KMM. What are your thoughts on this? Its is reasonable to assume that those nodes will never need an OOT module?

controllers/node_kernel_clusterclaim.go
qbarrand: Surprisingly, that works. Fixed.

controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,111 @@
...
+	logger.Info("Reconciled ClusterClaim", "res", opRes)
+
+	return ctrl.Result{}, err
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *NodeKernelClusterClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.
+		NewControllerManagedBy(mgr).
+		Named("node-clusterclaim").
```
ybettan, line 87:
```suggestion
		Named("node-kernel-clusterclaim").
```
To be consistent?

controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,111 @@
...
+		// Each time a Node is updated, enqueue a reconciliation request
```
ybettan, line 88: I would remove this comment or update it (only when the kernelversion in the NodeInfo is updated).

controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (r *NodeKernelClusterClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+
+	nodeList := v1.NodeList{}
+
+	logger.Info("Listing all nodes")
+
+	if err := r.client.List(ctx, &nodeList); err != nil {
+		return ctrl.Result{}, err
```
ybettan, line 44:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to list nodes: %v", err)
```

controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,111 @@
...
+
+	logger.Info("Creating or patching ClusterClaim")
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, r.client, &cc, func() error {
+		cc.Spec = v1alpha1.ClusterClaimSpec{
+			Value: strings.Join(kernelsSlice, "\n"),
+		}
+
+		return nil
+	})
```
ybettan, line 76: We also need to report the DTK images.
I am wondering if we can leverage some code from https://github.com/rh-ecosystem-edge/kernel-module-management/pull/96

qbarrand: This PR is upstream - let's revisit that during the backport.

controllers/node_kernel_clusterclaim.go
```diff
@@ -0,0 +1,112 @@
...
+		NewControllerManagedBy(mgr).
+		Named("node-kernel-clusterclaim").
+		// Each time a Node's .status.nodeInfo.kernelVersion is updated, enqueue a reconciliation request
+		For(
+			&v1.Node{},
+			builder.WithPredicates(
+				filter.NodeUpdateKernelChangedPredicate(),
+			),
+		).
+		// Each time our ClusterClaim, enqueue an empty reconciliation request.
```
mresvanis, line 96:
```suggestion
		// Each time our ClusterClaim is updated, enqueue an empty reconciliation request.
```

PR #153, opened 2022-11-14
Add the ManagedClusterModule CRD and reconciler

mresvanis:
The `ManagedClusterModule` CRD is a new Module wrapper that features a selector for ManagedClusters to target specific Spoke clusters.

The respective reconciler creates a `ManifestWork` in the corresponding namespace to deploy a Module in the Spoke cluster. When crafting the `ManifestWork`, the reconciler removes any build or sign section.

`kmm` already supports building and signing kernel modules via the `Module` CRD. This reconciler uses the same features, in order to build and sign the kernel modules only on the Hub.


controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+	logger := log.FromContext(ctx)
+
+	mcm, err := r.getRequestedManagedClusterModule(ctx, req.NamespacedName)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("ManagedClusterModule deleted")
+			return res, nil
+		}
+
+		return res, fmt.Errorf("failed to get the requested %s KMMO CR: %w", req.NamespacedName, err)
```
qbarrand, line 63: The logger should already contain the resource name.
```suggestion
		return res, fmt.Errorf("failed to get the requested CR: %v", err)
```

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	logger.Info("Requested KMMO ManagedClusterModule", "namespace", mcm.Name, "name", mcm.Namespace)
```
qbarrand, line 66: The logger from the context has those fields already.
```suggestion
	logger.Info("Requested KMMO ManagedClusterModule")
```

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+func (r *ManagedClusterModuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&kmmv1beta1.ManagedClusterModule{}).
+		Complete(r)
+}
+
+func (r *ManagedClusterModuleReconciler) getRequestedManagedClusterModule(ctx context.Context, namespacedName types.NamespacedName) (*kmmv1beta1.ManagedClusterModule, error) {
+	mcm := kmmv1beta1.ManagedClusterModule{}
+	if err := r.Client.Get(ctx, namespacedName, &mcm); err != nil {
+		return nil, fmt.Errorf("failed to get the kmmo ManagedClusterModule %s: %w", namespacedName, err)
```
qbarrand, line 81:
```suggestion
		return nil, fmt.Errorf("failed to get the ManagedClusterModule %s: %v", namespacedName, err)
```

api/v1beta1/managedclustermodule_types.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// ManagedClusterModuleStatus defines the observed state of ManagedClusterModule
+type ManagedClusterModuleStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
```
ybettan, line 39: What does those 2 means?

config/samples/kmm_v1beta1_managedclustermodule.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: kmm.sigs.k8s.io/v1beta1
```
ybettan, line 1: What is that file for?

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// ManagedClusterModuleReconciler reconciles a ManagedClusterModule object
+type ManagedClusterModuleReconciler struct {
+	client.Client
+
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=managedclustermodules,verbs=get;list;watch;create;update;patch;delete
```
ybettan, line 40: Why does the controller should be able to create `ManagedClusterModules`?

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=managedclustermodules/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=managedclustermodules/finalizers,verbs=update
+
+func NewManagedClusterModuleReconciler(
```
ybettan, line 44: Don't we need to initialize `Scheme` as well?

controllers/managedclustermodule_reconciler.go
ybettan: Can we add the resource name to the error message? So we can differentiate between different error in different modules?

controllers/managedclustermodule_reconciler.go
ybettan: Is the resource name part of the logger?

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+		return res, fmt.Errorf("failed to get the requested %s KMMO CR: %w", req.NamespacedName, err)
+	}
+
+	logger.Info("Requested KMMO ManagedClusterModule", "namespace", mcm.Name, "name", mcm.Namespace)
+
+	return res, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ManagedClusterModuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
ybettan, line 72: I don't see any caller in `main.go`. I guess it is not implemented yet?!

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&kmmv1beta1.ManagedClusterModule{}).
+		Complete(r)
```
ybettan, line 75: Don't we want to add a `Name` and `Filter` to the reconciler?

yevgeny-shnaidman:
@mresvanis this PR's reconciler is just a skeleton, right? Because i doesn't do much right now

controllers/managedclustermodule_reconciler.go
mresvanis: > @mresvanis this PR's reconciler is just a skeleton, right? Because i doesn't do much right now

That's correct, it's just a first skeleton (mostly what is scaffolded by `kubebuilder` and just a few minor details. I opened the PR (Draft) early, because it's gonna be too big (I believe) for only a final review.

mresvanis:
> @mresvanis this PR's reconciler is just a skeleton, right? Because i doesn't do much right now

@yevgeny-shnaidman that's correct, it's just the skeleton now (i.e. mostly what is scaffolded by `kubebuilder` and just a few minor details). I opened the PR (as a draft) early, because it's gonna be too big (I believe) for only a final review.

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,84 @@
...
+// ManagedClusterModuleReconciler reconciles a ManagedClusterModule object
+type ManagedClusterModuleReconciler struct {
+	client.Client
+
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=managedclustermodules,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=managedclustermodules/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=kmm.sigs.k8s.io,resources=managedclustermodules/finalizers,verbs=update
```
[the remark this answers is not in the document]
mresvanis: This was scaffolded by `kubebuilder` when I created the CRD and the reconciler. But it's an interesting question and I think the reason is the [OwnerReferencesPermissions admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement), which when enabled in a cluster does not allow the update of the `metadata.finalizers` field of a Custom Resource.

api/v1beta1/managedclustermodule_types.go
mresvanis: These 2 `kubebuilder` [markers](https://book.kubebuilder.io/reference/markers.html) were scaffolded by `kubebuilder` when I created the CRD and the reconciler.

The [first](https://book.kubebuilder.io/reference/markers/object.html) one enables interface implementation generation for this type. The [second](https://book.kubebuilder.io/reference/markers/crd.html) enables the `/status` subresource for this type.

Does this make sense?

config/samples/kmm_v1beta1_managedclustermodule.yaml
mresvanis: This was scaffolded by `kubebuilder` when I created the CRD. It's created [in order to be used by developers](https://book.kubebuilder.io/cronjob-tutorial/running.html?highlight=samples#optional), when trying to test the operator.   

Is this helpful?

controllers/managedclustermodule_reconciler.go
qbarrand: Yes, resource name and namespace.

controllers/managedclustermodule_reconciler.go
mresvanis: I'll revert to using `%w` for the error, in order to be able to check for a `NotFound` API error after calling this method.

```go
return nil, fmt.Errorf("failed to get the ManagedClusterModule %s: %w", namespacedName, err)
```

Does this make sense?

ybettan: Makes sense to me.

api/v1beta1/managedclustermodule_types.go
ybettan: Yup. Thank you for the explanation.

config/samples/kmm_v1beta1_managedclustermodule.yaml
ybettan: Yup. Thanks again.

controllers/managedclustermodule_reconciler.go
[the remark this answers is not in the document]
mresvanis: But I would guess that we won't need any `finalizer` for this CRD and reconciler, so I'll remove this permission.

controllers/managedclustermodule_reconciler.go
mresvanis: This is a very good question. This marker was created by `kubebuilder` when I created the CRD and the reconciler, but I can't find any reason for the reconciler's `ServiceAccount` to have `create` and `delete` permissions over `ManagedClusterModule` CRs. I will remove them.

controllers/managedclustermodule_reconciler.go
mresvanis: I added a name, but I don't think we need a `Filter` at this point. I will add any `Owns` statements as we move along with the rest of the implementation though. What do you think?

controllers/managedclustermodule_reconciler.go
mresvanis: @ybettan PTAL, I pushed additional business logic, which I think answers this question. The `Scheme` can be defined when initializing the `manifestAPI` struct.

controllers/managedclustermodule_reconciler.go
qbarrand: It does 👍🏻

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	logger.Info("Requested KMMO ManagedClusterModule")
+
+	clusters, err := r.getSelectedManagedClusters(ctx, mcm)
+	if err != nil {
+		return res, fmt.Errorf("failed to get selected clusters: %v", err)
+	}
+
+	if err := r.manifestAPI.GarbageCollect(ctx, *clusters, *mcm); err != nil {
```
yevgeny-shnaidman, line 81: I think it is better to issue garbage collection for manifests at the end of the reconcile loop.

mresvanis: @yevgeny-shnaidman thank you for the suggestion, I changed the code accordingly :+1:

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,156 @@
...
+		client.MatchingLabelsSelector{
+			Selector: labels.Set(mcm.Spec.Selector).AsSelector(),
+		},
+	}
+	err := r.Client.List(ctx, clusterList, opts...)
+
+	return clusterList, err
+}
+
+func (r *ManagedClusterModuleReconciler) getSkipUpdatesPredicate() predicate.Predicate {
```
yevgeny-shnaidman, line 152: we have a filter package that contains all the filters for our reconcilers, i think it is better to move this function there

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,156 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ManagedClusterModuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&kmmv1beta1.ManagedClusterModule{}).
+		Owns(&workv1.ManifestWork{}).
+		Watches(
+			&source.Kind{Type: &clusterv1.ManagedCluster{}},
+			&handler.EnqueueRequestForObject{},
+			builder.WithPredicates(r.getSkipUpdatesPredicate())).
```
yevgeny-shnaidman, line 120: i am probably missing something from the flow: why don't we want to watch for update events on the ManagedClusters?

mresvanis: Actually, since we have a cluster labels selector on the `ManagedClusterModule` CR, those labels could change and this would affect the cluster selector. So we should include `ManagedCluster` update events. 

Thanks for the suggestion, I removed this predicate altogether :+1:

controllers/managedclustermodule_reconciler.go
mresvanis: Thanks for pointing this out. Given [this](https://github.com/kubernetes-sigs/kernel-module-management/pull/153#discussion_r1023775040) resolution, I have removed this method.

controllers/managedclustermodule_reconciler.go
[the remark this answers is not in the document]
mresvanis: I added this permission back, as in order to have the [BlockOwnerDeletion enabled](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L79) the former is necessary.

controllers/managedclustermodule_reconciler.go
mresvanis: The resource name is already included when logging the error, as @qbarrand suggested [above](https://github.com/kubernetes-sigs/kernel-module-management/pull/153#discussion_r1021895089).

api/v1beta1/managedclustermodule_types.go
```diff
@@ -0,0 +1,61 @@
...
+type ManagedClusterModuleSpec struct {
+	// ModuleSpec describes how the KMM operator should deploy a Module on those nodes that need it.
+	ModuleSpec ModuleSpec `json:"moduleSpec,omitempty"`
+
+	// Selector describes on which managed clusters the ModuleSpec should be applied.
+	Selector map[string]string `json:"selector"`
+}
+
+// ManagedClusterModuleStatus defines the observed state of ManagedClusterModule
+type ManagedClusterModuleStatus struct {
```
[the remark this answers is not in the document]
mresvanis: We are indeed going to monitor the status of the Spoke `Module` CRs via the `ManifestWork` CRs' `Status`. But this functionality will be implemented in a subsequent PR. So I have removed the `ManagedClusterModuleStatus` from this PR, as suggested :+1:

controllers/managedclustermodule_reconciler.go
mresvanis: @ybettan this reconciler will be only included in the KMM-Hub version of KMM, i.e. either by adding a configuration flag or by building a dedicated KMM-Hub container image (this is a design discussion and it's still ongoing). As a result, we won't be enabling this reconciler in `main.go` in this PR. What do you think?

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,151 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/manifestwork"
+)
+
+// ManagedClusterModuleReconciler reconciles a ManagedClusterModule object
+type ManagedClusterModuleReconciler struct {
+	client.Client
```
qbarrand, line 43: Could we make client a private variable? I don't understand why kubebuilder scaffolds things that way.

internal/filter/filter.go
```diff
@@ -107,6 +107,59 @@ func (f *Filter) FindModulesForNode(node client.Object) []reconcile.Request {
...
+
+			sel = sel.Add(*requirement)
+		}
+
+		if !sel.Matches(clusterLabelsSet) {
+			logger.V(1).Info("Cluster labels do not match the ManagedClusterModule's selector; skipping")
+			continue
+		}
+
+		nsn := types.NamespacedName{Name: mod.Name, Namespace: mod.Namespace}
```
qbarrand, line 152:
```suggestion
		nsn := types.NamespacedName{Name: mod.Name}
```

internal/manifestwork/manifestwork.go
```diff
@@ -0,0 +1,120 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
+)
+
+//go:generate mockgen -source=manifestwork.go -package=manifestwork -destination=mock_manifestwork.go
+
+type ManifestWorkCreator interface {
```
qbarrand, line 23: Is `Creator` the right name, considering that it also runs garbage collection?

internal/manifestwork/manifestwork.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+}
+
+func (mwg *manifestWorkGenerator) GarbageCollect(ctx context.Context, clusters clusterv1.ManagedClusterList, mcm kmmv1beta1.ManagedClusterModule) error {
+	manifestWorks, err := mwg.getOwnedManifestWorks(ctx, mcm)
+	if err != nil {
+		return fmt.Errorf("failed to get owned ManifestWorks: %w", err)
+	}
+
+	clusterNames := map[string]string{}
```
qbarrand, line 46: I think using the [`sets.String`](https://pkg.go.dev/k8s.io/apimachinery@v0.25.4/pkg/util/sets#String) type from apimachinery would be more explicit here.

internal/manifestwork/manifestwork.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (mwg *manifestWorkGenerator) SetManifestWorkAsDesired(ctx context.Context, mw *workv1.ManifestWork, mcm kmmv1beta1.ManagedClusterModule) error {
+	if mw == nil {
+		return errors.New("mw cannot be nil")
+	}
+
+	// Ensure no Build and Sign specs reach the Spoke cluster
+	mcm.Spec.ModuleSpec.ModuleLoader.Container.Build = nil
+	mcm.Spec.ModuleSpec.ModuleLoader.Container.Sign = nil
```
qbarrand, line 71: We should also go through all kernel mappings to strip those sections as well.

controllers/managedclustermodule_reconciler.go
mresvanis: Absolutely, thanks for pointing this out, I should have corrected it as I'm familiar with this not optimal scaffolding here.

internal/filter/filter.go
mresvanis: Oh nice catch! I applied this change :+1:

internal/manifestwork/manifestwork.go
mresvanis: Awesome! Couldn't make more sense :+1:

internal/manifestwork/manifestwork.go
mresvanis: Great catch, I completely missed this. I adjusted the code and also added the respective unit test case.

internal/manifestwork/manifestwork.go
mresvanis: That is a valid point and we should definitely revisit the naming of this API, but because of the lack of a better name, let's keep the convention already found in the codebase (like with the [DaemonSetCreator](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/internal/daemonset/daemonset.go#L33)) and come back to this at a later point.

controllers/managedclustermodule_reconciler.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ManagedClusterModuleReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&kmmv1beta1.ManagedClusterModule{}).
+		Owns(&workv1.ManifestWork{}).
+		Watches(
+			&source.Kind{Type: &clusterv1.ManagedCluster{}},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindManagedClusterModulesForCluster)).
```
qbarrand, line 121: How about adding a predicate here that filters out `ManagedCluster` update events but with unchanged labels?

controllers/managedclustermodule_reconciler.go
ybettan: I am wondering how it will look like. Correct me if I am wrong but the way I see it we have 2 options:
* runtime option: in that case we should have an `if` statement in `main` to start the controller in case some runtime argument were/weren't passed.
* buildtime option: in that case I wouldn't expect to compile the reconciler code in that image.

Am I missing something here?

controllers/managedclustermodule_reconciler.go
ybettan: Ack.

controllers/managedclustermodule_reconciler.go
mresvanis: The runtime option is clear, you're not missing anything there :+1: 

Regarding the different builds option, currently all controllers belong to the same `controllers` package. Which means that since this package needs to be `import`ed in a `main.go` for a user to run one or more of those controllers, then all controllers will be compiled and get into the binary (as you already said). 

One option we have been thinking to alleviate this issue is to move each of the optional controllers to their own Golang package and import them explicitly in the respective `main`s (we would anyway have as many `main`s as binaries). That way all binaries would only contain what's absolutely needed.

But until we reach a consensus about `runtime` vs `build` options, I thought it best to include this reconciler along with the rest of the reconcilers. If we go for the `build` option, then we would move it to its own package and also implement the rest of the supporting details around this, e.g. create the additional `cmd/kmm-hub/main.go`.

Does this make sense?

ybettan: Yup. It make a lot of sense.

Seems like the build option is way too complicated for what we really need and I tend to go with the runtime option but this is indeed a conversation which is not tightly coupled with this PR.

PR #156, opened 2022-11-18
Decouple build and sign from ModuleReconciler

mresvanis:
This PR decouples the `build` and `sign` internal APIs from the `ModuleReconciler`, in order to be able to use them in other reconcilers as well, i.e. the `ManagedClusterModuleReconciler`.

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -22,14 +22,12 @@ import (
 	"strings"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
-	"github.com/kubernetes-sigs/kernel-module-management/internal/auth"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/daemonset"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/metrics"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/rbac"
-	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/sign"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/statusupdater"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
@@ -60,7 +58,6 @@ type ModuleReconciler struct {
 	daemonAPI        daemonset.DaemonSetCreator
 	kernelAPI        module.KernelMapper
 	metricsAPI       metrics.Metrics
-	registry         registry.Registry
 	filter           *filter.Filter
 	statusUpdaterAPI statusupdater.ModuleStatusUpdater
 }
@@ -74,7 +71,6 @@ func NewModuleReconciler(
 	kernelAPI module.KernelMapper,
 	metricsAPI metrics.Metrics,
 	filter *filter.Filter,
-	registry registry.Registry,
 	statusUpdaterAPI statusupdater.ModuleStatusUpdater) *ModuleReconciler {
 	return &ModuleReconciler{
 		Client:           client,
@@ -85,7 +81,6 @@ func NewModuleReconciler(
 		kernelAPI:        kernelAPI,
 		metricsAPI:       metricsAPI,
 		filter:           filter,
-		registry:         registry,
 		statusUpdaterAPI: statusUpdaterAPI,
 	}
 }
@@ -146,9 +141,9 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	for kernelVersion, m := range mappings {
-		requeue, err := r.handleBuild(ctx, mod, m, kernelVersion, m.ContainerImage)
+		requeue, err := r.handleBuild(ctx, mod, m, kernelVersion)
 		if err != nil {
-			return res, fmt.Errorf("failed to handle build for kernel version %s: %w", kernelVersion, err)
+			return res, fmt.Errorf("failed to handle build for kernel version %s: %v", kernelVersion, err)
 		}
 		if requeue {
 			logger.Info("Build requires a requeue; skipping handling driver container for now", "kernelVersion", kernelVersion, "image", m)
@@ -156,7 +151,7 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			continue
 		}
 
-		signrequeue, err := r.handleSigning(ctx, mod, m, kernelVersion, m.ContainerImage)
+		signrequeue, err := r.handleSigning(ctx, mod, m, kernelVersion)
 		if err != nil {
 			return res, fmt.Errorf("failed to handle signing for kernel version %s: %v", kernelVersion, err)
 		}
@@ -265,33 +260,20 @@ func (r *ModuleReconciler) getNodesListBySelector(ctx context.Context, mod *kmmv
 func (r *ModuleReconciler) handleBuild(ctx context.Context,
 	mod *kmmv1beta1.Module,
 	km *kmmv1beta1.KernelMapping,
-	kernelVersion string,
-	containerImage string) (bool, error) {
+	kernelVersion string) (bool, error) {
 
-	//if theres no build specified skip this section
-	if !r.willBuild(mod, km) {
-		return false, nil
-	}
-	//if build is specified AND sign is specified then we want to build an intermediate image
-	// and let sign produce the one specified in containerImage
-	if r.willSign(mod, km) {
-		containerImage = r.appendToTag(containerImage, mod.Namespace+"_"+mod.Name+"_kmm_unsigned")
-	}
-
-	// build is specified and containerImage is either the final image, or the intermediate image tag depending on if sign is defined
-	// either way if containerImage exists we can skip building it
-	exists, err := r.checkImageExists(ctx, mod, km, containerImage)
+	shouldSync, err := r.buildAPI.ShouldSync(ctx, *mod, *km)
 	if err != nil {
-		return false, fmt.Errorf("failed to check existence of image %s for kernel %s: %w", containerImage, kernelVersion, err)
+		return false, fmt.Errorf("could not check if build synchronization is needed: %w", err)
 	}
-	if exists {
+	if !shouldSync {
 		return false, nil
 	}
 
-	logger := log.FromContext(ctx).WithValues("kernel version", kernelVersion, "image", containerImage)
+	logger := log.FromContext(ctx).WithValues("kernel version", kernelVersion, "image", km.ContainerImage)
 	buildCtx := log.IntoContext(ctx, logger)
 
-	buildRes, err := r.buildAPI.Sync(buildCtx, *mod, *km, kernelVersion, containerImage, true)
+	buildRes, err := r.buildAPI.Sync(buildCtx, *mod, *km, kernelVersion, true, mod)
 	if err != nil {
 		return false, fmt.Errorf("could not synchronize the build: %w", err)
 	}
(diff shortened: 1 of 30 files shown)


internal/build/job/manager.go
```diff
@@ -101,3 +128,13 @@ func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1b
 		return build.Result{}, fmt.Errorf("unknown status: %v", job.Status)
 	}
 }
+
+func (jbm *jobManager) checkIfImageExists(ctx context.Context, mod *kmmv1beta1.Module, km *kmmv1beta1.KernelMapping, imageName string) (bool, error) {
```
qbarrand, line 132: This is a duplicate of the method of the same name in `controllers/module_reconciler.go`. I don't know where would be a good location, though.

mresvanis: Yes, valid point. I'm currently thinking about where this could be hosted, that's why it's still in both places. I'll come back to this one.

controllers/module_reconciler.go
```diff
@@ -305,40 +285,14 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
...
-
 func (r *ModuleReconciler) handleSigning(ctx context.Context,
 	mod *kmmv1beta1.Module,
 	km *kmmv1beta1.KernelMapping,
 	kernelVersion string,
 	containerImage string) (bool, error) {
 
 	// if sign is not specified skip
-	if !r.willSign(mod, km) {
+	if !module.ShouldBeSigned(mod, km) {
```
yevgeny-shnaidman, line 295: Why checkImageExists removed from handleBuild and left in handleSigning?

internal/build/job/manager.go
```diff
@@ -52,6 +58,27 @@ func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1b
 
 	logger.Info("Building in-cluster")
 
+	// if theres no build specified skip
+	if !module.ShouldBeBuilt(&mod, &m) {
```
yevgeny-shnaidman, line 62: I would prefer if ShouldBeBuild, ShouldBeSigned and checkImageExists will not be in the Sync function.  There are other flows ( like preflight), that has done image verification in a different way, and just want to execute the Build flow, without running the checks again. WDYT about moving those verification to a separate API? Then this API can be called prior to calling Sync in a regular flow, and not called in the flow that does not need it. Or add and API that call those checks and then calls the Sync API

internal/registry/registry.go
```diff
@@ -70,6 +70,16 @@ func NewRegistry() Registry {
 	return &registry{}
 }
 
+// AppendToTag adds the specified tag to the image name cleanly, i.e. by avoiding messing up
+// the name or getting "name:-tag"
+func AppendToTag(name string, tag string) string {
```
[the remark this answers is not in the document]
mresvanis: I don't have a strong opinion on this, so in order to leave the `registry` package as is (it's already big enough, but focused as you said on the image registry functionality), I've moved this helper function to the `module` package helpers :+1:

controllers/module_reconciler.go
mresvanis: There are 2 reasons why this is for the time being not refactored yet:

- I wanted to confirm the refactoring effort with the `handleBuild` business logic first and then move to the `handleSignging` refactoring
- there is an open [question](https://github.com/kubernetes-sigs/kernel-module-management/pull/156#discussion_r1026542716) about where to host the `checkIfImageExists` method/function

So I will soon come back to this one. Does this make sense?

yevgeny-shnaidman: Sure! Are you going to do it in a different PR, or later in this one?

mresvanis: I'm targeting this PR. As soon as we have covered the rest of the review comments, I'll go for it.

internal/build/job/manager.go
mresvanis: I moved this method under the `module` package, [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/156/files#diff-4c273cc04ac4611a608fe8e4037b9834c162daba892070611f56ffc83e47fc4dR49). What do you think?

internal/build/job/manager.go
yevgeny-shnaidman: great, thank you!

controllers/module_reconciler.go
```diff
@@ -264,33 +259,20 @@ func (r *ModuleReconciler) getNodesListBySelector(ctx context.Context, mod *kmmv
...
-		return false, fmt.Errorf("failed to check existence of image %s for kernel %s: %w", containerImage, kernelVersion, err)
+		return false, fmt.Errorf("could not check if build synchronization is needed: %w", err)
 	}
-	if exists {
+	if !shouldSync {
 		return false, nil
 	}
 
-	logger := log.FromContext(ctx).WithValues("kernel version", kernelVersion, "image", containerImage)
+	logger := log.FromContext(ctx).WithValues("kernel version", kernelVersion, "image", km.ContainerImage)
```
qbarrand, line 272: Not sure how that would play out if the `containerImage` is defined at the top-level, i.e. if `km.ContainerImage` is empty. Probably outside of the scope of this PR because I don't think it introduces a bug, but the code could use some consolidation there.

mresvanis: This is a valid point and I think that we have ourselves a bug here, although we have to test and verify it. The bug being: the `ModuleLoaderContainer.Spec.ContainerImage` field is never used, we only use the `ModuleLoaderContainerSpec.KernelMappings[].ContainerImage` field instead. 

I suggest we open an issue for this bug, verify it and resolve it in a separate PR. WDYT?

internal/build/job/maker.go
```diff
@@ -18,13 +14,37 @@ import (
...
+
+	TargetKernel string
+
+	ContainerImage string
+
+	ShouldPushImage bool
+
+	RegistryTLS *kmmv1beta1.TLSOptions
+
+	Owner metav1.Object
```
qbarrand, line 43: Could we remove the empty lines? Would make everything nice and aligned 🙂

mresvanis: This change has been reverted, so only for the sake of argument, the empty lines are pretty useful IMHO. At first they might seem a bit out of order, but they help a lot regarding version control and the future diffs, for example when one of those lines is removed or another one is added. 

For this specific change, I should have also added the doc strings for every one of those exported fields. This way everything would be clearer, more maintainable and easier to diff in the future.

internal/sign/job/manager.go
```diff
@@ -5,40 +5,82 @@ import (
...
-func (jbm *signJobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, imageToSign string, targetImage string, pushImage bool) (utils.Result, error) {
+func (jbm *signJobManager) Sync(
+	ctx context.Context,
+	modSpec kmmv1beta1.ModuleSpec,
+	owner metav1.Object,
+	m kmmv1beta1.KernelMapping,
+	targetKernel string,
+	imageToSign string,
+	targetImage string,
+	pushImage bool) (utils.Result, error) {
```
qbarrand, line 69: How about adding variadic options to that function as well? Ultimately we will also have the TLS options here.

qbarrand (commented):
Final (and rather minor) comments. Thank you!

internal/build/job/maker.go
```diff
@@ -48,8 +84,89 @@ func NewMaker(client client.Client, helper build.Helper, jobHelper utils.JobHelp
...
+func (m *maker) MakeJobTemplate(
+	ctx context.Context,
+	modSpec kmmv1beta1.ModuleSpec,
+	buildConfig *kmmv1beta1.Build,
+	targetKernel string,
+	containerImage string,
+	owner metav1.Object,
+	opts ...JobTemplateOption) (*batchv1.Job, error) {
+
+	var options jobTemplateOptions
```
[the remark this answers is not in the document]
mresvanis: Totally agree, it's clearer :+1:

internal/build/job/manager.go
```diff
@@ -47,19 +58,66 @@ func (jbm *jobManager) GarbageCollect(ctx context.Context, mod kmmv1beta1.Module
...
+	// if build AND sign are specified, then we will build an intermediate image
+	// and let sign produce the one specified in targetImage
+	if module.ShouldBeSigned(modSpec, m) {
+		targetImage = module.IntermediateImageName(owner.GetName(), owner.GetNamespace(), targetImage)
+	}
 
-	jobTemplate, err := jbm.maker.MakeJobTemplate(ctx, mod, buildConfig, targetKernel, targetImage, pushImage, m.RegistryTLS)
+	jobTemplate, err := jbm.maker.MakeJobTemplate(ctx, modSpec, buildConfig, targetKernel, targetImage, owner,
+		PushImage(pushImage),
+		RegistryTLS(m.RegistryTLS),
```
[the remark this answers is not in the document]
mresvanis: Got your point, let me try it out and come back to you.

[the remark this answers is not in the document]
mresvanis: WDYT about [this](https://github.com/kubernetes-sigs/kernel-module-management/pull/156/files#diff-0e82e65de93c2cf1120b942e748d056de42d3fcaf08317e832dfbe74069bd06dR125-R133)?

internal/build/job/maker.go
```diff
@@ -18,13 +14,62 @@ import (
...
+// empty no image is pushed.
+func ContainerImage(containerImage string) JobTemplateOption {
+	return func(opts *jobTemplateOptions) {
+		opts.containerImage = containerImage
+	}
+}
+
+// RegistryTLS configures the TLSOptions to be used when
+// interacting with the registry.
+func RegistryTLS(registryTLS *kmmv1beta1.TLSOptions) JobTemplateOption {
```
yevgeny-shnaidman, line 49: is this functions used somewhere beside job package? If not, then it should be private

internal/build/job/maker.go
```diff
@@ -18,13 +14,62 @@ import (
...
+	return func(opts *jobTemplateOptions) {
+		opts.registryTLS = registryTLS
+	}
+}
+
+// Owner configures the controller reference owner to be set on the build
+// Job created. When no owner is specified, the kmmv1betav1.Module is by
+// default the owner of the build Job.
+func Owner(owner metav1.Object) JobTemplateOption {
```
yevgeny-shnaidman, line 58: The same comment: if not used outside of the package, it should be private

internal/build/job/manager.go
```diff
@@ -47,19 +57,85 @@ func (jbm *jobManager) GarbageCollect(ctx context.Context, mod kmmv1beta1.Module
...
+	if options.Owner == nil {
+		options.Owner = &mod
+	}
+
 	logger := log.FromContext(ctx)
 
 	logger.Info("Building in-cluster")
 
-	buildConfig := jbm.helper.GetRelevantBuild(mod, m)
+	buildConfig := jbm.helper.GetRelevantBuild(mod.Spec, m)
```
yevgeny-shnaidman, line 117: WDYT about moving all the decision for buildConfig and job Option into the MakeJobTemplate function of the maker?  Since that function create and Job and makes all the decision regarding how it should be set, maybe it needs to take those decisions also. this way the Sync code will be more clear and readable and will deal only with managing the job

internal/build/manager.go
```diff
@@ -19,9 +21,41 @@ type Result struct {
 	Status  Status
 }
 
+// SyncOptions configures optional arguments of the Sync call.
+type SyncOptions struct {
+	// ContainerImage configures the image to be pushed. When left
+	// empty no image is pushed.
+	ContainerImage string
```
yevgeny-shnaidman, line 28: WDYT regarding changing this field to boolean, and renaming it to something like pushImage? we have container imahge in the kmm mapping and in mod.Spec, and it might be confusing

internal/build/manager.go
```diff
@@ -19,9 +21,41 @@ type Result struct {
...
+
+	// Owner configures the controller reference owner to be set on the build
+	// Job created. When no owner is specified, the kmmv1betav1.Module is by
+	// default the owner of the build Job.
+	Owner metav1.Object
```
yevgeny-shnaidman, line 33: Great idea! I already see how it will help me with some issues in the preflight

internal/sign/job/manager.go
```diff
@@ -5,40 +5,102 @@ import (
...
+		options.Owner = &mod
+	}
+
 	logger := log.FromContext(ctx)
 
 	logger.Info("Signing in-cluster")
 
-	signConfig := jbm.helper.GetRelevantSign(mod, m)
-	jobLabels := jbm.jobHelper.JobLabels(mod, targetKernel, "sign")
+	signConfig := jbm.helper.GetRelevantSign(mod.Spec, m)
```
yevgeny-shnaidman, line 87: The same comment as in Build: WDYT regarding moving all the jobOptions into the MakeJobTemplate?

internal/sign/manager.go
```diff
@@ -3,10 +3,70 @@ package sign
...
+
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 )
 
+// SyncOptions configures optional arguments of the Sync call.
+type SyncOptions struct {
+	// TargetImage configures the signed image to be pushed. When left
+	// empty no image is pushed, the signed files are just listed.
+	TargetImage string
```
yevgeny-shnaidman, line 16: The same comment as for the build: WDYT about rename it and making it boolean?

internal/build/job/maker.go
```diff
@@ -18,13 +14,62 @@ import (
...
+	owner metav1.Object
+}
+
+// JobTemplateOption configures MakeJobTemplate.
+type JobTemplateOption func(*jobTemplateOptions)
+
+// ContainerImage configures the image to be pushed. When left
+// empty no image is pushed.
+func ContainerImage(containerImage string) JobTemplateOption {
+	return func(opts *jobTemplateOptions) {
```
ybettan, line 42: Why are we returning a function instead of the object? Why are we delaying the execution of `opts.containerImage = containerImage` to when the function will be executed rather than just now?

internal/build/manager.go
qbarrand: A boolean is what we had before (`pushImage`). Since image pushing is optional, I think it makes more sense to make the target image name an optional option.

internal/sign/manager.go
qbarrand: See https://github.com/kubernetes-sigs/kernel-module-management/pull/156#discussion_r1037104809

yevgeny-shnaidman: ok

internal/build/manager.go
yevgeny-shnaidman: ok

internal/build/job/manager.go
```diff
@@ -47,19 +54,75 @@ func (jbm *jobManager) GarbageCollect(ctx context.Context, mod kmmv1beta1.Module
 	return deleteNames, nil
 }
 
-func (jbm *jobManager) Sync(ctx context.Context, mod kmmv1beta1.Module, m kmmv1beta1.KernelMapping, targetKernel string, targetImage string, pushImage bool) (build.Result, error) {
+func (jbm *jobManager) ShouldSync(
+	ctx context.Context,
+	mod kmmv1beta1.Module,
+	m kmmv1beta1.KernelMapping,
+	opts ...build.SyncOption) (bool, error) {
```
yevgeny-shnaidman, line 61: why not just pass a namespace parameter to this function? We are construction all those options, but all we care about is just a namespace. The code that is calling this function should now which namespace to use ( it is probably the namespace of the Module)

internal/build/manager.go
```diff
@@ -19,9 +21,47 @@ type Result struct {
 	Status  Status
 }
 
+// SyncOptions configures optional arguments of the Sync call.
+type SyncOptions struct {
```
yevgeny-shnaidman, line 25: Do we expect the SyncOption fields to be expanded in the future?

internal/preflight/preflight.go
```diff
@@ -137,7 +137,7 @@ func (p *preflightHelper) verifyBuild(ctx context.Context,
 	mapping *kmmv1beta1.KernelMapping,
 	mod *kmmv1beta1.Module) (bool, string) {
 	// at this stage we know that eiher mapping Build or Container build are defined
-	buildRes, err := p.buildAPI.Sync(ctx, *mod, *mapping, pv.Spec.KernelVersion, mapping.ContainerImage, pv.Spec.PushBuiltImage)
+	buildRes, err := p.buildAPI.Sync(ctx, *mod, *mapping, pv.Spec.KernelVersion, pv.Spec.PushBuiltImage, mod)
```
yevgeny-shnaidman, line 140: the last parameter should be pv instead of mod

mresvanis: Nice catch! Fixed.

PR #159, opened 2022-11-22
Migrate Prebuilt kernel module e2e job to Prow

enriquebelarte:
In order to sync CI jobs with m/s and make code
reusable through components, actions scripts
are migrated and simplified in two scripts, one
responsible for creating infrastructure and another one for tests.


ci/github/create_infra
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+build_cluster () {
+START_ARGS="$@"
+wget https://storage.googleapis.com/minikube/releases/latest/minikube_latest_amd64.deb
+sudo dpkg -i ./minikube_latest_amd64.deb
+#echo "{version}={$(minikube version --short)}" >> $GITHUB_OUTPUT
+minikube start --driver=docker $START_ARGS
+kubectl wait --for=condition=available deployment coredns -n kube-system
```
ybettan, line 9: Can we fix the indention?

ci/github/build_driver_container
```diff
@@ -0,0 +1,11 @@
...
+set -euxo pipefail
+
+build_driver_container () {
+      cd ci/kmm-kmod && make KERNEL_SRC_DIR="/usr/src/linux-headers-$(uname -r)" && \
+      docker build -t kmm-kmod:local --build-arg=KERNEL_VERSION=$(uname -r) . && \
+      minikube image load kmm-kmod:local 
+}
+
+build_driver_container
+
```
ybettan, line 11: IMO, we should have one file to contain everything needed in order for preparing the infra.
Having function in it is great but no need to split it into multiple files.

Also, this should be part of the infra step as the m/s CI will build all the images for us.

ci/github/build_controller
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+build_controller () {
+       minikube image build -t kmm:local .
```
ybettan, line 5: That's a nice one!

ci/github/build_driver_container
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+build_driver_container () {
+      cd ci/kmm-kmod && make KERNEL_SRC_DIR="/usr/src/linux-headers-$(uname -r)" && \
+      docker build -t kmm-kmod:local --build-arg=KERNEL_VERSION=$(uname -r) . && \
+      minikube image load kmm-kmod:local
```
ybettan, line 7: Here is a good place to use `minikube image build` as we don't have any Makefile target for `kmm-kmod`.

ci/github/collect_troubleshooting
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euxo pipefail
+collect_troubleshooting () {
+      kubectl get all -n kmm-operator-system
+      kubectl get all
+      kubectl describe node
```
ybettan, line 6:
```suggestion
      kubectl describe nodes
```
(Just because it is more clear) - the output is the same

ci/github/build_driver_container
enriquebelarte: Could not find a way to make `minikube image build` work with arguments. That's why I splitted it. Any suggestion on how to make it work will be much appreciated.

ci/github/collect_troubleshooting
```diff
@@ -0,0 +1,13 @@
...
+      kubectl describe daemonset -l kmm.node.kubernetes.io/module.name
+      kubectl describe pod -l kmm.node.kubernetes.io/module.name
+      sudo dmesg
```
ybettan, line 9: How does that work on an automated flow? Won't it require some user input?

ci/github/create_infra
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+build_cluster () {
+START_ARGS="$@"
+wget https://storage.googleapis.com/minikube/releases/latest/minikube_latest_amd64.deb
+sudo dpkg -i ./minikube_latest_amd64.deb
```
ybettan, line 6: Aren't we root on the github runner? Won't it idle and wait for user input?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env bash
```
ybettan, line 1: What is the difference between that and `#!/usr/bin/bash`?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+set -euxo pipefail
+
+echo "Label the node so that it gets the module..."
+kubectl label node minikube wants-module=kmm_ci_a
```
ybettan, line 6: Do we really need to label the node? Can we just tell it to run on all workers/masters?

Here is a selector example that does that:
```
  selector:
    node-role.kubernetes.io/worker:
```

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Build and deploy the controller image..."
+./ci/github/build_controller
+
+echo "Build and deploy the DriverContainer..."
```
ybettan, line 11: Isn't it just building? Anyway it should be part of the infra, otherwise we are going to re-build an already built image m/s.

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+kubectl label node minikube wants-module=kmm_ci_a
+
+echo "Build and deploy the controller image..."
+./ci/github/build_controller
+
+echo "Build and deploy the DriverContainer..." 
+./ci/github/build_driver_container
+
+echo "Deploy KMMO..."
+make deploy KUSTOMIZE_CONFIG_DEFAULT=ci/install-ci
```
ybettan, line 15: How about we use the default tags when building the images and just use `make deploy`? If the `imagePullPolicy` is `ifNotPresent` then we should use the correct image. 

WDYT?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+
+echo "Wait until the KMMO Deployment is Available..."
+kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
```
ybettan, line 19: Is this needed? I don't mind keeping it but I think the controller will now how to reconcile even if the module is created before the operator is deployed.

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Describe the Deployment and get its YAML if that failed..."
+kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
+kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
+#        if: ${{ failure() }}
```
ybettan, line 24: I think we should move it to the infra part with all the other data gathered.

Also, IMO, we should call it only on failures.

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Check that the kmm_ci_a module is not loaded on the node..."
```
ybettan, line 26: Are we deploying more than one module? is the name `kmm_ci_a` still a valid one?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+          if minikube ssh -- lsmod | grep kmm_ci_a; then
+            echo "Unexpected lsmod output - the module should not be loaded"
+            exit 1
+          fi
+
+echo "Add an kmm-ci Module that contains a valid mapping..."
```
ybettan, line 32:
```suggestion
echo "Add a kmm-ci Module that contains a valid mapping..."
```

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+          sed -e "s/KVER_CHANGEME/$(uname -r)/g" \
+            -e s/NAME_CHANGEME/kmm-ci/ \
+            -e s/KMOD_CHANGEME/kmm_ci_a/ \
+            ci/module-kmm-ci.template.yaml | tee module-kmm-ci.yaml
+          kubectl apply -f module-kmm-ci.yaml
```
ybettan, line 37:
```suggestion
            ci/module-kmm-ci.template.yaml | kubectl apply -f -
```

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Check that the module gets loaded on the node..."
```
ybettan, line 39:
```suggestion
echo "Check that the kernel module gets loaded on the node..."
```

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+          until minikube ssh -- lsmod | grep kmm_ci_a; do
+            sleep 3
+          done
```
ybettan, line 42: What happens if it is not present? Should we add a timeout?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Check that the node gets labeled with the module's name..."
+          until kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do
+            sleep 3
+          done
```
ybettan, line 47: What happens if it is not present? Should we add a timeout?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Check that the daemon-set for device plugin is running..."
+          until kubectl get ds  -l 'kmm.node.kubernetes.io/module.name=kmm-ci,kmm.node.kubernetes.io/role=device-plugin' -ojson | jq -e '.items[] | select(.status.numberReady == 1)'; do 
+            sleep 3
+          done
```
ybettan, line 52: Why do we need to check it? if the daemonset doesn't exist then the kernel module won't be loaded.

I think we should only test input-output stuff and not be tightly coupled with the internal implementation (what if we stop using daemonsets for example?)

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+echo "Remove the Module..."
+kubectl delete -f module-kmm-ci.yaml
+
+echo "Check that the module gets unloaded from the node..."
+          until ! minikube ssh -- lsmod | grep kmm_ci_a; do
+            sleep 3
+          done
```
ybettan, line 59: Timeout?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,66 @@
...
+
+echo "Check that the node gets unlabeled with the module's name..."
+          until ! kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do
+            sleep 3
+          done
```
ybettan, line 64: Timeout?

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
```
ybettan, line 3: Some the comments from [e2e_prebuilt_kernel_module](https://github.com/kubernetes-sigs/kernel-module-management/pull/159/files#diff-9c62c51855541d7fc09bf9d7021b7f647169e182aac4396e55c9e58f1bd86eca)` applies here as well.

Here I will just add additional comments if there are such.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,76 @@
...
+          sed -e s/NAME_CHANGEME/kmm-ci-a/g \
+            -e s/KMOD_CHANGEME/kmm_ci_a/g \
+            -e "s/KVER_CHANGEME/$(uname -r)/g" \
+            ci/module-kmm-ci.template.yaml | tee module-kmm-ci-a.yaml
+
+          # Node minikube-m02 gets module b
+          sed -e s/NAME_CHANGEME/kmm-ci-b/g \
+            -e s/KMOD_CHANGEME/kmm_ci_b/g \
+            -e "s/KVER_CHANGEME/$(uname -r)/g" \
+            ci/module-kmm-ci.template.yaml | tee module-kmm-ci-b.yaml
```
ybettan, line 40: The kernel version is the same on the 2 modules, do we really need 2? Can we use the samemodule on the 2 nodes?

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,76 @@
...
+
+          kubectl apply -f module-kmm-ci-a.yaml -f module-kmm-ci-b.yaml
+
+
+echo "Label the first node to have it receive module a..."
+kubectl label node minikube wants-module=kmm_ci_a
```
ybettan, line 46: Here we do need to label because we want to differ between the 2 nodes.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,76 @@
...
+echo "Check that module b gets unloaded from the second node..."
+          until ! minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do
+            sleep 3
+          done 
+
+echo "Verify that the DaemonSet gets garbage collected..."
+          # Cannot use kubectl wait because it will fail if there is no initial match
+          until [ $(kubectl get daemonsets.apps -l 'kmm.node.kubernetes.io/module.name=kmm-ci-b, kmm.node.kubernetes.io/role!=device-plugin' -o go-template='{{ len .items }}') -eq 0 ]; do
+            sleep 3
+          done
```
ybettan, line 74: Not sure we need that, if the kernel module was removed then we mush have cleaned the daemonset isn't it? This is implementation details IMO.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,76 @@
...
+
+echo "Label the second node to have it receive module b..."
+kubectl label node minikube-m02 wants-module=kmm_ci_b
+
+echo "Check that module b gets loaded on the second node..."
+          until minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do
+            sleep 3
+          done
+
+echo "Remove the wants-module label on the second node..."
```
ybettan, line 62:
```suggestion
echo "Remove the wants-module label on one of the nodes ..."
```

Doesn't really matter which node it is.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,76 @@
...
+kubectl label node minikube-m02 wants-module-
+
+echo "Check that module b gets unloaded from the second node..."
```
ybettan, line 65:
```suggestion
echo "Check that module b gets unloaded from that node..."
```

Doesn't really matter which node it is. (it just makes you wonder what is special about the "second node").

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,35 @@ env:
...
-
-      - name: Cache binaries needed by Makefile
-        id: cache-bin
-        uses: actions/cache@v3
-        with:
-          path: ./bin
-          key: ${{ runner.os }}-bin-${{ env.GO_VERSION }}-${{ hashFiles('Makefile') }}
+        #Arguments for minikube can be set after create_cluster command as they will be taken
+        #by the script
+        run: ./ci/github/create_infra
```
ybettan, line 21: How about keeping that part in another job (previously named `build-operator-image`) so it is called only once  as a required job for all other jobs?

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,35 @@ env:
...
     name: Prebuilt kernel module
 
-    needs: [build-operator-image]
-
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
 
+      - uses: actions/checkout@v3
       - name: Create the minikube cluster
```
ybettan, line 18:
```suggestion
      - name: Create the test's infra
```

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,35 @@ env:
...
-          minikube image load kmm-kmod_local.tar
-
-      - name: Cache binaries needed by Makefile
-        id: cache-bin
-        uses: actions/cache@v3
-        with:
-          path: ./bin
-          key: ${{ runner.os }}-bin-${{ env.GO_VERSION }}-${{ hashFiles('Makefile') }}
+        #Arguments for minikube can be set after create_cluster command as they will be taken
+        #by the script
```
ybettan, line 20:
```suggestion
```

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,35 @@ env:
...
+        run: ./ci/github/create_infra
 
-      - uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
-        if: steps.cache-bin.outputs.cache-hit != 'true'
+      - name: Run Tests
```
ybettan, line 23:
```suggestion
      - name: Run e2e_prebuilt_kernel_module
```

ci/github/build_driver_container
ybettan: Can we use the default image name/tag as suggested [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/159#discussion_r1029168571)? In that case you don't need any argument.

ci/prow/e2e_prebuilt_kernel_module
qbarrand: https://stackoverflow.com/a/16365367

ci/prow/e2e_prebuilt_kernel_module
qbarrand: I think that's worth trying but maybe in a subsequent PR?

ci/prow/e2e_prebuilt_kernel_module
qbarrand: If the controller fails to deploy for any reason, that step will tell us very early and clearly in the flow.

ci/prow/e2e_prebuilt_kernel_module
qbarrand: The DriverContainer image we create contains two kmods: https://github.com/kubernetes-sigs/kernel-module-management/tree/main/ci/kmm-kmod
The name is thus valid.

ci/prow/e2e_prebuilt_kernel_module
qbarrand: It would be great to have a timeout, but I am not sure how to implement it in bash.

ci/prow/e2e_prebuilt_kernel_module
qbarrand: This is to verify that the device plugin DaemonSet is running, I don't think there is any other way to make that check.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
qbarrand: I think it adds clarity to the flow. Node 1 receives module A, node 2 receives module B. Useful when trying to understand the test.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
qbarrand: How do we verify that garbage collection works? This is an important feature.

.github/workflows/e2e.yaml
qbarrand: Why change that and give less information?

ci/prow/e2e_prebuilt_kernel_module
enriquebelarte: Probably this would work?

https://man7.org/linux/man-pages/man1/timeout.1.html

.github/workflows/e2e.yaml
ybettan: Because: 
* we are not only creating the cluster, we are also building the images.
* minikube is just an implementation BE and may change

ci/github/collect_troubleshooting
ybettan: And what happens if we drop `sudo`? Will it fail?

ci/prow/e2e_prebuilt_kernel_module
ybettan: Sure, I don't mind in which PR it is. Just wish that it will be as close as possible to what a contributor is used to run locally to make the tests easier to understand.

ci/prow/e2e_prebuilt_kernel_module
ybettan: Oh I missed the `for device plugin` in the description.
Let's keep it then.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
ybettan: But in that case, we could have check only the "first node" instead of the second node right?! there is nothing special about the "second node".

I don't mind keeping it, was just confusing to me cause I searched what is happening with the "first node"...

ci/prow/e2e_prebuilt_kernel_module
ybettan: Looks like github has its own timeout https://github.com/kubernetes-sigs/kernel-module-management/actions/runs/3533048741/jobs/5928212666 but maybe it is better to have our own.

ci/github/collect_troubleshooting
enriquebelarte: Yes it fails as the builder user in the worker has no privileges:

```
dpkg: error: requested operation requires superuser privilege
```

ci/prow/e2e_prebuilt_kernel_module_two_nodes
qbarrand: Well, this is exactly what this `until` does: wait until the DaemonSets get garbage-collected. And it's from the existing workflow, so it's part of the test.

.github/workflows/e2e.yaml
enriquebelarte: Did not exactly the same, because requiring `build_images()` would fail when using minikube within the function as it was not installed in that first step. Kind of similar except that I like the idea of two different functions for the build of different images `build_controller()` and `build_driver_container()`

ci/prow/e2e_prebuilt_kernel_module_two_nodes
ybettan: Ohh sorry, I completely misunderstood that part.
We definitely needs it.

Sorry for the noise.

ci/github/collect_troubleshooting
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euxo pipefail
+collect_troubleshooting () {
```
ybettan, line 3: Why do we need this? We already have `collect_troubleshooting` in `create_infra`.

ci/github/create_infra
```diff
@@ -0,0 +1,35 @@
...
+      kubectl wait --for=condition=available deployment coredns -n kube-system
+}
+
+build_controller () {
+      make docker-build  
+}
+
+build_driver_container () {
+      cd ci/kmm-kmod && make KERNEL_SRC_DIR="/usr/src/linux-headers-$(uname -r)" && \
+      minikube image build -t kmm-kmod:local --build-opt build-arg=KERNEL_VERSION=$(uname -r) .
```
ybettan, line 19:
```suggestion
      cd ci/kmm-kmod
      make KERNEL_SRC_DIR="/usr/src/linux-headers-$(uname -r)"
      minikube image build -t kmm-kmod:local --build-opt build-arg=KERNEL_VERSION=$(uname -r) . 
```

WDYT?

ci/github/create_infra
```diff
@@ -0,0 +1,35 @@
...
+collect_troubleshooting () {
+      kubectl get all -n kmm-operator-system
+      kubectl get all
+      kubectl describe node
+      kubectl describe daemonset -l kmm.node.kubernetes.io/module.name
+      kubectl describe pod -l kmm.node.kubernetes.io/module.name
+      sudo dmesg
+      kubectl logs deployment.apps/kmm-operator-controller-manager -n kmm-operator-system
+      # If fails
+      echo "Describe the Deployment and get its YAML if that failed..."
```
ybettan, line 31: If what "failed"? I think all this data should be printed only if the workflow fail. I don't see a reason to differ between all the troubleshooting data above and the next few lines.

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,55 @@
...
+set -euxo pipefail
+
+echo "Label the node so that it gets the module..."
+kubectl label node minikube wants-module=kmm_ci_a
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available..."
+kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
```
ybettan, line 12:
```suggestion
echo "Deploy KMMO..."
make deploy
kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
```

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,55 @@
...
+
+
+echo "Check that the kmm_ci_a module is not loaded on the node..."
+          if minikube ssh -- lsmod | grep kmm_ci_a; then
+            echo "Unexpected lsmod output - the module is present on the node before the module was applied to the cluster"
+            exit 1
+          fi
```
ybettan, line 19: Why do we have an indention here?

ci/prow/e2e_prebuilt_kernel_module
```diff
@@ -0,0 +1,55 @@
...
+
+echo "Add a kmm-ci Module that contains a valid mapping..."
+          sed -e "s/KVER_CHANGEME/$(uname -r)/g" \
+            -e s/NAME_CHANGEME/kmm-ci/ \
+            -e s/KMOD_CHANGEME/kmm_ci_a/ \
+            ci/module-kmm-ci.template.yaml | tee module-kmm-ci.yaml
+          kubectl apply -f module-kmm-ci.yaml
```
ybettan, line 26: Not sure why we have all those indentions.

ci/prow/e2e_prebuilt_kernel_module_two_nodes
ybettan: How about moving this part to the "classic flow"? `e2e_prebuilt_kernel_module`

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,71 @@
...
+set -euxo pipefail
+
+echo "Label the node so that it gets the module..."
+kubectl label node minikube wants-module=kmm_ci_a
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available..."
+kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
```
ybettan, line 12:
```suggestion
echo "Deploy KMMO..."
make deploy
kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
```

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,71 @@
...
+
+
+echo "Describe the Deployment and get its YAML if that failed..."
+kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
+kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
+
+echo "Check that the kmm_ci_a module is not loaded on the node..."
+          if minikube ssh -- lsmod | grep kmm_ci_a; then
+            echo "Unexpected lsmod output - the module should not be loaded"
```
ybettan, line 21: Can we update the error message to the same as the previous test?

ci/prow/e2e_prebuilt_kernel_module_two_nodes
```diff
@@ -0,0 +1,71 @@
...
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available..."
+kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
+
+
+echo "Describe the Deployment and get its YAML if that failed..."
+kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
+kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
```
ybettan, line 17: Can we call the `collect_troubleshooting` instead here?
Also, can we call it only if the flow failed?

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,33 @@ env:
...
-        with:
-          kernel-version: ${{ env.KERNEL_VERSION }}
 
-      - name: Import images into minikube
-        run: |
-          minikube image load kmm_local.tar
-          minikube image load kmm-kmod_local.tar
+      - uses: actions/checkout@v3
+      - name: Create the test's infra 
+        run: source ./ci/github/create_infra && build_cluster && build_controller && build_driver_container
```
ybettan, line 20:
```suggestion
        run: |
          source ./ci/github/create_infra
          build_cluster
          build_controller
          build_driver_container
```

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,33 @@ env:
...
-        uses: ./.github/actions/build-drivercontainer-image
-        with:
-          kernel-version: ${{ env.KERNEL_VERSION }}
 
-      - name: Import images into minikube
-        run: |
-          minikube image load kmm_local.tar
-          minikube image load kmm-kmod_local.tar
+      - uses: actions/checkout@v3
+      - name: Create the test's infra
```
ybettan, line 19: Can we extract it to a separate job so it is not called for each job?

.github/workflows/e2e.yaml
```diff
@@ -6,522 +6,33 @@ env:
...
-          until minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Remove the Module
-        run: kubectl delete -f ci/module-kmm-ci-variable.yaml
+      - uses: actions/checkout@v3
+      - name: Create the test's infra
+        run: source ./ci/github/create_infra --nodes 2 && build_cluster && build_controller && build_driver_container
```
ybettan, line 34: Can we extract it?

enriquebelarte:
Close as all jobs will be migrated by this [PR](https://github.com/kubernetes-sigs/kernel-module-management/pull/166) at once.

Issue #115, opened 2022-10-19
decouple building process from Module CRD/CR


uMartinXu:
### Issue Summary:
The building process should be decoupled from Module CRD/CR for driver version control feature 
### Suggested Priority(P1-P3) & Urgent(Urgent, medium, Low):
P2 & Medium
### Issue Detail:
Let start from the the roles and players. Normally there are three players and roles in the whole KMMO picture.
1. Cluster Administrators who use KMMO to ( build ) deploy and manage the OOT drivers.
2. Driver developer who develop and release the driver code
3. KMMO developer who develop and maintain the KMMO operator in some K8S platform
Currently, the Driver building process is embedded into Module CR yaml, and this CR yaml file owner should be cluster Administrators. But the driver build process should be maintained by Driver developer, and the building process might change to different version of driver. That is out of the Cluster Administrators' knowledge scope. So to maintain the CR yaml and let it match with each driver version should not be handled by Cluster Administers.  We should find a way to decouple building process from Module CRD/CR, and let the build process can natively match with each driver version.
### Solution Proposal
Let driver developer maintain a building process which can be docker file or scripts for each driver version in the driver project. With the moduleVersion and moduleProjectURL KMMO knows which build process can be used to build the driver container image. So clusterAdminstrators don't need to maintain the build process in the module CRD/CR yaml file anymore

qbarrand:
I think Preflight could be a solution to that problem - it makes KMM behave as if it had seen a node with a new kernel. An administrator using that feature could build images targeting kernel versions that are not currently running in the cluster. WDYT?
@yevgeny-shnaidman does that sound right to you?

qbarrand:
@uMartinXu @hershpa @chaitanya1731 just to confirm that extracting the Dockerfile out into a ConfigMap would make sense? Then the `build` section of the CRD would have a field like `configMapName` that would point to that ConfigMap.

enriquebelarte:
FYI. There's a scheduled kmm technical enablement in a few days which relies heavily on [labs](https://github.com/kubernetes-sigs/kernel-module-management/tree/main/docs/labs).  
This change would break all examples in labs.

qbarrand:
@uMartinXu @hershpa @chaitanya1731 I believe this was implemented in #160. OK to close?

PR #160, opened 2022-11-22
Moving Build Dockerfile to be defined in the ConfigMap

yevgeny-shnaidman:
Currently dockerfile is defined a multiline string in the Build field of the KernelMapping. This PR does the following: 1) dockerfile is now defined in a dedicated configmap (created by customer)
   with the key being: dockerfile
2) Build struct now contains reference to the ConfigMap, instead of
   a string
3) change the flow of e2e workflow for the in-cluster build
   to use the new configuration

### DIFF
--- a/api/v1beta1/module_types.go
+++ b/api/v1beta1/module_types.go
@@ -60,7 +60,8 @@ type Build struct {
 	// BuildArgs is an array of build variables that are provided to the image building backend.
 	BuildArgs []BuildArg `json:"buildArgs"`
 
-	Dockerfile string `json:"dockerfile"`
+	// ConfigMap that holds Dockerfile contents
+	DockerfileConfigMap *v1.LocalObjectReference `json:"dockerfileConfigMap"`
 
 	// +optional
 	// Pull contains settings determining how to pull the base images of the build process.
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -95,6 +95,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list
 //+kubebuilder:rbac:groups="core",resources=serviceaccounts,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch;delete
 
--- a/internal/build/helper.go
+++ b/internal/build/helper.go
@@ -47,7 +47,6 @@ func (m *helper) ApplyBuildArgOverrides(args []v1beta1.BuildArg, overrides ...v1
 
 func (m *helper) GetRelevantBuild(mod kmmv1beta1.Module, km kmmv1beta1.KernelMapping) *kmmv1beta1.Build {
 	if mod.Spec.ModuleLoader.Container.Build == nil {
-		// km.Build cannot be nil in case mod.Build is nil, checked above
 		return km.Build.DeepCopy()
 	}
 
@@ -56,9 +55,7 @@ func (m *helper) GetRelevantBuild(mod kmmv1beta1.Module, km kmmv1beta1.KernelMap
 	}
 
 	buildConfig := mod.Spec.ModuleLoader.Container.Build.DeepCopy()
-	if km.Build.Dockerfile != "" {
-		buildConfig.Dockerfile = km.Build.Dockerfile
-	}
+	buildConfig.DockerfileConfigMap = km.Build.DockerfileConfigMap
 
 	buildConfig.BuildArgs = m.ApplyBuildArgOverrides(buildConfig.BuildArgs, km.Build.BuildArgs...)
 
--- a/internal/build/job/maker.go
+++ b/internal/build/job/maker.go
@@ -1,6 +1,7 @@
 package job
 
 import (
+	"context"
 	"fmt"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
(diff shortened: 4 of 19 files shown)


qbarrand:
Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/115

api/v1beta1/module_types.go
```diff
@@ -60,7 +60,8 @@ type Build struct {
 	// BuildArgs is an array of build variables that are provided to the image building backend.
 	BuildArgs []BuildArg `json:"buildArgs"`
 
-	Dockerfile string `json:"dockerfile"`
+	// ConfigMap that holds Dockerfile contents
+	DockerfileRef *v1.LocalObjectReference `json:"dockerfileRef"`
```
ybettan, line 64: I think this will cause an issue:
Change to the config map won't trigger a reconcile.

qbarrand: @yevgeny-shnaidman can we leave `Dockerfile` as an option? It has the benefit of not breaking the API, which requires a version bump and a conversion webhook...

yevgeny-shnaidman: we don't want to be triggered by ConfigMap. If the image is already build, then we won't run the Build flow again, and if the build has failed, then the reconcile will be triggered by us

controllers/module_reconciler.go
```diff
@@ -95,6 +95,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list
```
qbarrand, line 98: Why do we need to list?
```suggestion
//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get
```

internal/build/job/maker.go
```diff
@@ -105,9 +111,13 @@ func (m *maker) MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.B
...
 	}
 
+	dockefileData, err := m.getDockerfileContents(ctx, buildConfig, mod.Namespace)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get dockerfile contents: %v", err)
+	}
 	specTemplate := v1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
-			Annotations: map[string]string{"Dockerfile": buildConfig.Dockerfile},
+			Annotations: map[string]string{"Dockerfile": dockefileData},
```
qbarrand, line 120: I would rather mount the ConfigMap into the pod.

controllers/module_reconciler.go
yevgeny-shnaidman: just in case - we already have permission for get, maybe later we will use the list

internal/build/job/maker.go
yevgeny-shnaidman: i will check if kaniko supports that option

ybettan: +1

internal/build/job/maker.go
```diff
@@ -148,6 +158,23 @@ func (m *maker) MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.B
 	return job, nil
 }
 
+func (m *maker) getDockerfileContents(ctx context.Context, buildConfig *kmmv1beta1.Build, namespace string) (string, error) {
+	if buildConfig.DockerfileRef == nil {
+		return "", fmt.Errorf("docker config map is mising from Build confgiration")
```
ybettan, line 163:
```suggestion
		return "", errors.New("docker config map is mising from Build confgiration")
```

controllers/module_reconciler.go
qbarrand: Then let us add it later, when we actually need it? 🙂

api/v1beta1/module_types.go
qbarrand: WDYT about having both fields and failing the reconciliation? When we implement the validating webhook, we can add that check.

yevgeny-shnaidman: I don't see the reason to support both options, it is just confusing, both to us and to the customer. Anyway you will have to support both options in the code, which adds to the confusion. Why are you worried about conversion webhook? I think that at this stage we can just tell customers to move to new version and change their code, i think that any customer we have right now is at very starting stage

controllers/module_reconciler.go
yevgeny-shnaidman: returning the list permissions: apparently, even if we use only Get API in the code, we still require the list permission, probably related to the caching

api/v1beta1/module_types.go
ybettan: If we are already breaking the API. Do you have any objection to have [that one](https://github.com/kubernetes-sigs/kernel-module-management/pull/86) as well without a conversion webhook?

yevgeny-shnaidman: i am good with that, @qbarrand ?

enriquebelarte:
FYI. There's a scheduled kmm technical enablement in a few days which relies heavily on [labs](https://github.com/kubernetes-sigs/kernel-module-management/tree/main/docs/labs).  
This change would break all examples in labs.

internal/build/job/maker.go
yevgeny-shnaidman: Fixed. Please review the changes:
1) new Volume definition\
2) new hash calculation
3) changed build config pickup implementation

api/v1beta1/module_types.go
qbarrand: OK to remove the field per @bthurber's decision.
How about changing the new field name to `DockerfileConfigMap`?

internal/build/helper.go
```diff
@@ -45,25 +47,34 @@ func (m *helper) ApplyBuildArgOverrides(args []v1beta1.BuildArg, overrides ...v1
 	return args
 }
 
-func (m *helper) GetRelevantBuild(mod kmmv1beta1.Module, km kmmv1beta1.KernelMapping) *kmmv1beta1.Build {
+func (m *helper) GetRelevantBuild(mod kmmv1beta1.Module, km kmmv1beta1.KernelMapping) (*kmmv1beta1.Build, error) {
 	if mod.Spec.ModuleLoader.Container.Build == nil {
-		// km.Build cannot be nil in case mod.Build is nil, checked above
-		return km.Build.DeepCopy()
+		//km.Build cannot be nil in case mod.Build is nil, checked above
+		if km.Build.DockerfileRef == nil {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: it is a pointer, won't it be set to nil by default...?

internal/build/helper.go
```diff
@@ -45,25 +47,34 @@ func (m *helper) ApplyBuildArgOverrides(args []v1beta1.BuildArg, overrides ...v1
...
+			return nil, errors.New("docker config map is mising from kernel mapping Build confgiration")
+		}
+		return km.Build.DeepCopy(), nil
 	}
 
 	if km.Build == nil {
-		return mod.Spec.ModuleLoader.Container.Build.DeepCopy()
+		if km.Build.DockerfileRef == nil {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed, adding unit test

internal/build/helper.go
```diff
@@ -45,25 +47,34 @@ func (m *helper) ApplyBuildArgOverrides(args []v1beta1.BuildArg, overrides ...v1
...
 	}
 
 	buildConfig := mod.Spec.ModuleLoader.Container.Build.DeepCopy()
-	if km.Build.Dockerfile != "" {
-		buildConfig.Dockerfile = km.Build.Dockerfile
+	if km.Build.DockerfileRef != nil {
+		buildConfig.DockerfileRef = km.Build.DockerfileRef
+	}
+	if buildConfig.DockerfileRef == nil {
+		return nil, errors.New("docker config map is missing from module loader Buid and is also not present in the kernel mapping")
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

api/v1beta1/module_types.go
ybettan: If we are already breaking the API then let's merge this one as well https://github.com/kubernetes-sigs/kernel-module-management/pull/86.

internal/build/job/maker.go
```diff
@@ -148,6 +157,33 @@ func (m *maker) MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.B
 	return job, nil
 }
 
+func (m *maker) getHashAnnotationValue(ctx context.Context, buildConfig *kmmv1beta1.Build, namespace string, podTemplate *v1.PodTemplateSpec) (uint64, error) {
+	dockerfileCM := &corev1.ConfigMap{}
+	namespacedName := types.NamespacedName{Name: buildConfig.DockerfileConfigMap.Name, Namespace: namespace}
+	err := m.client.Get(ctx, namespacedName, dockerfileCM)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get dockerfile configmap %s: %v", buildConfig.DockerfileConfigMap.Name, err)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/build/job/maker.go
```diff
@@ -148,6 +157,33 @@ func (m *maker) MakeJobTemplate(mod kmmv1beta1.Module, buildConfig *kmmv1beta1.B
...
+	}
+	data, ok := dockerfileCM.Data[constants.DockerfileCMKey]
+	if !ok {
+		return 0, fmt.Errorf("invalid Dockerfile ConfigMap %s format, %s key is missing", buildConfig.DockerfileConfigMap.Name, constants.DockerfileCMKey)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/build/helper.go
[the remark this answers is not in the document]
qbarrand: I think Kubernetes won't let you create an object where it's unset.

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
+  newTag: configmap
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/build/helper.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

PR #162, opened 2022-11-24
Adding TLS options to the sign logic.

ybettan:
Adding TLS options to the sign logic.

When checking if the `signed` image already exist or when pulling the
image to sign, we should be able to set some TLS options.

This flow follows the same logic we have for builds.

In additions, this commit also update the type of the
`ModuleLoaderContainerSpec.RegistryTLS` API so we use the default values
when not specified.


ybettan:
This PR is following on https://github.com/kubernetes-sigs/kernel-module-management/pull/86#issuecomment-1326466261

ybettan:
/hold
Adding UT for the new additions.

internal/sign/job/signer.go
```diff
@@ -66,6 +69,24 @@ func (m *signer) MakeJobTemplate(mod kmmv1beta1.Module, signConfig *kmmv1beta1.S
...
 
+	if signConfig.UnsignedImageRegistryTLS.Insecure {
+		args = append(args, "--insecure-pull")
+	}
+
+	if signConfig.UnsignedImageRegistryTLS.InsecureSkipTLSVerify {
+		args = append(args, "--skip-tls-verify-pull")
+	}
+
+	if registryTLS != nil {
```
yevgeny-shnaidman, line 80: this question is more related to a previous PR: 
why not RegistryTLS in kernel mapping needs to be a pointer? Why not make is a struct, and then we don't need to check if the pointer is not nil. In case user does not set it, then  by default its fields are false

ybettan: This is a good point. To be honest I don't know what is the best practice here? I was following what we have in the `build` section.

One of the benefits of using a pointer is that we can just send `nil` in all the unit tests that doesn't care about this part, and then we can test only that part instead of forcing all previous tests to adapt.

When writing tests, I am happy that I can just send `nil`, when looking at the code then we could have indeed remove that `nil` check.

I am not sure which option is better, but I lean toward having `nil` if not specified.

yevgeny-shnaidman: I always prefer to prioritize the production code over tests, especially unit-test.  In the unit-test itself, you can always leave that part out, and it will be initialized by default to false.

ybettan:
/hold
This PR should be rebased on top op https://github.com/kubernetes-sigs/kernel-module-management/pull/163

internal/sign/job/signer.go
ybettan: Fixed. You can see the changes in the last commit. Once https://github.com/kubernetes-sigs/kernel-module-management/pull/163 (which is a blocker for this PR) is merged, I will rebase and squash.

ybettan:
Waiting to see the resolution of https://github.com/kubernetes-sigs/kernel-module-management/issues/164

ybettan:
Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/164

ybettan:
Will be rebased on top https://github.com/kubernetes-sigs/kernel-module-management/pull/156 once merged.

internal/sign/job/signer.go
```diff
@@ -80,6 +89,14 @@ func (m *signer) MakeJobTemplate(
 		args = append(args, "-filestosign", strings.Join(signConfig.FilesToSign, ":"))
 	}
 
+	if signConfig.UnsignedImageRegistryTLS.Insecure {
```
yevgeny-shnaidman, line 92: why is sign TLS logic different from build? they both can push images. Why we don't need to determine registryTLS just as we did in the build?

ybettan: They are not different. Same stuff are done in sign and build.

internal/module/helper.go
```diff
@@ -14,10 +14,10 @@ import (
 )
 
 func TLSOptions(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping) *kmmv1beta1.TLSOptions {
-	if km.RegistryTLS != nil {
-		return km.RegistryTLS
+	if kmRegistry := km.RegistryTLS; kmRegistry != nil {
```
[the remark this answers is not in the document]
ybettan: Fixed.

ybettan:
/hold
Wait for CI

PR #165, opened 2022-11-28
Build signer images in CI

qbarrand:
Use alpine base images for a smaller footprint (46 vs 209MB).
Rework the Dockerfile for better caching.

/cc @chr15p


chr15p:
looks good to me, although I dont think I can add the tag. 
I am wondering if we can reduce the number of layers by reducing all those COPYs down with some wildcards, but as they're in the temporary images it's not super important

PR #166, opened 2022-11-28
Migrate e2e tests

enriquebelarte:
In order to sync CI jobs with m/s and make code reusable
through components, actions scripts tests are migrated to prow scripts.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -76,74 +76,8 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
         if: steps.cache-bin.outputs.cache-hit != 'true'
 
-      - name: Deploy KMMO
-        run: make deploy
-        env:
-          KUSTOMIZE_CONFIG_DEFAULT: ci/install-ci
-
-      - name: Wait until the KMMO Deployment is Available
-        run: kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-        timeout-minutes: 1
-
-      - name: Describe the Deployment and get its YAML if that failed
-        run: |
-          kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-          kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-        if: ${{ failure() }}
-
-      - name: Check that the kmm_ci_a module is not loaded on the node
-        run: |
-          if minikube ssh -- lsmod | grep kmm_ci_a; then
-            echo "Unexpected lsmod output - the module should not be loaded"
-            exit 1
-          fi
-
-      - name: Add an kmm-ci Module that contains a valid mapping
-        run: |
-          sed -e "s/KVER_CHANGEME/$(uname -r)/g" \
-            -e s/NAME_CHANGEME/kmm-ci/ \
-            -e s/KMOD_CHANGEME/kmm_ci_a/ \
-            ci/module-kmm-ci.template.yaml | tee module-kmm-ci.yaml
-
-          kubectl apply -f module-kmm-ci.yaml
-
-      - name: Check that the module gets loaded on the node
-        run: |
-          until minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Check that the node gets labeled with the module's name
-        run: |
-          until kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Check that the daemon-set for device plugin is running
-        run: |
-          until kubectl get ds  -l 'kmm.node.kubernetes.io/module.name=kmm-ci,kmm.node.kubernetes.io/role=device-plugin' -ojson | jq -e '.items[] | select(.status.numberReady == 1)'; do 
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Remove the Module
-        run: kubectl delete -f module-kmm-ci.yaml
-
-      - name: Check that the module gets unloaded from the node
-        run: |
-          until ! minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Check that the node gets unlabeled with the module's name
-        run: |
-          until ! kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do
-            sleep 3
-          done
-        timeout-minutes: 1
+      - name: Run e2e-prebuilt-kernel-module tests
+        run: ./ci/prow/e2e-prebuilt-kernel-module
 
       - name: Collect troubleshooting data
         uses: ./.github/actions/collect-troubleshooting
@@ -195,77 +129,8 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
         if: steps.cache-bin.outputs.cache-hit != 'true'
 
-      - name: Deploy KMMO
-        run: make deploy
-        env:
-          KUSTOMIZE_CONFIG_DEFAULT: ci/install-ci
-
-      - name: Wait until the KMMO Deployment is Available
-        run: kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-        timeout-minutes: 1
-
-      - name: Describe the Deployment / pods and get their YAML if that failed
-        run: |
-          kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-          kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-
-          kubectl describe pod -n kmm-operator-system
-          kubectl get -o yaml pod -n kmm-operator-system
-        if: ${{ failure() }}
-
-      - name: Create one Module for each node
-        run: |
-          # Node minikube gets module a
-          sed -e s/NAME_CHANGEME/kmm-ci-a/g \
-            -e s/KMOD_CHANGEME/kmm_ci_a/g \
-            -e "s/KVER_CHANGEME/${KERNEL_VERSION}/g" \
-            ci/module-kmm-ci.template.yaml | tee module-kmm-ci-a.yaml
-
-          # Node minikube-m02 gets module b
-          sed -e s/NAME_CHANGEME/kmm-ci-b/g \
-            -e s/KMOD_CHANGEME/kmm_ci_b/g \
-            -e "s/KVER_CHANGEME/${KERNEL_VERSION}/g" \
-            ci/module-kmm-ci.template.yaml | tee module-kmm-ci-b.yaml
-
-          kubectl apply -f module-kmm-ci-a.yaml -f module-kmm-ci-b.yaml
-
-      - name: Label the first node to have it receive module a
-        run: kubectl label node minikube wants-module=kmm_ci_a
-
-      - name: Check that module a gets loaded on the first node
-        run: |
-          until minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Label the second node to have it receive module b
-        run: kubectl label node minikube-m02 wants-module=kmm_ci_b
-
-      - name: Check that module b gets loaded on the second node
-        run: |
-          until minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Remove the wants-module label on the second node
-        run: kubectl label node minikube-m02 wants-module-
-
-      - name: Check that module b gets unloaded from the second node
-        run: |
-          until ! minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Verify that the DaemonSet gets garbage collected
-        run: |
-          # Cannot use kubectl wait because it will fail if there is no initial match
-          until [ $(kubectl get daemonsets.apps -l 'kmm.node.kubernetes.io/module.name=kmm-ci-b, kmm.node.kubernetes.io/role!=device-plugin' -o go-template='{{ len .items }}') -eq 0 ]; do
-            sleep 3
-          done
-        timeout-minutes: 1
+      - name: Run e2e-prebuilt-kernel-module-two-nodes tests
+        run: ./ci/prow/e2e-prebuilt-kernel-module-two-nodes
 
       - name: Collect troubleshooting data
         uses: ./.github/actions/collect-troubleshooting
@@ -328,28 +193,7 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
         if: steps.cache-bin.outputs.cache-hit != 'true'
 
-      - name: Deploy KMMO
-        run: make deploy
-        env:
-          KUSTOMIZE_CONFIG_DEFAULT: ci/install-ci
-
-      - name: Wait until the KMMO Deployment is Available
-        run: kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-        timeout-minutes: 1
-
-      - name: Describe the Deployment / pods and get their YAML if that failed
-        run: |
-          kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-          kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-
-          kubectl describe pod -n kmm-operator-system
-          kubectl get -o yaml pod -n kmm-operator-system
-        if: ${{ failure() }}
-
-      - name: Create a build secret
-        run: kubectl create secret generic build-secret --from-literal=ci-build-secret=super-secret-value
-
-      # The minikube registry-alias addon creates a Job that adds registry.minikube to the CoreDNS configuration.
+     # The minikube registry-alias addon creates a Job that adds registry.minikube to the CoreDNS configuration.
       # https://github.com/kubernetes/minikube/blob/master/deploy/addons/registry-aliases/patch-coredns-job.tmpl
       # This job sometimes does not finish before the operator starts looking for the image in the registry, which
       # results in failed DNS resolution.
@@ -361,76 +205,9 @@ jobs:
           kubectl wait --for=condition=Complete --timeout -1s job/wait-minikube-registry-alias
         timeout-minutes: 6
 
-      - name: Check the status of components in the kube-system namespace
-        run: kubectl get all -n kube-system
-        if: ${{ failure() }}
 
-      - name: Add a configmap that contain the kernel module build dockerfile
-        run: kubectl apply -f ci/kmm-kmod-dockerfile.yaml
-
-      - name: Add an kmm-ci Module that contains a valid mapping
-        run: |
-          sed -e "s/KVER_CHANGEME/$(uname -r)/g" ci/module-kmm-ci-build.template.yaml | tee module-kmm-ci.yaml
-
-          kubectl apply -f module-kmm-ci.yaml
-
-      - name: Wait for the job to be created
-        run: |
-          until kubectl get job -l kmm.node.kubernetes.io/module.name | grep kmm; do
-            sleep 1
-          done
-        timeout-minutes: 1
-
-      - name: Check that the module gets loaded on the node
-        run: |
-          until minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Check that the DriverContainer prints the secret's value to the standard output
-        run: |
-          POD_NAME=$(kubectl get pod -l kmm.node.kubernetes.io/module.name --template='{{ (index .items 0).metadata.name }}')
-
-          echo "::group::Looking for the build secret"
-
-          until kubectl exec $POD_NAME -- grep super-secret-value /ci-build-secret; do
-            sleep 3
-          done
-
-          echo "::endgroup::"
-          echo "::group::Looking for the build argument"
-
-          until kubectl exec $POD_NAME -- grep some-build-arg /build-arg; do
-            sleep 3
-          done
-
-          echo "::endgroup::"
-          echo "::group::Looking for the kernel version"
-
-          until kubectl exec $POD_NAME -- grep $KERNEL_VERSION /kernel-version; do
-            sleep 3
-          done
-
-          echo "::endgroup::"
-          echo "::group::Looking for the build argument with a default value"
-
-          until kubectl exec $POD_NAME -- grep default-value /default-value; do
-            sleep 3
-          done
-
-          echo "::endgroup::"
-        timeout-minutes: 1
-
-      - name: Remove the Module
-        run: kubectl delete -f module-kmm-ci.yaml
-
-      - name: Check that the module gets unloaded from the node
-        run: |
-          until ! minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
+      - name: Run e2e-incluster-build tests
+        run: ./ci/prow/e2e-incluster-build
 
       - name: Collect troubleshooting data
         uses: ./.github/actions/collect-troubleshooting
@@ -483,47 +260,8 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
         if: steps.cache-bin.outputs.cache-hit != 'true'
 
-      - name: Deploy KMMO
-        run: make deploy
-        env:
-          KUSTOMIZE_CONFIG_DEFAULT: ci/install-ci
-
-      - name: Wait until the KMMO Deployment is Available
-        run: kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-        timeout-minutes: 1
-
-      - name: Describe the Deployment and get its YAML if that failed
-        run: |
-          kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-          kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
-        if: ${{ failure() }}
-
-      - name: Check that the kmm_ci_a module is not loaded on the node
-        run: |
-          if minikube ssh -- lsmod | grep kmm_ci_a; then
-            echo "Unexpected lsmod output - the module should not be loaded"
-            exit 1
-          fi
-
-      - name: Add an kmm-ci Module that contains a valid mapping
-        run: kubectl apply -f ci/module-kmm-ci-variable.yaml
-
-      - name: Check that the module gets loaded on the node
-        run: |
-          until minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
-
-      - name: Remove the Module
-        run: kubectl delete -f ci/module-kmm-ci-variable.yaml
-
-      - name: Check that the module gets unloaded from the node
-        run: |
-          until ! minikube ssh -- lsmod | grep kmm_ci_a; do
-            sleep 3
-          done
-        timeout-minutes: 1
+      - name: Run e2e-crd-variable tests
+        run: ./ci/prow/e2e-crd-variable
 
       - name: Collect troubleshooting data
         uses: ./.github/actions/collect-troubleshooting
--- a/ci/prow/e2e-crd-variable
+++ b/ci/prow/e2e-crd-variable
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available..."
+timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager
+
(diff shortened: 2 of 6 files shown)


enriquebelarte:
This PR has two uses:

1.  Migrating all the tests, including the newest ones to prow which are not reflected in original [PR](https://github.com/kubernetes-sigs/kernel-module-management/pull/159).
2.  After discussing the need or benefits of including infra as github actions or as shell scripts, reach an agreement to one of both choices. This PR keeps the infra code in github actions but it could be swiftly updated if needed.  
      
    Please cc/ @ybettan @qbarrand

.github/workflows/e2e.yaml
```diff
@@ -64,90 +64,8 @@ jobs:
...
-        id: cache-bin
-        uses: actions/cache@v3
-        with:
-          path: ./bin
-          key: ${{ runner.os }}-bin-${{ env.GO_VERSION }}-${{ hashFiles('Makefile') }}
-
-      - uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
-        if: steps.cache-bin.outputs.cache-hit != 'true'
```
qbarrand, line 77: Shall we keep this?

.github/workflows/e2e.yaml
```diff
@@ -183,93 +101,10 @@ jobs:
...
```
qbarrand, line 196: Shall we keep this?

.github/workflows/e2e.yaml
```diff
@@ -316,125 +151,8 @@ jobs:
...
```
qbarrand, line 329: Shall we keep this?

.github/workflows/e2e.yaml
```diff
@@ -471,60 +189,6 @@ jobs:
...
```
qbarrand, line 484: Shall we keep this?

ci/prow/e2e-crd-variable
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
```
qbarrand, line 3: I think this might prevent us from collecting troubleshooting data if one of the commands fails.

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,70 @@
+##!/usr/bin/env bash
```
qbarrand, line 1:
```suggestion
#!/usr/bin/env bash
```

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,70 @@
...
+
+echo "Create a build secret..."
+kubectl create secret generic build-secret --from-literal=ci-build-secret=super-secret-value
+
+      # The minikube registry-alias addon creates a Job that adds registry.minikube to the CoreDNS configuration.
+      # https://github.com/kubernetes/minikube/blob/master/deploy/addons/registry-aliases/patch-coredns-job.tmpl
+      # This job sometimes does not finish before the operator starts looking for the image in the registry, which
+      # results in failed DNS resolution.
+      # Add a job that tries to resolve registry.minikube, like the operator.
+      # Wait up to 6 minutes, which corresponds to our job's maximum lifetime including backoffs.
```
qbarrand, line 20:
```suggestion
# The minikube registry-alias addon creates a Job that adds registry.minikube to the CoreDNS configuration.
# https://github.com/kubernetes/minikube/blob/master/deploy/addons/registry-aliases/patch-coredns-job.tmpl
# This job sometimes does not finish before the operator starts looking for the image in the registry, which
# results in failed DNS resolution.
# Add a job that tries to resolve registry.minikube, like the operator.
# Wait up to 6 minutes, which corresponds to our job's maximum lifetime including backoffs.
```

qbarrand: Thinking more about this, I think it rather belongs to the infra setup. We won't use the minikube internal registry midstream.

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,70 @@
...
+          echo "::group::Looking for the kernel version"
+          until timeout 1m kubectl exec $POD_NAME -- grep $KERNEL_VERSION /kernel-version; do
+            sleep 3
+          done
+          echo "::endgroup::"
+          echo "::group::Looking for the build argument with a default value"
+          until timeout 1m kubectl exec $POD_NAME -- grep default-value /default-value; do
+            sleep 3
+          done
+          echo "::endgroup::"
```
qbarrand, line 61: Not sure the grouping is relevant in those scripts anymore, if they are to be used downstream.

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,54 @@
+##!/usr/bin/env bash
```
qbarrand, line 1:
```suggestion
#!/usr/bin/env bash
```

ci/prow/e2e-prebuilt-kernel-module-two-nodes
```diff
@@ -0,0 +1,71 @@
+##!/usr/bin/env bash
```
qbarrand, line 1:
```suggestion
#!/usr/bin/env bash
```

ybettan:
I am not sure I understand the purpose of this PR:
1. Is this just another implementation of https://github.com/kubernetes-sigs/kernel-module-management/pull/159?
2. Why are we building the driver container in each job instead of just once?

ci/prow/e2e-crd-variable
```diff
@@ -0,0 +1,31 @@
...
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available and "
+echo "Describe the Deployment and get its YAML if that failed..."
+timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager || kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager && kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
```
ybettan, line 10: How about having something like that:
```
#!/usr/bin/env bash
on_error() { 
   echo "Some error occurred"
}

trap 'on_error' ERR

ls ~/dir_not_exists
```
See ref: https://tecadmin.net/execute-specific-command-on-error-in-bash-script/

ci/prow/e2e-crd-variable
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available and "
+echo "Describe the Deployment and get its YAML if that failed..."
```
ybettan, line 9: I think we can remove that comment.

ci/prow/e2e-crd-variable
```diff
@@ -0,0 +1,31 @@
...
+timeout 1m bash -c 'until minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Remove the Module..."
+kubectl delete -f ci/module-kmm-ci-variable.yaml
+
+echo "Check that the module gets unloaded from the node..."
+timeout 1m bash -c 'until ! minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Collect troubleshooting data..."
+./ci/github/collect_troubleshooting
```
ybettan, line 31: Won't it collect only on success?

ybettan: **Also
(can't update the original comment for some reason)

qbarrand: We collect on success today.

.github/workflows/e2e.yaml
ybettan: If we keep it should be extracted from the tests and move to the "infra" part that should run only once per PR (and not for each job).

ci/prow/e2e-crd-variable
enriquebelarte: IMHO it is simpler to use an operand in this case. If it succeeds stop and if it does not succeed then show whatever.

ci/prow/e2e-crd-variable
ybettan: Any reason to not do it only on failure if we are changing the structure of the tests?

ci/prow/e2e-incluster-build
ybettan: We won't use minikube at all m/s :) 
+1 on moving to infra.

ci/prow/e2e-crd-variable
enriquebelarte: Now in actions it is configured as [always()](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/.github/workflows/e2e.yaml#L530), so I just tried to keep it similar.

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,70 @@
...
+
+echo "Add a configmap that contain the kernel module build dockerfile..."
+kubectl apply -f ci/kmm-kmod-dockerfile.yaml
+
+echo "Add an kmm-ci Module that contains a valid mapping..."
+sed -e "s/KVER_CHANGEME/$(uname -r)/g" ci/module-kmm-ci-build.template.yaml | tee module-kmm-ci.yaml
+kubectl apply -f module-kmm-ci.yaml
+
+echo "Wait for the job to be created..."
+timeout 1m bash -c 'until kubectl get job -l kmm.node.kubernetes.io/module.name | grep kmm; do sleep 1; done'
```
ybettan, line 34:
```suggestion
echo "Add an kmm-ci Module that contains a valid mapping..."
sed -e "s/KVER_CHANGEME/$(uname -r)/g" ci/module-kmm-ci-build.template.yaml | tee module-kmm-ci.yaml
kubectl apply -f module-kmm-ci.yaml
timeout 1m bash -c 'until kubectl get job -l kmm.node.kubernetes.io/module.name | grep kmm; do sleep 1; done'
```

.github/workflows/e2e.yaml
qbarrand: Every job runs in a different VM, so we cannot really run this once. Maybe both steps could be combined into a single reusable action, but that action would need to be called in each job that needs the binaries (all of them as of today).

ci/prow/e2e-incluster-build
enriquebelarte: It already [exists](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/.github/workflows/e2e.yaml#L377) in current tests and it looks relevant info in my opinion. Why removing that comment?

ci/prow/e2e-crd-variable
enriquebelarte: Why? It is pretty informative about what next lines will do.

ci/prow/e2e-crd-variable
ybettan: But then you need to repeat that for every command that may fail. Isn't it?

ci/prow/e2e-crd-variable
ybettan: I think that the comments should describe the happy flow, in case of an error we just print some debugging info but there is no reason to put it there to begin with.

"Too much documentation" isn't always good IMO.

But if you insist, we can keep it.

ci/prow/e2e-crd-variable
ybettan: I think we should update it. No one will look at it in case of success, this is just more code that can fail, more CPU cycles, slower CI etc.

This is why I have suggested https://github.com/kubernetes-sigs/kernel-module-management/pull/166#discussion_r1033541627 so we don't have to have add the following to each command:
```
<some command> || <handle error>
```

ybettan:
> We are building it just once and downloading it in all jobs, just like before.

Umm I see `./.github/actions/build-drivercontainer-image` in every job. Am I missing something?
@qbarrand

ci/prow/e2e-incluster-build
ybettan: Same reason as before, this is my option only and others may think other wise but eventually we are "applying a module", all the rest is just implementation details - so we are adding redundant info that shadows the main step which is just applying a module.

If you don't agree with this we can keep it.

ci/prow/e2e-crd-variable
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+set -x
```
ybettan, line 3:
```suggestion
set -xe
```

ci/prow/e2e-crd-variable
enriquebelarte: This makes sense to me. WDYT @qbarrand ?

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+set -x
```
ybettan, line 3:
```suggestion
set -xe
```

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,54 @@
...
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available"
+echo "And Describe the Deployment / pods and get their YAML if that failed..."
+timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager || kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager && kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager && kubectl describe pod -n kmm-operator-system && kubectl get -o yaml pod -n kmm-operator-system
```
ybettan, line 10:
```suggestion
echo "Deploy KMMO..."
make deploy
timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager
```

ybettan: By using https://github.com/kubernetes-sigs/kernel-module-management/pull/166#discussion_r1033541627, those error handling looks much cleaner and it is easier to understand what this step is supposed to do.

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,54 @@
...
+            sleep 3
+          done
+          echo "Looking for the kernel version"
+          until timeout 1m kubectl exec $POD_NAME -- grep $KERNEL_VERSION /kernel-version; do
+            sleep 3
+          done
+          echo "Looking for the build argument with a default value"
+          until timeout 1m kubectl exec $POD_NAME -- grep default-value /default-value; do
+            sleep 3
+          done
```
ybettan, line 45: Can we fix the indention?

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,54 @@
...
+timeout 1m bash -c 'until kubectl get job -l kmm.node.kubernetes.io/module.name | grep kmm; do sleep 1; done'
+
+echo "Check that the module gets loaded on the node..."
+timeout 1m bash -c 'until minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Check that the DriverContainer prints the secret's value to the standard output..."
+POD_NAME=$(kubectl get pod -l kmm.node.kubernetes.io/module.name --template='{{ (index .items 0).metadata.name }}')
+KERNEL_VERSION=$(uname -r)
+          echo "Looking for the build secret"
+          until timeout 1m kubectl exec $POD_NAME -- grep super-secret-value /ci-build-secret; do
```
ybettan, line 31: Isn't the `timeout` supposed to wrap the  `until`?

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,54 @@
...
+          done
+
+echo "Remove the Module..."
+kubectl delete -f module-kmm-ci.yaml
+
+echo "Check that the module gets unloaded from the node..."
+timeout 1m bash -c 'until ! minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Collect troubleshooting data ..."
+./ci/github/collect_troubleshooting
```
ybettan, line 54: Can we collect only on errors? We don't need it on success.

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
```
ybettan, line 3: Should we have the same config for all the jobs?

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,54 @@
...
+
+echo "Label the node so that it gets the module..."
+kubectl label node minikube wants-module=kmm_ci_a
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available and show info if it fails..."
+timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m || kubectl describe deployments.apps -n kmm-operator-system kmm-operator-controller-manager && kubectl get -o yaml deployments.apps -n kmm-operator-system kmm-operator-controller-manager
```
ybettan, line 12: Same comment as before.

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,54 @@
...
+          until ! timeout 1m minikube ssh -- lsmod | grep kmm_ci_a; do
+            sleep 3
+          done
+
+echo "Check that the node gets unlabeled with the module's name..."
+          until ! timeout 1m kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do
+            sleep 3
+          done
+
+
```
ybettan, line 54: Is it just the Github UI or something is wrong with the indention here?

ci/prow/e2e-prebuilt-kernel-module-two-nodes
```diff
@@ -0,0 +1,71 @@
...
+echo "Check that module b gets unloaded from the second node..."
+          until ! minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do
+            sleep 3
+          done 
+
+echo "Verify that the DaemonSet gets garbage collected..."
+          # Cannot use kubectl wait because it will fail if there is no initial match
+          until [ $(kubectl get daemonsets.apps -l 'kmm.node.kubernetes.io/module.name=kmm-ci-b, kmm.node.kubernetes.io/role!=device-plugin' -o go-template='{{ len .items }}') -eq 0 ]; do
+            sleep 3
+          done
```
ybettan, line 69: Intentions?

ci/prow/e2e-crd-variable
enriquebelarte: I think that it we use -e here it will exit before getting to the troubleshoot if any of the previous commands fail, as @qbarrand commented [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/166#discussion_r1033489465)

ci/prow/e2e-incluster-build
enriquebelarte: Agree. Will fix.

ci/prow/e2e-crd-variable
qbarrand: This currently takes one second: https://github.com/kubernetes-sigs/kernel-module-management/actions/runs/3573079866/jobs/6006759230
I would keep collecting it because it has virtually no cost and can be useful to understand issues where the script exits normally but still something isn't working as expected.
I would move it back to `e2e.yaml` though, because CI jobs calling that script might use an alternative collection method (e.g. must-gather on OCP).

ybettan: > can be useful to understand issues where the script exits normally but still something isn't working as expected

Do you have any example for that? We should not encounter such when using `set -e`.

ybettan: Yup. It should be part of the infra anyway. 

I think we still need to print only on failures. I think there is an easy option to set it in the actions if I am not mistaken.

qbarrand: How about this?

> I would keep collecting it because it has virtually no cost and can be useful to understand issues where the script exits normally but still something isn't working as expected.

Let us not change anything from the current logic unless it actually causes issues.

ci/prow/e2e-crd-variable
qbarrand: If we move the troubleshooting back to the Action YAML, `set -euxo pipefail` makes sense IMHO.

ci/prow/e2e-crd-variable
qbarrand: We could put the `kubectl describe` in the troubleshooting step. If we collect that output at the end, we can remove the operands and the comment.

ci/prow/e2e-crd-variable
qbarrand: Let's pull all the troubleshooting in one step called by the Action YAML?

ci/prow/e2e-incluster-build
qbarrand: I would leave it. When reading the log file, the comment is easier to understand that the command.

qbarrand:
> Umm I see ./.github/actions/build-drivercontainer-image in every job. Am I missing something?

I stand here corrected, my apologies. We used to build both only once, but now it's only the operator image.
The reason is that not all GitHub Actions VMs run the same kernel; this caused issues because the kmod was built in a VM running kernel X, while another VM trying to load that kmod was running kernel Y.
Original fix PR: https://github.com/qbarrand/oot-operator/pull/72

ci/prow/e2e-crd-variable
enriquebelarte: It looks like a good solution. WDYT @ybettan ?

.github/workflows/e2e.yaml
```diff
@@ -76,78 +76,12 @@ jobs:
...
-            sleep 3
-          done
-        timeout-minutes: 1
+      - name: Run e2e-prebuilt-kernel-module tests
+        run: ./ci/prow/e2e-prebuilt-kernel-module
 
       - name: Collect troubleshooting data
         uses: ./.github/actions/collect-troubleshooting
-        if: ${{ always() }}
+        if: ${{ always() }}
```
qbarrand, line 84:
```suggestion
        if: ${{ always() }}
```

ci/prow/e2e-incluster-build
```diff
@@ -0,0 +1,43 @@
...
+POD_NAME=$(kubectl get pod -l kmm.node.kubernetes.io/module.name --template='{{ (index .items 0).metadata.name }}')
+KERNEL_VERSION=$(uname -r)
+echo "Looking for the build secret"
+timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep super-secret-value /ci-build-secret; do sleep 3; done'
+echo "Looking for the build argument"
+timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep some-build-arg /build-arg; do sleep 3; done'
+echo "Looking for the kernel version"
+timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep $KERNEL_VERSION /kernel-version; do sleep 3; done'
+echo "Looking for the build argument with a default value"
+timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep default-value /default-value; do sleep 3; done'
```
qbarrand, line 36:
```suggestion
echo "Check that the DriverContainer prints the secret's value to the standard output..."
POD_NAME=$(kubectl get pod -l kmm.node.kubernetes.io/module.name --template='{{ (index .items 0).metadata.name }}')
KERNEL_VERSION=$(uname -r)

echo "Looking for the build secret"
timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep super-secret-value /ci-build-secret; do sleep 3; done'

echo "Looking for the build argument"
timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep some-build-arg /build-arg; do sleep 3; done'

echo "Looking for the kernel version"
timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep $KERNEL_VERSION /kernel-version; do sleep 3; done'

echo "Looking for the build argument with a default value"
timeout 1m bash -c 'until kubectl exec $POD_NAME -- grep default-value /default-value; do sleep 3; done'
```

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,49 @@
...
+    -e s/NAME_CHANGEME/kmm-ci/ \
+    -e s/KMOD_CHANGEME/kmm_ci_a/ \
+    ci/module-kmm-ci.template.yaml | tee module-kmm-ci.yaml
+kubectl apply -f module-kmm-ci.yaml
+
+echo "Check that the kernel module gets loaded on the node..."
+timeout 1m bash -c 'until minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Check that the node gets labeled with the module's name..."
+kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do
```
qbarrand: Missing `until` here? Shouldn't there be a timeout as well?

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,47 @@
...
+    -e s/NAME_CHANGEME/kmm-ci/ \
+    -e s/KMOD_CHANGEME/kmm_ci_a/ \
+    ci/module-kmm-ci.template.yaml | tee module-kmm-ci.yaml
+kubectl apply -f module-kmm-ci.yaml
+
+echo "Check that the kernel module gets loaded on the node..."
+timeout 1m bash -c 'until minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Check that the node gets labeled with the module's name..."
+timeout 1m bash -c 'kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do sleep 3; done'
```
qbarrand, line 31:
```suggestion
timeout 1m bash -c 'until kubectl get node minikube -o jsonpath='{.metadata.labels}' | jq -e 'has("kmm.node.kubernetes.io/kmm-ci.ready")'; do sleep 3; done'
```

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,47 @@
...
+
+echo "Check that the daemon-set for device plugin is running..."
+until kubectl get ds  -l 'kmm.node.kubernetes.io/module.name=kmm-ci,kmm.node.kubernetes.io/role=device-plugin' -ojson | jq -e '.items[] | select(.status.numberReady == 1)'; do
```
qbarrand, line 34: Don't we need a timeout here?

ci/prow/e2e-prebuilt-kernel-module-two-nodes
```diff
@@ -0,0 +1,60 @@
...
+kubectl label node minikube-m02 wants-module=kmm_ci_b
+
+echo "Check that module b gets loaded on the second node..."
+timeout 1m bash -c 'until minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do sleep 3; done'
+
+echo "Remove the wants-module label on the second node..."
+kubectl label node minikube-m02 wants-module-
+
+echo "Check that module b gets unloaded from the second node..."
+until ! minikube ssh -n minikube-m02 -- lsmod | grep kmm_ci_b; do
```
qbarrand, line 51: Don't we need a timeout here?

ci/prow/e2e-prebuilt-kernel-module-two-nodes
```diff
@@ -0,0 +1,60 @@
...
+  sleep 3
+done 
+
+echo "Verify that the DaemonSet gets garbage collected..."
+# Cannot use kubectl wait because it will fail if there is no initial match
+until [ $(kubectl get daemonsets.apps -l 'kmm.node.kubernetes.io/module.name=kmm-ci-b, kmm.node.kubernetes.io/role!=device-plugin' -o go-template='{{ len .items }}') -eq 0 ]; do
```
qbarrand, line 57: Don't we need a timeout here?

ci/prow/e2e-crd-variable
ybettan: How will it work m/s? will you collect it twice? once in the job and another time in the must-gather artifact?

ybettan:
> @ybettan I stand here corrected, my apologies. We used to build both only once, but now it's only the operator image.
The reason is that not all GitHub Actions VMs run the same kernel; this caused issues because the kmod was built in a VM running kernel X, while another VM trying to load that kmod was running kernel Y.
Original fix PR: https://github.com/qbarrand/oot-operator/pull/72

I see. In that case I would add a comment to explain it in the test, otherwise, anyone looking at it may just change it without understanding the problem.

ci/prow/e2e-prebuilt-kernel-module
```diff
@@ -0,0 +1,49 @@
...
+set -euxo pipefail
+
+echo "Label the node so that it gets the module..."
+kubectl label node minikube wants-module=kmm_ci_a
+
+echo "Deploy KMMO..."
+make deploy
+
+echo "Wait until the KMMO Deployment is Available and show info if it fails..."
+timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager --request-timeout=1m
```
ybettan, line 12: Why are we using both `timeout 1m` and `--request-timeout=1m`? I think we should use the same mechanism everywhere.

ci/prow/e2e-crd-variable
qbarrand: No, because @enriquebelarte has moved the collection to an Action step. We can reuse the script m/s because it does not do any collection whatsoever.

ybettan:
What happened to all the files in the PR? Seems like there were all deleted...

enriquebelarte:
> What happened to all the files in the PR? Seems like there were all deleted...

Bad commit. Fixed yet.

PR #167, opened 2022-11-29
Update Labs examples with ConfigMaps

enriquebelarte:
Example YAML files have been updated in order to match the new way
of setting the build of modules using a ConfigMap as a build parameter in
the declaration of Module objects.


docs/labs/gke/build-configmap.yaml
```diff
@@ -0,0 +1,27 @@
+dockerfile: |
```
ybettan, line 1: Aren't we missing
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: build-module
data:
```
here?

enriquebelarte: Absolutely. Fixed, thank you!

docs/labs/azure/kmm-kmod.yaml
```diff
@@ -9,39 +9,11 @@ spec:
         moduleName: kmm-ci-a 
 
       kernelMappings:
-        - literal: 5.4.0-1089-azure 
+        - regexp: '^.+.azure'
```
qbarrand, line 12:
```suggestion
        - regexp: '^.+-azure$'
```

docs/labs/aws/kmm-kmod.yaml
```diff
@@ -9,43 +9,14 @@ spec:
         moduleName: kmm-ci-a 
 
       kernelMappings:
-        - literal: 5.4.196-108.356.amzn2.x86_64
-          containerImage: <aws_account_number>.dkr.ecr.<aws_region>.amazonaws.com/kmmo-lab:kmm-2.2
+        - regexp: '^.+.amzn2'
```
qbarrand, line 12:
```suggestion
        - regexp: '^.+-amzn2$' 
```

docs/labs/gke/gke-ubuntu-nodes.yaml
```diff
@@ -9,38 +9,15 @@ spec:
         moduleName: kmm-ci-a 
 
       kernelMappings:
-        - literal: 5.4.0-1078-gke 
+        - regexp: '^.+.gke'
```
qbarrand, line 12:
```suggestion
        - regexp: '^.+-gke$'
```

docs/labs/ubuntu/kmm-kmod.yaml
```diff
@@ -9,37 +9,11 @@ spec:
         moduleName: kmm-ci-a 
 
       kernelMappings:
-        - literal: 5.15.0-43-generic
+        - regexp: '^.+.generic'
```
qbarrand, line 12:
```suggestion
        - regexp: '^.+-generic$'
```

docs/labs/aws/kmm-kmod.yaml
enriquebelarte: This won't be modified as regexp would not match with kernel version string.

PR #168, opened 2022-11-29
Fix check-api-changes script

enriquebelarte:
Change the use of make generate manifests for make generate manifests bundle so
 we make sure we push both targets in the PRs.


qbarrand:
test-infra PR to add a job that runs this script: https://github.com/kubernetes/test-infra/pull/28130

PR #174, opened 2022-11-30
Pushing all missing generated files.

ybettan:
Those files were supposed to be added to previous PRs but were forgotten.

We now have a CI job to make sure we are not forgetting to push all generated files with each change.

This commit is for making sure we are aligned.

### DIFF
--- a/config/manifests/bases/kernel-module-management.clusterserviceversion.yaml
+++ b/config/manifests/bases/kernel-module-management.clusterserviceversion.yaml
@@ -15,6 +15,12 @@ spec:
       kind: Module
       name: modules.kmm.sigs.k8s.io
       version: v1alpha1
+    - description: ManagedClusterModule is the Schema for the managedclustermodules
+        API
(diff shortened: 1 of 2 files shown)


config/manifests/bases/kernel-module-management.clusterserviceversion.yaml
```diff
@@ -15,6 +15,12 @@ spec:
       kind: Module
       name: modules.kmm.sigs.k8s.io
       version: v1alpha1
+    - description: ManagedClusterModule is the Schema for the managedclustermodules
```
mresvanis, line 18: This was intentionally not pushed to the repository, because we may have a separate image for this CRD and its reconciler. But I messed things up when I pushed  [this](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/config/crd/kustomization.yaml#L7).

We don't have to do anything about this at this point, we will make sure to remove if we indeed create a separate image for that CRD and reconciler.

PR #178, opened 2022-12-02
Add build and sign in Hub for ManagedClusterModule reconciler

mresvanis:
This PR implements KMM's `Build and Sign` feature for Hub & Spoke environments. Builds and signings take place only on the Hub cluster. 

The changes included are: 
- add a new  `ManagedClusterModule.Spec.JobNamespace` field. Since the `ManagedClusterModule` CRD is cluster-scoped, this field indicates the namespace where all `{build,sign}` k8s `Job`s will be deployed to. 
- rename the `ManagedClusterModule.Spec.Namespace` to the more explicit `MangedClusterModule.Spec.SpokeNamespace`, in order to make the API clearer since we're introducing the `ManagedClusterModule.Spec.JobNamespace`
- add an  internal `clusterAPI` package, which implements the functions needed for the `ManagedClusterModule` reconciler to be able to `build and sign`, as well as to deploy `Module`s to Spoke clusters.
- add `build and sign` as a `ManagedClusterModule` reconcile step.
- add garbage collection for the `{build,sign}` k8s `Job`s, which are owned by the `ManagedClusterModule` CR, at the end of each reconcile loop.
 

internal/cluster/cluster.go
```diff
@@ -0,0 +1,267 @@
...
+		client:    client,
+		kernelAPI: kernelAPI,
+		buildAPI:  buildAPI,
+		signAPI:   signAPI,
+	}
+}
+
+type clusterAPI struct {
+	client client.Client
+
```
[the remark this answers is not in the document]
qbarrand: This one is not very clear - it's about removing the new line.

internal/cluster/cluster.go
```diff
@@ -0,0 +1,267 @@
...
+	clusters clusterv1.ManagedClusterList,
+	mcm kmmv1beta1.ManagedClusterModule) ([]string, error) {
+
+	var err error
+	collected := make([]string, 0)
+
+	for _, cluster := range clusters.Items {
+		deleted, clusterErr := c.buildAPI.GarbageCollect(ctx, mcm.Name, cluster.Name)
+		if clusterErr != nil {
+			err = fmt.Errorf("failed to collect logs for cluster %s: %v", cluster.Name, clusterErr)
```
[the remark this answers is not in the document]
mresvanis: That's true. This change is going to be refactored by my next change. I'll keep you posted.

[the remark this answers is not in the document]
mresvanis: This change was reverted, as now we have one namespace where all `{build,sign}` Jobs are deployed to.

ybettan:
> Since the ManagedClusterModule CRD is cluster-scoped

What is the reason for it being cluster scoped?

api/v1beta1/managedclustermodule_types.go
```diff
@@ -28,6 +28,9 @@ type ManagedClusterModuleSpec struct {
 	// Namespace describes the Spoke namespace, in which the ModuleSpec should be applied.
 	Namespace string `json:"namespace,omitempty"`
 
+	// JobNamespace describes the Hub namespace, to which Build and Sign Jobs should be deployed.
+	JobNamespace string `json:"jobNamespace,omitempty"`
```
ybettan, line 32: Now that we have this, should we change the `Namespace` field to `SpokeNamespace`? Your comment is excellent but I always prefer self documented code :)

controllers/managedclustermodule_reconciler.go
```diff
@@ -82,25 +87,35 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, req ctrl
...
 	}
 
 	for _, cluster := range clusters.Items {
+		logger := log.FromContext(ctx).WithValues("cluster", cluster.Name)
+		clusterCtx := log.IntoContext(ctx, logger)
+
+		if requeue := r.clusterAPI.BuildAndSign(clusterCtx, *mcm, cluster); requeue {
+			logger.Info("Build and Sign require a requeue; skipping ManifestWork reconciliation")
+			res.Requeue = true
+			continue
```
ybettan, line 102: Not sure about the inner implementation but does it mean that we are not going to update the status of other clusters as long as a prior one has been requed?

controllers/managedclustermodule_reconciler.go
```diff
@@ -82,25 +87,35 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, req ctrl
...
 
-		opRes, err := controllerutil.CreateOrPatch(ctx, r.client, mw, func() error {
+		opRes, err := controllerutil.CreateOrPatch(clusterCtx, r.client, mw, func() error {
 			return r.manifestAPI.SetManifestWorkAsDesired(ctx, mw, *mcm)
 		})
 
 		if err != nil {
-			return res, fmt.Errorf("failed to create/patch ManifestWork for managed cluster %s: %v", cluster.Name, err)
+			logger.Error(err, "failed to create/patch ManifestWork for managed cluster: %v", err)
+			continue
```
ybettan, line 118: Why do we skip other clusters here? Don't we want a best effort still trying to create a `ManifestWork` for other clusters?

mresvanis:
> > Since the ManagedClusterModule CRD is cluster-scoped
> 
> What is the reason for it being cluster scoped?

 The deployment of KMM `Module`s to the managed clusters is done via OCM's [ManifestWork](https://open-cluster-management.io/concepts/manifestwork/) Custom Resources. A `ManagedClusterModule` CR is responsible for multiple `ManifestWork` CRs, each deployed to the namespace of the managed cluster it's targeting. Thus the former ends up needing to own resources in multiple namespaces, which can only be done via a [cluster-scoped CRD](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).
 
 > Cross-namespace owner references are disallowed by design. Namespaced dependents can specify cluster-scoped or namespaced owners.
 
 Does this make sense?

controllers/managedclustermodule_reconciler.go
mresvanis: Yes, exactly. We don't skip other clusters, we are logging the error and then `continue`-ing the loop in order to try to reconcile the rest of the managed clusters, which the `ManagedClusterModule` selector is targeting.

controllers/managedclustermodule_reconciler.go
mresvanis: The goal of enabling the `res.Requeue` and then `continue`-ing through the loop is to indicate that this managed cluster cannot move forward with the deployment of the KMM `Module` via the creation/patching of the respective OCM `ManifestWork` CR. One or more `{build,sign}` Jobs for this `Module` might be still running. So we just indicate that we need to have the specified `ManagedClusterModule` CR go into a additional reconcile loop, even if only one of the managed clusters needs to wait for a `{build/sign}` Job to complete.

WDYT?

internal/cluster/cluster.go
```diff
@@ -0,0 +1,264 @@
...
+
+	logger := log.FromContext(ctx).WithValues(
+		"kernel version", kernelVersion,
+		"image", kernelMapping.ContainerImage)
+	buildCtx := log.IntoContext(ctx, logger)
+
+	buildRes, err := c.buildAPI.Sync(buildCtx, mod, *kernelMapping, kernelVersion, true, mcm)
+	if err != nil {
+		return false, fmt.Errorf("could not synchronize the build: %w", err)
+	}
```
ybettan, line 226: This logic is also used in `controllers/module_reconciler.go`. Can we extract that functionality to a separate API (utils for example) and reuse it instead of duplicating it?

internal/cluster/cluster.go
```diff
@@ -0,0 +1,264 @@
...
+
+	logger := log.FromContext(ctx).WithValues(
+		"kernel version", kernelVersion,
+		"image", kernelMapping.ContainerImage)
+	signCtx := log.IntoContext(ctx, logger)
+
+	signRes, err := c.signAPI.Sync(signCtx, mod, *kernelMapping, kernelVersion, previousImage, true, mcm)
+	if err != nil {
+		return false, fmt.Errorf("could not synchronize the signing: %w", err)
+	}
```
ybettan, line 261: This logic is also used in `controllers/module_reconciler.go`. Can we extract that functionality to a separate API (utils for example) and reuse it instead of duplicating it?

controllers/managedclustermodule_reconciler.go
ybettan: Yup. Sound good to me. Thanks for the explanation.

controllers/managedclustermodule_reconciler.go
ybettan: Seems like I forgot what `continue` is doing ;) 
I confused it with `break` but for loops haha.

Thanks. It makes perfect sense now.

ybettan:
> The deployment of KMM Modules to the managed clusters is done via OCM's [ManifestWork](https://open-cluster-management.io/concepts/manifestwork/) Custom Resources. A ManagedClusterModule CR is responsible for multiple ManifestWork CRs, each deployed to the namespace of the managed cluster it's targeting. Thus the former ends up needing to own resources in multiple namespaces, which can only be done via a [cluster-scoped CRD](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).

So if I want some work in a spoke cluster in namesapce `a` I have to create a `ManifestWork` in the hub in namespace `a`? (this is what I understand from the doc) I would expect a field in the `ManifestWork` instead but if this is how `ManifestWork` work then I guess this is the only option.

mresvanis:
> > The deployment of KMM Modules to the managed clusters is done via OCM's [ManifestWork](https://open-cluster-management.io/concepts/manifestwork/) Custom Resources. A ManagedClusterModule CR is responsible for multiple ManifestWork CRs, each deployed to the namespace of the managed cluster it's targeting. Thus the former ends up needing to own resources in multiple namespaces, which can only be done via a [cluster-scoped CRD](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/).
> 
> So if I want some work in a spoke cluster in namesapce `a` I have to create a `ManifestWork` in the hub in namespace `a`? (this is what I understand from the doc) I would expect a field in the `ManifestWork` instead but if this is how `ManifestWork` work then I guess this is the only option.

Yes, exactly. On the Hub cluster there is one namespace per managed cluster. So the `ManifestWork` deployment is dictated by the current OCM [design](https://open-cluster-management.io/concepts/manifestwork/):

> In the open-cluster-management project, a ManifestWork resource must be created in the cluster namespace. 

Example `ManifestWork`:

```yaml
apiVersion: work.open-cluster-management.io/v1
kind: ManifestWork
metadata:
  namespace: <target managed cluster>
  name: hello-work-demo
spec:
  ...
```

internal/cluster/cluster.go
mresvanis: Good point and I was also thinking about this when implementing this part. I was trying to find a way to refactor the `buildAPI` itself to accommodate for this combined functionality, i.e. a call that checks whether a sync should run or not and runs the actual `sync`. I realized that it would be better to skip adding a layer on top of the `buildAPI` in another internal package.

So while I also thought that this is duplicating code, I figured I'd leave as is, because I couldn't find a better trade-off. 

WDYT? Should we go about adding a layer or refactor the current `buildAPI` (or maybe a 3rd option that I'm not thinking about)? 

PS. if we decide to refactor the current `buildAPI`, I think that we could do that in a separate PR. So that we do it with more focus and to not make this one bigger.

internal/cluster/cluster.go
mresvanis: Same as [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/178#discussion_r1042063545). Let's keep the conversation there, everything that applies there, applies here as well by default :)

internal/cluster/cluster.go
mresvanis: @ybettan @yevgeny-shnaidman @qbarrand maybe there are more ideas, as to how to tackle this?

yevgeny-shnaidman: We will probably schedule a refactoring of the code after GA, we have a few places where we can improve our code, so lets get back to that issue then

mresvanis: Duly noted, I agree with @yevgeny-shnaidman and I believe we can afford to come back to this piece of code after GA.

yevgeny-shnaidman:
@mresvanis kets just squash the commits and then unhold the PR

PR #179, opened 2022-12-02
Fix module reconciler not being able to watch for ConfigMaps error

mresvanis:
This fix adds the `watch` permission for `ConfigMap`s to the `Module` reconciler, in order to be able to successfully `client.Get()` the build Dockerfile `ConfigMap`. This `client.Get()` fails because the contoller-runtime client reads from its cache, instead of directly from the API server. The client needs the `watch` permission, in order to be able to build its cache.

- https://github.com/kubernetes-sigs/controller-runtime/issues/1156

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -54,6 +54,7 @@ rules:
   verbs:
   - get
   - list
+  - watch
 - apiGroups:
   - ""
   resources:
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -90,7 +90,7 @@ func NewModuleReconciler(
 //+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
-//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list;watch
 //+kubebuilder:rbac:groups="core",resources=serviceaccounts,verbs=create;delete;get;list;patch;watch
 //+kubebuilder:rbac:groups="batch",resources=jobs,verbs=create;list;watch;delete
 


yevgeny-shnaidman:
@mresvanis we have an e2e test that is runs the build flow with configmap, and it has been passing successfully since the ConfigMap has been added. What has changed that we need the watch permission ( get and list previously was enough)?

yevgeny-shnaidman:
> @mresvanis we have an e2e test that is runs the build flow with configmap, and it has been passing successfully since the ConfigMap has been added. What has changed that we need the watch permission ( get and list previously was enough)?
I am going to lgtm , but we need to better understand what's going on there: we are not watching ConfigMaps in any of our controllers, only Getting it, but not having permissions to watch interefers with our watch for Builds..

mresvanis:
> > @mresvanis we have an e2e test that is runs the build flow with configmap, and it has been passing successfully since the ConfigMap has been added. What has changed that we need the watch permission ( get and list previously was enough)?
> > I am going to lgtm , but we need to better understand what's going on there: we are not watching ConfigMaps in any of our controllers, only Getting it, but not having permissions to watch interefers with our watch for Builds..

@yevgeny-shnaidman I apologize for the misunderstanding, this PR was about fixing the reconciler error regarding its client's cache `watch` permission on `ConfigMap`s. The module build flow is and was working correctly without this PR's change, i.e. without the `Module` reconciler's `watch` permission on `ConfigMap`s.

### Context
The cause of this error is based on the "internals" of the cache of the `controller-runtime/client`. By default this client when reading resources, e.g. when [Get()-ing](https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.13/pkg/client/typed_client.go#L135-L147) or `List()`ing, tries to read from its cache first. If there is no cache for the resource we're trying to read, then it [creates a new cached client](https://github.com/kubernetes-sigs/controller-runtime/blob/d9912251cd811210b2f205ecd3ecb2a5285c6bc6/pkg/client/client_cache.go#L54-L71) and then tries to build the respective cache. The cache setup includes setting up informers for the resource we're trying to read and those informers `watch` for such resources, in order to keep the cache as fresh as possible.

### Possible Solutions
In order to resolve this reconciler cache `watch` permission error (and only this error, as the build flow is anyway working as expected), we have 2 options:

1. setup a new client for this `ConfigMap` `client.Get()` call, with its cache disabled, so that no informers will be setup and no `watch` permission is needed
2. add the `watch` permission to `ConfigMap`s, so that the client can leverage caching

### This PR's solution

This PR followed option 2, but we can definitely revisit this if we would like to get rid of the `watch` permission on `ConfigMap`s.

Does this help?

yevgeny-shnaidman:
@mresvanis Thank you for explanation, very helpful

PR #180, opened 2022-12-05
add job watch to preflight controller setup

yevgeny-shnaidman:
This PR add a watch to preflight controller in order to react quicker to the Build/Sign completions. In addition the reconciliation scheduling timeout has been increased to 3 minutes

### DIFF
--- a/controllers/preflightvalidation_reconciler.go
+++ b/controllers/preflightvalidation_reconciler.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	batchv1 "k8s.io/api/batch/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


controllers/preflightvalidation_reconciler.go
```diff
@@ -105,6 +107,10 @@ func (r *PreflightValidationReconciler) Reconcile(ctx context.Context, req ctrl.
 		log.Info("PreflightValidation reconciliation success")
 		return ctrl.Result{}, nil
 	}
+	// if not all the modules has been reconciled, then maybe there is a problem with build
+	// configuration. Since build configuration is stored in the ConfigMap, and we cannot watch
+	// ConfigMap (since we don't know which ones to watch), then we need a requeue in order to run
```
ybettan, line 112: > since we don't know which ones to watch

How so? We know the name of the configmap from the module no?!

ybettan:
> in order to react quicker to the Build/Sign completions

Can you elaborate on why it is reacting quicker with this change? How was it reacting before if it didn't watch jobs?

controllers/preflightvalidation_reconciler.go
yevgeny-shnaidman: yes, but that can only be filtered after the event has been raised. You cannot filter out ConfigMap before event has been raised. Since there can be a lot of ConfigMap in the cluster, we don't want to start filter each time

yevgeny-shnaidman:
> > in order to react quicker to the Build/Sign completions
> 
> Can you elaborate on why it is reacting quicker with this change? How was it reacting before if it didn't watch jobs? Also, why did we increase the timeout?

Previously we had a timeout of 1 minutes, so reconcile loop was eventually scheduled. No we will get scheduled on the Build/Sign completion

PR #182, opened 2022-12-05
Build/Sign job selection should be done based on owner field also

yevgeny-shnaidman:
Currently, our code queries for jobs in 2 scenarious: 1) check if there is a running kaniko build/sign job for specific module/kernel 2) get all the job of the module during garbage collection scenario

With introduction of preflight flow we need to deferentiate the jobs also by owner, since preflight flow also uses and garbage collects build/sign jobs

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -403,7 +403,7 @@ func (r *ModuleReconciler) garbageCollect(ctx context.Context,
 	logger.Info("Garbage-collected DaemonSets", "names", deleted)
 
 	// Garbage collect for successfully finished build jobs
-	deleted, err = r.buildAPI.GarbageCollect(ctx, mod.Name, mod.Namespace)
+	deleted, err = r.buildAPI.GarbageCollect(ctx, mod.Name, mod.Namespace, mod)
 	if err != nil {
 		return fmt.Errorf("could not garbage collect build objects: %v", err)
 	}
--- a/internal/build/manager.go
+++ b/internal/build/manager.go
@@ -24,7 +24,7 @@ type Result struct {
 //go:generate mockgen -source=manager.go -package=build -destination=mock_manager.go
 
 type Manager interface {
-	GarbageCollect(ctx context.Context, modName, namespace string) ([]string, error)
+	GarbageCollect(ctx context.Context, modName, namespace string, owner metav1.Object) ([]string, error)
 
 	ShouldSync(
 		ctx context.Context,
--- a/internal/sign/job/manager.go
+++ b/internal/sign/job/manager.go
@@ -73,7 +73,7 @@ func (jbm *signJobManager) Sync(
 		return utils.Result{}, fmt.Errorf("could not make Job template: %v", err)
 	}
 
-	job, err := jbm.jobHelper.GetModuleJobByKernel(ctx, mod.Name, mod.Namespace, targetKernel, utils.JobTypeSign)
+	job, err := jbm.jobHelper.GetModuleJobByKernel(ctx, mod.Name, mod.Namespace, targetKernel, utils.JobTypeSign, owner)
 	if err != nil {
 		if !errors.Is(err, utils.ErrNoMatchingJob) {
 			return utils.Result{}, fmt.Errorf("error getting the signing job: %v", err)
--- a/internal/utils/jobhelper.go
+++ b/internal/utils/jobhelper.go
@@ -36,8 +36,8 @@ type Result struct {
 type JobHelper interface {
 	IsJobChanged(existingJob *batchv1.Job, newJob *batchv1.Job) (bool, error)
 	JobLabels(modName string, targetKernel string, jobType string) map[string]string
-	GetModuleJobByKernel(ctx context.Context, modName, namespace, targetKernel, jobType string) (*batchv1.Job, error)
-	GetModuleJobs(ctx context.Context, modName, namespace, jobType string) ([]batchv1.Job, error)
+	GetModuleJobByKernel(ctx context.Context, modName, namespace, targetKernel, jobType string, owner metav1.Object) (*batchv1.Job, error)
+	GetModuleJobs(ctx context.Context, modName, namespace, jobType string, owner metav1.Object) ([]batchv1.Job, error)
 	DeleteJob(ctx context.Context, job *batchv1.Job) error
 	CreateJob(ctx context.Context, jobTemplate *batchv1.Job) error
 	GetJobStatus(job *batchv1.Job) (Status, bool, error)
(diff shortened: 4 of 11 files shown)


controllers/module_reconciler.go
```diff
@@ -403,7 +403,7 @@ func (r *ModuleReconciler) garbageCollect(ctx context.Context,
 	logger.Info("Garbage-collected DaemonSets", "names", deleted)
 
 	// Garbage collect for successfully finished build jobs
-	deleted, err = r.buildAPI.GarbageCollect(ctx, mod.Name, mod.Namespace)
+	deleted, err = r.buildAPI.GarbageCollect(ctx, mod.Name, mod.Namespace, mod)
```
ybettan, line 406:
```suggestion
	deleted, err = r.buildAPI.GarbageCollect(ctx, mod)
```
Can we just pass `mod` and cast it to a `metav1.Object` before used in `GarbageCollect`?

internal/utils/jobhelper.go
```diff
@@ -36,8 +36,8 @@ type Result struct {
 type JobHelper interface {
 	IsJobChanged(existingJob *batchv1.Job, newJob *batchv1.Job) (bool, error)
 	JobLabels(modName string, targetKernel string, jobType string) map[string]string
-	GetModuleJobByKernel(ctx context.Context, modName, namespace, targetKernel, jobType string) (*batchv1.Job, error)
-	GetModuleJobs(ctx context.Context, modName, namespace, jobType string) ([]batchv1.Job, error)
+	GetModuleJobByKernel(ctx context.Context, modName, namespace, targetKernel, jobType string, owner metav1.Object) (*batchv1.Job, error)
+	GetModuleJobs(ctx context.Context, modName, namespace, jobType string, owner metav1.Object) ([]batchv1.Job, error)
```
ybettan, line 40: Can we remove `modName` and `namespace` if we pass the whole module anyway?

internal/utils/jobhelper.go
```diff
@@ -69,26 +69,32 @@ func (jh *jobHelper) JobLabels(modName string, targetKernel string, jobType stri
...
 
-func (jh *jobHelper) GetModuleJobs(ctx context.Context, modName, namespace, jobType string) ([]batchv1.Job, error) {
+func (jh *jobHelper) GetModuleJobs(ctx context.Context, modName, namespace, jobType string, owner metav1.Object) ([]batchv1.Job, error) {
 	matchLabels := moduleLabels(modName, jobType)
-	return jh.getJobs(ctx, namespace, matchLabels)
+	jobs, err := jh.getJobs(ctx, namespace, matchLabels)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get jobs for module %s, namespace %s: %v", modName, namespace, err)
+	}
+	foundJobs := filterJobsByOwner(jobs, owner)
```
ybettan, line 96:
```suggestion
	moduleOwnedJobs := filterJobsByOwner(jobs, owner)
```

internal/utils/jobhelper.go
```diff
@@ -150,3 +156,13 @@ func moduleLabels(moduleName, jobType string) map[string]string {
 		constants.JobType:         jobType,
 	}
 }
+
+func filterJobsByOwner(jobs []batchv1.Job, owner metav1.Object) []batchv1.Job {
+	foundJobs := []batchv1.Job{}
```
ybettan, line 161:
```suggestion
	ownedJobs := []batchv1.Job{}
```

controllers/module_reconciler.go
mresvanis: In the case of the `ManagedClusterModule` reconciler, where build/sign jobs are not deployed in the `Module`'s namespace, this API would not work. In addition, for `Preflight` and `ManagedClusterModule`, the owner of the jobs is not the `Module` .

If we decide to have the `Module` as an argument, then we need to add an additional `buildAPI.GarbageCollect()` function to accommodate the Preflight and ManagedClusterModule reconcilers.

internal/build/manager.go
```diff
@@ -24,7 +24,7 @@ type Result struct {
 //go:generate mockgen -source=manager.go -package=build -destination=mock_manager.go
 
 type Manager interface {
-	GarbageCollect(ctx context.Context, modName, namespace string) ([]string, error)
+	GarbageCollect(ctx context.Context, modName, namespace string, owner metav1.Object) ([]string, error)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: owner object does not have to be Module, it can be preflight, so it is best to keep as is

internal/sign/job/manager.go
```diff
@@ -73,7 +73,7 @@ func (jbm *signJobManager) Sync(
 		return utils.Result{}, fmt.Errorf("could not make Job template: %v", err)
 	}
 
-	job, err := jbm.jobHelper.GetModuleJobByKernel(ctx, mod.Name, mod.Namespace, targetKernel, utils.JobTypeSign)
+	job, err := jbm.jobHelper.GetModuleJobByKernel(ctx, mod.Name, mod.Namespace, targetKernel, utils.JobTypeSign, owner)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: The same answer as above, the owner should be a different field, since it might now be Module, but Preflight or something else

internal/utils/jobhelper.go
yevgeny-shnaidman: fixed

internal/utils/jobhelper.go
```diff
@@ -69,26 +69,32 @@ func (jh *jobHelper) JobLabels(modName string, targetKernel string, jobType stri
...
-func (jh *jobHelper) GetModuleJobByKernel(ctx context.Context, modName, namespace, targetKernel, jobType string) (*batchv1.Job, error) {
+func (jh *jobHelper) GetModuleJobByKernel(ctx context.Context, modName, namespace, targetKernel, jobType string, owner metav1.Object) (*batchv1.Job, error) {
 	matchLabels := moduleKernelLabels(modName, targetKernel, jobType)
 	jobs, err := jh.getJobs(ctx, namespace, matchLabels)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get module %s, jobs by kernel %s: %v", modName, targetKernel, err)
 	}
 
-	numFoundJobs := len(jobs)
+	foundJobs := filterJobsByOwner(jobs, owner)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/module_reconciler.go
mresvanis: The PR regarding the `MangedClusterModule` reconciler build garbage collection is still open, but [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/178/files#diff-ecbdbfaa3e0752590ccb7b5c0387c308de661ba622120129edf662c02fa1ddadR141) you could see how this API is used. The major point here is that the `namespace` is not the `Module'`s namespace, as the builds/signs only get deployed in the Hub cluster, whereas the `Module`s are deployed on the Spoke clusters.

I completely understand your point though, i.e. currently this API not that clear.

ybettan: Sorry. Missed the sign part.
Ack.

ybettan:
/lgtm
/hold
Wait for Github actions that are currently not enforced by `tide`.

PR #183, opened 2022-12-06
Fixing sign job hash annotation

yevgeny-shnaidman:
Sign job annotation is used for deciding whether update flow should be executed on the job ( deletion and creation anew with new parameters). Currently the code includes verification of currently running job hash versus the new parameters hash, but the actual code that sets the hash annotation during job template is missing. This PR does the following:
1) add hash annotation calculation during job template creation 2) unit test

### DIFF
--- a/internal/sign/job/signer.go
+++ b/internal/sign/job/signer.go
@@ -1,6 +1,7 @@
 package signjob
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


internal/sign/job/signer.go
```diff
@@ -112,29 +128,54 @@ func (m *signer) MakeJobTemplate(
...
+						RestartPolicy: v1.RestartPolicyOnFailure,
+						Volumes:       volumes,
+						NodeSelector:  mod.Spec.Selector,
 					},
-					RestartPolicy: v1.RestartPolicyOnFailure,
-					Volumes:       volumes,
-					NodeSelector:  mod.Spec.Selector,
 				},
-			},
+			*/
```
ybettan, line 178: Forgot to remove.

internal/sign/job/signer.go
```diff
@@ -144,3 +185,43 @@ func (m *signer) MakeJobTemplate(
...
+	if err != nil {
+		return 0, fmt.Errorf("failed to get private secret %s for signing: %v", privateSecret, err)
+	}
+	publicKeyData, err := s.getSecretData(ctx, publicSecret, constants.PublicSignDataKey, namespace)
+	if err != nil {
+		return 0, fmt.Errorf("failed to get public secret %s for signing: %v", publicSecret, err)
+	}
+
+	return getHashValue(podTemplate, publicKeyData, privateKeyData)
+}
```
ybettan, line 200: I see that build is doing something similar.
Should we extract it to a `Hash` API (with different implementation for `build` and `sign`) and be consumed in `build` and `sign` logic?

Same for `getHashValue`

internal/sign/job/signer.go
yevgeny-shnaidman: fixed

internal/sign/job/signer.go
yevgeny-shnaidman: it means that you want to define them as an interface, right? but what's the point? it is not a problem to test it in unit test. We will end up with 2 different implementation  ( just as we have right now), but it addition the API will not be as clear as it is now

ybettan: The point is that this is not a build/sign logic, it is a hash logic, extracted to it's own API and consumed by another API (using dependency injection)

ybettan: As agreed, `Hash` will be extracted to its own API and consumed twice in `build` and `sign`.
https://github.com/kubernetes-sigs/kernel-module-management/issues/184.

PR #186, opened 2022-12-06
Adding sign verification to the preflight flow

yevgeny-shnaidman:
This PR add the sign verification stage to the preflight: in case image verification has failed and either build section is not defined or build verification was successful, preflight will now verify sign , if it was defined in the Module

### DIFF
--- a/internal/preflight/preflight.go
+++ b/internal/preflight/preflight.go
@@ -9,6 +9,7 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/sign"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/statusupdater"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 
@@ -32,10 +33,11 @@ type PreflightAPI interface {
 func NewPreflightAPI(
 	client client.Client,
 	buildAPI build.Manager,
+	signAPI sign.SignManager,
 	registryAPI registry.Registry,
 	statusUpdater statusupdater.PreflightStatusUpdater,
 	kernelAPI module.KernelMapper) PreflightAPI {
-	helper := newPreflightHelper(client, buildAPI, registryAPI)
+	helper := newPreflightHelper(client, buildAPI, signAPI, registryAPI)
 	return &preflight{
 		kernelAPI:     kernelAPI,
 		statusUpdater: statusUpdater,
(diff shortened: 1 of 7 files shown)


internal/preflight/preflight.go
```diff
@@ -63,39 +65,60 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
+	if shouldBuild {
+		err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageBuild)
+		if err != nil {
+			log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to build stage"), "module", mod.Name, "error", err)
+		}
+
+		verified, msg = p.helper.verifyBuild(ctx, pv, mapping, mod)
+		if !verified {
+			shouldSign = false
+		}
```
ybettan, line 91:
```suggestion
		if verified, msg = p.helper.verifyBuild(ctx, pv, mapping, mod); !verified {
			return false, msg
		}
```

internal/preflight/preflight.go
```diff
@@ -63,39 +65,60 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
+		shouldBuild = false
+		shouldSign = false
 	}
 
-	err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageBuild)
-	if err != nil {
-		log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to build stage"), "module", mod.Name, "error", err)
+	if shouldBuild {
+		err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageBuild)
+		if err != nil {
```
ybettan, line 84:
```suggestion
		if err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageBuild); err != nil {
```

internal/preflight/preflight.go
```diff
@@ -63,39 +65,60 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
+		verified, msg = p.helper.verifyBuild(ctx, pv, mapping, mod)
+		if !verified {
+			shouldSign = false
+		}
 	}
 
-	return p.helper.verifyBuild(ctx, pv, mapping, mod)
+	if shouldSign {
+		err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageSign)
+		if err != nil {
```
ybettan, line 96:
```suggestion
		if err = p.statusUpdater.PreflightSetVerificationStage(ctx, pv, mod.Name, kmmv1beta1.VerificationStageSign); err != nil {
```

internal/preflight/preflight.go
```diff
@@ -63,39 +65,60 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
+			log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to sign stage"), "module", mod.Name, "error", err)
+		}
+		verified, msg = p.helper.verifySign(ctx, pv, mapping, mod)
```
ybettan, line 99:
```suggestion
		if verified, msg = p.helper.verifySign(ctx, pv, mapping, mod); !verified {
		    return false, msg
		}
```

internal/preflight/preflight.go
```diff
@@ -63,39 +65,60 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
+	}
+	return verified, msg
```
ybettan, line 101:
```suggestion
	return true, fmt.Sprintf("Preflight upgrade check went successfuly")
```

ybettan: I think this whole function will be more readable if written like:
```
diff --git a/internal/preflight/preflight.go b/internal/preflight/preflight.go
index ebeea84..5623d06 100644
--- a/internal/preflight/preflight.go
+++ b/internal/preflight/preflight.go
@@ -75,8 +75,7 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr

        verified, msg := p.helper.verifyImage(ctx, mapping, mod, kernelVersion)
        if verified {
-               shouldBuild = false
-               shouldSign = false
+               return false, msg
        }

        if shouldBuild {
@@ -85,9 +84,8 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
                        log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to build stage"), "module", mod.Name, "error", err)
                }

-               verified, msg = p.helper.verifyBuild(ctx, pv, mapping, mod)
-               if !verified {
-                       shouldSign = false
+               if verified, msg = p.helper.verifyBuild(ctx, pv, mapping, mod); !verified {
+                       return false, msg
                }
        }

@@ -96,9 +94,11 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
                if err != nil {
                        log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to sign stage"), "module", mod.Name, "error", err)
                }
-               verified, msg = p.helper.verifySign(ctx, pv, mapping, mod)
+               if verified, msg = p.helper.verifySign(ctx, pv, mapping, mod); !verified {
+                       return false, msg
+               }
        }
-       return verified, msg
+       return true, fmt.Sprintf("Preflight upgrade check went successfuly")
 }

```

internal/preflight/preflight.go
```diff
@@ -153,3 +176,31 @@ func (p *preflightHelper) verifyBuild(ctx context.Context,
...
+	mapping *kmmv1beta1.KernelMapping,
+	mod *kmmv1beta1.Module) (bool, string) {
+	log := ctrlruntime.LoggerFrom(ctx)
+
+	previousImage := ""
+	if module.ShouldBeBuilt(mod.Spec, *mapping) {
+		previousImage = module.IntermediateImageName(mod.Name, mod.Namespace, mapping.ContainerImage)
+	}
+
+	// at this stage we know that eiher mapping Sign or Container sign are defined
```
ybettan, line 191: Why is that?

internal/preflight/preflight.go
```diff
@@ -153,3 +176,31 @@ func (p *preflightHelper) verifyBuild(ctx context.Context,
...
+	signRes, err := p.signAPI.Sync(ctx, *mod, *mapping, pv.Spec.KernelVersion, previousImage, pv.Spec.PushBuiltImage, pv)
+	if err != nil {
+		return false, fmt.Sprintf("Failed to verify signing for module %s, kernel version %s, error %s", mod.Name, pv.Spec.KernelVersion, err)
+	}
+
+	if signRes.Status == utils.StatusCompleted {
+		msg := "sign completes"
+		if pv.Spec.PushBuiltImage {
+			msg += " and image pushed"
```
ybettan, line 200: This doesn't seems right? Why are we adding "build" info in the `sign` section? Can it be added elsewhere?

internal/preflight/preflight.go
```diff
@@ -63,39 +65,60 @@ func (p *preflight) PreflightUpgradeCheck(ctx context.Context, pv *kmmv1beta1.Pr
...
 		log.Info(utils.WarnString("failed to update the stage of Module CR in preflight to image stage"), "module", mod.Name, "error", err)
 	}
 
-	imageVerified, msg := p.helper.verifyImage(ctx, mapping, mod, kernelVersion)
-	if imageVerified || (mapping.Build == nil && mod.Spec.ModuleLoader.Container.Build == nil) {
-		return imageVerified, msg
+	verified, msg := p.helper.verifyImage(ctx, mapping, mod, kernelVersion)
+	if verified {
+		shouldBuild = false
+		shouldSign = false
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
yevgeny-shnaidman: fixed

internal/preflight/preflight.go
yevgeny-shnaidman: i prefer to not change the code: i personally think it is more readable

internal/preflight/preflight.go
yevgeny-shnaidman: i prefer to not change the code: i personally think it is more readable

internal/preflight/preflight.go
yevgeny-shnaidman: 2 reason why not to change it:
1) verified can be false - for example verifySign failed
2) the message is constructed in each handle specifically, so we want to return it

internal/preflight/preflight.go
yevgeny-shnaidman: are you talking about the comment? Because if we entered the verifySign function, then the verification that either Module or Mapping sign sections are present, has been done above

internal/preflight/preflight.go
yevgeny-shnaidman: not sure i understand what you are talking about... 
this one: pv.Spec.PushBuiltImage?

internal/preflight/preflight.go
ybettan: > verified can be false - for example verifySign failed

If `verifySign` failed then we already returned `false` earlier.

> the message is constructed in each handle specifically, so we want to return it

I am OK with return `msg` instead of `fmt.Sprintf("Preflight upgrade check went successfuly")` I have suggested.

PR #187, opened 2022-12-06
add support for KERNEL_VERSION replacement for signing

chr15p:
filesToSign (the filepaths of the kmods to sign) and unsignedImage (the name of the image that contains the kmods if its not built in cluster) both normally would contain the kernel version, this commit allows the embedding of "$(KERNEL_VERSION)" in them and have it replaced with the actual kernel version

### DIFF
--- a/internal/sign/helper.go
+++ b/internal/sign/helper.go
@@ -2,51 +2,57 @@ package sign
 
 import (
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 )
 
 //go:generate mockgen -source=helper.go -package=sign -destination=mock_helper.go
 
 type Helper interface {
-	GetRelevantSign(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping) *kmmv1beta1.Sign
+	GetRelevantSign(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping, kernel string) (*kmmv1beta1.Sign, error)
 }
 
-type helper struct{}
+type helper struct {
+}
 
 func NewSignerHelper() Helper {
 	return &helper{}
 }
 
-func (m *helper) GetRelevantSign(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping) *kmmv1beta1.Sign {
+func (m *helper) GetRelevantSign(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping, kernel string) (*kmmv1beta1.Sign, error) {
+	var signConfig *kmmv1beta1.Sign
 	if modSpec.ModuleLoader.Container.Sign == nil {
 		// km.Sign cannot be nil in case mod.Sign is nil, checked above
-		return km.Sign.DeepCopy()
+		signConfig = km.Sign.DeepCopy()
+	} else if km.Sign == nil {
+		signConfig = modSpec.ModuleLoader.Container.Sign.DeepCopy()
+	} else {
+		signConfig = modSpec.ModuleLoader.Container.Sign.DeepCopy()
+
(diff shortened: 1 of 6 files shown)


internal/sign/helper.go
```diff
@@ -2,12 +2,13 @@ package sign
...
 import (
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"strings"
 )
 
 //go:generate mockgen -source=helper.go -package=sign -destination=mock_helper.go
 
 type Helper interface {
-	GetRelevantSign(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping) *kmmv1beta1.Sign
+	GetRelevantSign(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping, macros map[string]string) *kmmv1beta1.Sign
```
ybettan, line 11: Why do we need to support a `map[string]string` of changes if all we ever do is just changing `KERNEL_VERSION`?

ybettan: Ack. In that case, I will comment on the implementation of the map.

internal/sign/helper.go
```diff
@@ -16,29 +17,45 @@ func NewSignerHelper() Helper {
...
-		// km.Sign cannot be nil in case mod.Sign is nil, checked above
-		return km.Sign.DeepCopy()
-	}
+func (m *helper) replaceInParam(macros map[string]string, params ...string) []string {
 
-	if km.Sign == nil {
-		return modSpec.ModuleLoader.Container.Sign.DeepCopy()
+	var replacedStrings []string
+	for k, v := range macros {
+		m := "${" + k + "}"
```
ybettan, line 24: nit.
```suggestion
		template := "${" + k + "}"
```

chr15p: as always you are correct, good nit, done!

internal/sign/helper.go
```diff
@@ -16,29 +17,45 @@ func NewSignerHelper() Helper {
...
-		// km.Sign cannot be nil in case mod.Sign is nil, checked above
-		return km.Sign.DeepCopy()
-	}
+func (m *helper) replaceInParam(macros map[string]string, params ...string) []string {
 
-	if km.Sign == nil {
-		return modSpec.ModuleLoader.Container.Sign.DeepCopy()
+	var replacedStrings []string
+	for k, v := range macros {
+		template := "${" + k + "}"
```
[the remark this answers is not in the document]
chr15p: that sounded so simple.... 

I've created an internal/utils/replaceStrings.go that contains functions to a) parse the kernel into major/minor/bugfix fields, and b) uses envsubst to do a find and replace with the idea that it will be much easier for users if we have a single set of substitutions we use everywhere.  I have then moved this PR to use it.

In some ways this  code duplicates whats in module/kernelmapper.go but is trying to be generic so we can use it anywhere else that needs similar functionality.  kernelmapper should probably be moved to being wrappers around replaceStrings as well, but I think that would make this PR much bigger and more sprawling than is good,

internal/utils/replaceStrings.go
```diff
@@ -0,0 +1,51 @@
...
+	"strings"
+)
+
+const (
+	kernelVersionMajorIdx = 0
+	kernelVersionMinorIdx = 1
+	kernelVersionPatchIdx = 2
+)
+
+func MapToEnvVars(m map[string]string) []string {
```
yevgeny-shnaidman, line 16: This function is not used anywhere, let's remove it

internal/sign/helper_test.go
```diff
@@ -35,8 +36,9 @@ var _ = Describe("GetRelevantSign", func() {
 	}
 
 	DescribeTable("should set fields correctly", func(mod kmmv1beta1.Module, km kmmv1beta1.KernelMapping) {
+		//macros := map[string]string{"KERNEL_VERSION": kernelVersion}
```
qbarrand, line 40:
```suggestion
```

PR #191, opened 2022-12-08
Adjust ClusterClaim permissions to enable creation and caching

mresvanis:
The `NodeKernelClusterClaim` reconciler requires permissions and leverages client caching on `ClusterClaim` resources. This change adds permissions to `[create,get,list,watch]` `ClusterClaim` resources, but only `[patch,update,delete]` on the specific `ClusterClaim` that the reconciler itself has created.

### DIFF
--- a/controllers/node_kernel_clusterclaim.go
+++ b/controllers/node_kernel_clusterclaim.go
@@ -22,7 +22,8 @@ import (
 )
 
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;patch;list;watch
-//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=create;delete;get;list;patch;update;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=create;get;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=delete;patch;update
 
 const clusterClaimName = "kernel-versions.kmm.node.kubernetes.io"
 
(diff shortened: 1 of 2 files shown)


controllers/node_kernel_clusterclaim.go
```diff
@@ -22,7 +22,8 @@ import (
 )
 
 //+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;patch;list;watch
-//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=create;delete;get;list;patch;update;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=create;get;list;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=delete;patch;update
```
qbarrand, line 26:
```suggestion
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=get;list;watch
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=create;delete;patch;update
```

mresvanis: I found out the hard way that this doesn't work :)

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources

> Note: You cannot restrict create or deletecollection requests by their resource name. For create, this limitation is because the name of the new object may not be known at authorization time. If you restrict list or watch by resourceName, clients must include a metadata.name field selector in their list or watch request that matches the specified resourceName in order to be authorized. For example, kubectl get configmaps --field-selector=metadata.name=my-configmap

qbarrand: Thank you for digging!

PR #192, opened 2022-12-08
Signing insecure options

chr15p:
Add support for insecure pull/push of images to signing job.
    
For some reason api support was there but pull/push via http or without TLS cert checking hadn't made it into the signing job itself. This fixes that issue by passing those options into the signing job and through to the appropriate functions in the registry package.


ybettan:
@chr15p That's by bad. I have added those options, and new I need to create a task to update the signing job but it just slipped out of my mind.

internal/registry/registry.go
```diff
@@ -427,3 +375,52 @@ func (r *registry) ExtractFileToFile(destination string, header *tar.Header, tar
...
+func (r *registry) WriteImageByName(imageName string, image v1.Image, auth authn.Authenticator, insecure bool, skipTLSVerify bool) error {
+	options := r.getTransportOptions(insecure, skipTLSVerify)
+	options = append(
+		options,
+		crane.WithAuth(auth),
+	)
+
+	err := crane.Push(image, imageName, options...)
+	if err != nil {
+		return fmt.Errorf("failed to push signed image: %w", err)
```
ybettan, line 408:
```suggestion
		return fmt.Errorf("failed to push signed image: %v", err)
```

After some discussions, it has been agreed to not wrap errors unless the caller of the function that may return an error is using `Is(err)` or `As(err)`.

internal/registry/registry.go
```diff
@@ -427,3 +375,52 @@ func (r *registry) ExtractFileToFile(destination string, header *tar.Header, tar
...
+func (r *registry) GetImageByName(imageName string, auth authn.Authenticator, insecure bool, skipTLSVerify bool) (v1.Image, error) {
+	options := r.getTransportOptions(insecure, skipTLSVerify)
+	options = append(
+		options,
+		crane.WithAuth(auth),
+	)
+
+	img, err := crane.Pull(imageName, options...)
+	if err != nil {
+		return nil, fmt.Errorf("could not get image: %w", err)
```
ybettan, line 422:
```suggestion
		return nil, fmt.Errorf("could not get image: %v", err)
```

After some discussions, it has been agreed to not wrap errors unless the caller of the function that may return an error is using `Is(err)` or `As(err)`.

chr15p:
@ybettan they were on my list of things to do as well, then I realised you'd already done the hard part :) 
I have updated to using %v for errors, and taken the opportunity to change a number of "my" functions in the registry package where I knew it would have no effect.

ybettan:
Thanks @chr15p 
/lgtm
/assign @yevgeny-shnaidman

internal/registry/registry.go
```diff
@@ -119,25 +118,7 @@ func (r *registry) getPullOptions(ctx context.Context, image string, tlsOptions
...
-			rt := http.DefaultTransport.(*http.Transport).Clone()
-			rt.TLSClientConfig.InsecureSkipVerify = true
-
-			options = append(
-				options,
-				crane.WithTransport(rt),
-			)
-		}
-	}
+	options := r.getTransportOptions(tlsOptions.Insecure, tlsOptions.InsecureSkipTLSVerify)
```
yevgeny-shnaidman, line 121: tlsOptions can be nil, no? so calling this function without verifying will cause a panic

internal/registry/registry.go
```diff
@@ -53,14 +52,14 @@ type Registry interface {
 	VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion, moduleFileName string) bool
 	GetLayersDigests(ctx context.Context, image string, tlsOptions *kmmv1beta1.TLSOptions, registryAuthGetter auth.RegistryAuthGetter) ([]string, *RepoPullConfig, error)
 	GetLayerByDigest(digest string, pullConfig *RepoPullConfig) (v1.Layer, error)
-	WriteImageByName(imageName string, image v1.Image, auth authn.Authenticator) error
 	WalkFilesInImage(image v1.Image, fn func(filename string, header *tar.Header, tarreader io.Reader, data []interface{}) error, data ...interface{}) error
 	GetLayerMediaType(image v1.Image) (types.MediaType, error)
 	AddLayerToImage(tarfile string, image v1.Image) (v1.Image, error)
-	GetImageByName(imageName string, auth authn.Authenticator) (v1.Image, error)
 	ParseReference(imageName string) (name.Reference, error)
```
yevgeny-shnaidman, line 58: Do we need this function at all now?

internal/registry/registry.go
chr15p: Good spot. I have reverted all the changes to this function.

I was trying to reduce the duplicated code but my testing is all focused on the signing image and doesn't call this function at all so I missed the risk of panic here. I still think we should replace this code with a call to getTransportOptions() but it is greatly expanding the surface area of this change into stuff like preflight so it probably needs to be in a commit of its own that can be tested separately.

PR #203, opened 2022-12-14
Add modes to the manager binary

qbarrand:
Rework `main.go` logic to add controllers depending on the mode.
Use `kmm` mode by default.
Upgrade to controller-gen v0.10.0.

/cc @yevgeny-shnaidman


qbarrand:
Superseded by #204.

PR #204, opened 2022-12-14
Add a Config object

qbarrand:
Make it possible to configure which controllers are started via a new Config object.

/cc @yevgeny-shnaidman


cmd/manager/main.go
```diff
@@ -107,87 +119,162 @@ func main() {
 		LeaderElectionID:       "c5baf8af.sigs.k8s.io",
 	}
 	if configFile != "" {
```
yevgeny-shnaidman, line 121: i think it it better to check that configFile is empty and fail, and then continue with the rest of the code

api/v1beta1/config_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// If true, allows KMM to build images.
+	AllowBuilds bool `json:"allowBuilds"`
+
+	// If true, allows KMM to start kmod signing jobs.
+	AllowSigning bool `json:"allowSigning"`
+}
+
+type Reconcilers struct {
+	// ManagedClusterModule contains settings for the ManagedClusterModule reconciler.
```
yevgeny-shnaidman, line 39: We have essentially 3 modes:
1) Regular Module
2) Hub
3) Spoke
Why not just define those 3 flags, and then initialize the reconcilers in the main according to those flags? Having flag for each reconciler is a little bit redundant and confusing: we cannot have ManagedClusterModule false and NodeKernelClusterClaim true and visa-versa, no? WDYT?

api/v1beta1/config_types.go
```diff
@@ -0,0 +1,84 @@
...
+
+type Feature struct {
+	Enabled bool `json:"enabled"`
+}
+
+type ModuleReconcilerFeature struct {
+	Feature `json:",inline"`
+
+	// If true, allows KMM to build images.
+	AllowBuilds bool `json:"allowBuilds"`
```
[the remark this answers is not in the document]
qbarrand: This is another safety net for Spokes, to make sure that they will not build under any circumstance.

cmd/manager/main.go
qbarrand: Agreed. Actually I think we should deprecate all command-line flags in favor of the config file. @mresvanis WDYT?

api/v1beta1/config_types.go
qbarrand: That was the spirit of #203. But instead of defining those modes in the code, we will create 3 bundles which will enable or disable the appropriate reconcilers for each use case. I think that's more flexible and future-proof.

api/v1beta1/config_types.go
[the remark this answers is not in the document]
yevgeny-shnaidman: do we need another safety net? our code is creating the Modules for Spoke, no?

api/v1beta1/config_types.go
yevgeny-shnaidman: i agree this is more flexible, but the code becomes much less readable: newcomers, or somebody not familiar with the one of the modes won't be able to understand from the code which controllers are needed in which mode.

cmd/manager/main.go
mresvanis: I would suggest that we deprecate the command-line flags altogether and just have the config file as the only way  to configure the operator. 

Having one way only would ease development and maintenance and at the same time it would follow [what](https://github.com/kubernetes/community/blob/master/archive/wg-component-standard/component-config/README.md) the Kubernetes components (e.g. `kubelet`, `scheduler`) are doing, which is replacing command-line flags with the config file.

qbarrand:
Closing in favour of #218.

PR #211, opened 2022-12-16
Add README.md to docs

enriquebelarte:
/hold

### DIFF
--- a/docs/mkdocs/index.md
+++ b/docs/mkdocs/index.md
@@ -1,11 +1,14 @@
 # Home
 
-Kernel Module Management optionally builds and runs DriverContainers on Kubernetes clusters.
+## Overview
 
-## Getting started
+Kernel Module Management (KMM) is a Kubernetes operator that manages, builds, signs and deploys out-of-tree kernel modules and device plugins on Kubernetes clusters.
 
-Install the bleeding edge version of the KMM Operator by running the following command:
 
-```bash
-kubectl apply -k https://github.com/kubernetes-sigs/kernel-module-management/config/default
-```
+KMM adds a new `Module` CRD which describes the desired state of an out-of-tree kernel module and its associated device plugin. `Module` resources contain fields that configure how to load the module, associates ModuleLoader images with kernel versions, and optionally instructions to build and sign modules for specific kernel versions.
+KMM is designed to accomodate several kernel versions at once for any kernel module, allowing for seamless node upgrades and reduced application downtime.
+
+## Installation Guide
+
+Check the [Install](documentation/install.md) section for instructions.
+


qbarrand:
Could this go to `docs/mkdocs/index.md` instead?

docs/README.md
```diff
@@ -0,0 +1,23 @@
+# Overview
+
+Kernel Module Management also known as KMM is a Kubernetes operator which provides a simple way to manage, build, sign or deploy out-of-tree kernel modules, drivers or device plugins on selected nodes of an existing Kubernetes cluster.
+
+Once installed we will get a Deployment object consisting of a ReplicaSet with a pod acting as a controller manager in the new Namespace kmm-operator-system.
```
ybettan, line 5:
```suggestion
Once installed we will get a Deployment object consisting of a ReplicaSet with a pod acting as a controller manager in the new Namespace `kmm-operator-system`. 
```

docs/README.md
```diff
@@ -0,0 +1,23 @@
+# Overview
+
+Kernel Module Management also known as KMM is a Kubernetes operator which provides a simple way to manage, build, sign or deploy out-of-tree kernel modules, drivers or device plugins on selected nodes of an existing Kubernetes cluster.
+
+Once installed we will get a Deployment object consisting of a ReplicaSet with a pod acting as a controller manager in the new Namespace kmm-operator-system. 
+
+This controller provides the use of a new Kubernetes API object called Module which sets if the driver image already exists and has to be loaded, if it has to be in-cluster built (a ConfigMap should be provided to get this), if it will use a device plugin and also on which nodes would be deployed as a new DaemonSet will be created for each node’s kernel version matching the settings in a DriverContainer or device plugin within the Module definition.
```
ybettan, line 7:
```suggestion
This controller provides the use of a new Kubernetes API object called `Module` which sets if the driver image already exists and has to be loaded, if it has to be in-cluster built (a ConfigMap should be provided to get this), if it will use a device plugin and also on which nodes would be deployed as a new DaemonSet will be created for each node’s kernel version matching the settings in a DriverContainer or device plugin within the Module definition.
```

docs/README.md
```diff
@@ -0,0 +1,23 @@
...
+- A Namespace kmm-operator-system where manager will run.
+- Serviceaccount and RBAC permissions for KMM to operate.
+- The new CustomResourceDefinitions to add in the Kubernetes API.
+- The deployment of the manager itself.
+
+The following command will install the operator with all above necessary components:
+
+```bash
+kubectl apply -k https://github.com/kubernetes-sigs/kernel-module-management/config/default
+```
```
ybettan, line 23: I will suggest referring to https://github.com/kubernetes-sigs/kernel-module-management#getting-started instead of duplicating the command.

The main reason for not having duplication is that in the future we may update this command in the README.md and forget to update it here as well, hence, having an outdated doc.

ybettan:
@enriquebelarte Do we still need that PR after https://github.com/kubernetes-sigs/kernel-module-management/pull/235 has merged?

enriquebelarte:
> @enriquebelarte Do we still need that PR after #235 has merged?

I'm not sure if it still makes sense. WDYT @qbarrand ?

iranzo:
I think that midstream should be in sync with upstream + extra stuff relevant to OCP and not to Kubernetes

iranzo (approved):
Thanks a lot!

docs/mkdocs/index.md
```diff
@@ -1,11 +1,14 @@
 # Home
 
-Kernel Module Management optionally builds and runs DriverContainers on Kubernetes clusters.
+## Overview
 
-## Getting started
+Kernel Module Management also known as KMM is a Kubernetes operator which provides a simple way to manage, build, sign or deploy out-of-tree kernel modules, drivers or device plugins on selected nodes of an existing Kubernetes cluster.
```
qbarrand, line 5:
```suggestion
Kernel Module Management (KMM) is a Kubernetes operator that manages, builds, signs and deploys out-of-tree kernel modules and device plugins on Kubernetes clusters.
```

docs/mkdocs/index.md
```diff
@@ -1,11 +1,14 @@
...
 
-Install the bleeding edge version of the KMM Operator by running the following command:
+Once installed we will get a Deployment object consisting of a ReplicaSet with a pod acting as a controller manager in the new Namespace `kmm-operator-system`.
```
qbarrand, line 7: I would remove this. If installed with OLM, the namespace is configurable, and the Deployment / ReplicaSet bits are common to most operators.
```suggestion
```

docs/mkdocs/index.md
```diff
@@ -1,11 +1,14 @@
...
+
+This controller provides the use of a new Kubernetes API object called `Module` which sets if the driver image already exists and has to be loaded, if it has to be built in-cluster (a ConfigMap should be provided to get this), if it will use a device plugin and also on which nodes would be deployed as a new DaemonSet will be created for each node’s kernel version matching the settings in a DriverContainer or device plugin within the Module definition.
```
qbarrand, line 9:
```suggestion
KMM adds a new `Module` CRD which describes the desired state of an out-of-tree kernel module and its associated device plugin. `Module` resources contain fields that configure how to load the module, associates ModuleLoader images with kernel versions, and optionally instructions to build and sign modules for specific kernel versions.
KMM is designed to accomodate several kernel versions at once for any kernel module, allowing for seamless node upgrades and reduced application downtime.
```

PR #214, opened 2022-12-19
Move to the `kmm.sigs.x-k8s.io` API group

qbarrand:
Move to the `x-k8s.io` API Group as we do not need a strict API review process.


ybettan:
Can we also remove https://github.com/ybettan/kernel-module-management/blob/ca81e844926ffbbb82c55a853c2fdad3291df2e4/Makefile#L100-L103 as part of the PR?

ybettan:
/hold
Wait for optional jobs

PR #217, opened 2022-12-20
Improve push/pull secret handling in signing

chr15p:
Re-work secret handling for pulling and pushing images Add support for dockerconfigjson and dockercfg style secrets Add support for multiple keys in a single secret (useful for pulling and pushing to different registries)

This also sets the stage for allowing service account secrets

We do this by mounting secrets in their own directories then reading all the files (keys) in that dir

### DIFF
--- a/cmd/signimage/signimage.go
+++ b/cmd/signimage/signimage.go
@@ -3,20 +3,143 @@ package main
 import (
 	"archive/tar"
 	"bytes"
+	"encoding/json"
 	"flag"
 	"fmt"
-	"github.com/docker/cli/cli/config"
-	dockertypes "github.com/docker/cli/cli/config/types"
 	"github.com/go-logr/logr"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
 	"io"
+	"io/fs"
 	"k8s.io/klog/v2/klogr"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 )
+
+// configDir should contain all the secrets available as individual files named for their keys
+// so we need to search through for the secrets we need for our pull and push repos.
+// If we do not find any authn, default to anonymous.
+// we're (trying to be) tolerant in our inputs, precise in our outputs, and opaque in our comments. Its the UNIX way!
+type repoAuth struct {
+	pullRepo  string
+	pushRepo  string
+	PullAuth  authn.Authenticator
+	PushAuth  authn.Authenticator
+	configDir string
+}
+
+func NewRepoAuth(configDir string, pullRepo string, pushRepo string) *repoAuth {
+	if pushRepo == "" {
+		pushRepo = pullRepo
+	}
+	r := &repoAuth{
+		pullRepo:  pullRepo,
+		pushRepo:  pushRepo,
+		PullAuth:  nil,
+		PushAuth:  nil,
+		configDir: filepath.Clean(configDir),
+	}
+	r.PopulateAuthFromFileList()
+	// if we haven't found appropriate secrets try anonymous
+	// its not clear if this is what the user wants so we need to log it and let them decide
+	if r.PullAuth == nil {
+		r.PullAuth = authn.Anonymous
+		logger.Info("no secrets for pull found, default to Anonymous")
+	}
+	if r.PushAuth == nil {
+		r.PushAuth = authn.Anonymous
+		logger.Info("no secrets for push found, default to Anonymous")
+	}
+	return r
+}
+
+func (r *repoAuth) PopulateAuthFromFileList() {
+	// not giving any secrets should short cuircuit the whole process
+	if r.configDir == "" {
+		return
+	}
+	logger.Info("walking the tree looking for secrets", "dir", r.configDir)
+	// otherwise lets hunt for repo secrets!
+	err := filepath.WalkDir(r.configDir, r.authDirFileHandler)
+	if err != nil {
+		logger.Info("no secret found, default to Anonymous", "error", err)
+		// we could return the error and die, but we want to be tolerant of secrets in the wrong format etc
+		return
+	}
+}
+
+func (r *repoAuth) authDirFileHandler(path string, d fs.DirEntry, err error) error {
+	if err != nil {
+		//theres an error in WalkDir (either fs.Stat or Readdir)
+		//not much we can do except log it, and move on
+		logger.Info("error walking tree", "filename", path, "error", err)
+		return nil
+	}
+	if path == r.configDir {
+		return nil
+	}
+	if strings.Contains(path, "..") {
+		if d.IsDir() {
+			return filepath.SkipDir
+		}
+		return nil
+	}
+	if d.IsDir() {
+		return nil
+	}
+	logger.Info("checking auth file ", "filename", path)
+	err = r.getAuthFromSingleFile(path)
+	if err != nil {
+		logger.Info("unable to read auth file", "filename", path, "error=", err)
+	}
+
+	return nil
+}
+
+func (r *repoAuth) getAuthFromSingleFile(secretFile string) error {
+	jsonFile, err := os.Open(secretFile)
+	if err != nil {
+		return fmt.Errorf("unable to open secret file %s: %v", secretFile, err)
+	}
+	// defer the closing of our jsonFile so that we can parse it later o
+	defer jsonFile.Close()
+
+	byteValue, err := io.ReadAll(jsonFile)
+	if err != nil {
+		return fmt.Errorf("error reading secret file %s: %v", secretFile, err)
+	}
+
+	var result map[string]json.RawMessage
+	var allAuthConfigs map[string]authn.AuthConfig
+
+	err = json.Unmarshal(byteValue, &result)
+	if err != nil {
+		return fmt.Errorf("error unmarshalling %s: %v", secretFile, err)
+	}
+
+	// either our secret file is a map containing a map of repo secrets
+	// or its just a straight map of secrets, so we need some juggling to cope
+	if result["auths"] != nil {
+		err = json.Unmarshal(result["auths"], &allAuthConfigs)
+	} else {
+		err = json.Unmarshal(byteValue, &allAuthConfigs)
+	}
+	if err != nil {
+		return fmt.Errorf("error unmarshalling file to authconfig %s: %v", secretFile, err)
+	}
+
+	if _, ok := allAuthConfigs[r.pullRepo]; ok {
+		r.PullAuth = authn.FromConfig(allAuthConfigs[r.pullRepo])
+		logger.Info("Found secret", "pull registry", r.pullRepo)
+	}
+	if _, ok := allAuthConfigs[r.pushRepo]; ok {
+		r.PushAuth = authn.FromConfig(allAuthConfigs[r.pushRepo])
+		logger.Info("Found secret", "push registry", r.pushRepo)
+	}
+	return nil
+}
 
 func checkArg(arg *string, varname string, fallback string) {
 	if *arg == "" {
@@ -50,41 +173,6 @@ func signFile(filename string, publickey string, privatekey string) error {
 	}
 	return nil
 }
-
-func getAuthFromFile(configfile string, repo string) (authn.Authenticator, error) {
-
-	if configfile == "" || configfile == "anonymous" {
-		logger.Info("no secret given, default to Anonymous")
-		return authn.Anonymous, nil
-	}
-
-	f, err := os.Open(configfile)
-	if err != nil {
-		return nil, err
-	}
-	defer f.Close()
-	cf, err := config.LoadFromReader(f)
-	if err != nil {
-		return nil, err
-	}
-
-	var cfg dockertypes.AuthConfig
-
-	cfg, err = cf.GetAuthConfig(repo)
-	if err != nil {
-		return nil, err
-	}
-
-	return authn.FromConfig(authn.AuthConfig{
-		Username:      cfg.Username,
-		Password:      cfg.Password,
-		Auth:          cfg.Auth,
-		IdentityToken: cfg.IdentityToken,
-		RegistryToken: cfg.RegistryToken,
-	}), nil
-
-}
-
 func die(exitval int, message string, err error) {
 	fmt.Fprintf(os.Stderr, "\n%s\n", message)
 	logger.Info("ERROR "+message, "err", err)
@@ -171,8 +259,7 @@ func main() {
 	var err error
 	var unsignedImageName string
 	var signedImageName string
-	var pullSecret string
-	var pushSecret string
+	var secretDir string
 	var extractionDir string
 	var filesList string
 	var privKeyFile string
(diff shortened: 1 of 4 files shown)


cmd/signimage/signimage.go
```diff
@@ -51,38 +51,88 @@ func signFile(filename string, publickey string, privatekey string) error {
...
-		return nil, err
+		return nil, fmt.Errorf("error reading secret file %s: %v", secretFile, err)
 	}
 
-	var cfg dockertypes.AuthConfig
+	var result map[string]json.RawMessage
+	var allAuthConfigs map[string]authn.AuthConfig
 
-	cfg, err = cf.GetAuthConfig(repo)
+	err = json.Unmarshal(byteValue, &result)
```
qbarrand, line 114: Wondering if it would make sense to instead feed `jsonFile` to a [`json.Decoder`](https://pkg.go.dev/encoding/json#Decoder). We potentially need to read it again, so we would need to `seek(0)`, but we still avoid reading the entire file in memory.

ybettan:
@chr15p Why do we have all that logic to parse the secret? `crane` have support for directly using a `dockerconfigjson` file for accessing images. See https://github.com/ybettan/kernel-module-management/blob/af8eb8630fcd64cfcd483e99770941773c45d39a/internal/registry/registry.go#L142-L151.

Am I missing something?

cmd/signimage/signimage.go
```diff
@@ -51,38 +51,88 @@ func signFile(filename string, publickey string, privatekey string) error {
 	return nil
 }
 
-func getAuthFromFile(configfile string, repo string) (authn.Authenticator, error) {
+// allSecretFiles  is a comma seperated list of files and directories to search for creds for repo
```
yevgeny-shnaidman, line 54: why the allSecretFiles is a list of files? It is the value of pullsecret input parameter, and it gets its value from the ImageRepo of the Module, which is a reference to a sing file

chr15p:
> @chr15p Why do we have all that logic to parse the secret? `crane` have support for directly using a `dockerconfigjson` file for accessing images. See https://github.com/ybettan/kernel-module-management/blob/af8eb8630fcd64cfcd483e99770941773c45d39a/internal/registry/registry.go#L142-L151.
> 
> Am I missing something?

You're missing a weird edge case interaction between k8s and crane used in a pod.

If you map a secret in as a volume the pod sees it as a directory named $SECRETNAME/ and then each secret key is a file within that ($SECRETNAME/$KEY).  We might need two secrets (one for pull, one for push, assuming we're pulling from one registry, and pushing to another), so two keys in the secret, so we get a directory with two files in it.

Crane's default keyring looks for credentials in a single file called config.json (the name is hardcoded) which is obviously a single file so if we used crane we could map one secret in, b ut not two, so we could pull, or push, but not both. 

Unless I am missing something obviously!    (it would make me happy if I was!)

(I could implement my own version of the keyring interface, but at that point I have to do pretty much the same thing anyway i.e. get an authn.Authenticator and wrap it in a keyring, rather than just get an Authenticator and use it directly,  so I don't see the benefit)

cmd/signimage/signimage.go
```diff
@@ -51,38 +51,88 @@ func signFile(filename string, publickey string, privatekey string) error {
...
+// so we need to search through them for the right one.
+// if we get an error we just log it and move on to the next file/dir i
+// and if we find nothing try anonymous auth
+// we're tolerant in our inputs and precise in our inputs. Its the UNIX way!
```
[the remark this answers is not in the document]
chr15p: tolerant in our inputs, precise in our outputs, and with incorrect documentation, it's even more the UNIX way! 

(Fixed, with embarrassment)

internal/sign/job/signer.go
```diff
@@ -122,10 +122,12 @@ func (m *signer) MakeJobTemplate(
...
-	if mod.Spec.ImageRepoSecret != nil {
-		args = append(args, "-pullsecret", "/docker_config/config.json")
-		volumes = append(volumes, utils.MakeSecretVolume(mod.Spec.ImageRepoSecret, v1.DockerConfigJsonKey, "config.json"))
-		volumeMounts = append(volumeMounts, utils.MakeSecretVolumeMount(mod.Spec.ImageRepoSecret, "/docker_config"))
+	imageSecret := mod.Spec.ImageRepoSecret
+	if imageSecret != nil {
+		args = append(args, "-pullsecret", "/docker_config/"+imageSecret.Name)
+		volumes = append(volumes, utils.MakeSecretVolume(imageSecret, "", ""))
+		volumeMounts = append(volumeMounts, utils.MakeSecretVolumeMount(imageSecret, "/docker_config/"+imageSecret.Name))
+
```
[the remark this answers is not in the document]
chr15p: done

cmd/signimage/signimage.go
```diff
@@ -3,14 +3,14 @@ package main
...
+	"encoding/json"
 	"flag"
 	"fmt"
-	"github.com/docker/cli/cli/config"
-	dockertypes "github.com/docker/cli/cli/config/types"
 	"github.com/go-logr/logr"
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/registry"
 	"io"
+	//"io/ioutil"
```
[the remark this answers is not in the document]
chr15p: done

cmd/signimage/signimage.go
chr15p: Because it allows us to pass in multiple secrets. :-)
The most obvious use of this will be in openshift to pass the secret associated with the builder service account alongside ImageRepoSecret ( e.g. ImageRepoSecret to pull from quay.io, the builder secret to push to the internal registry).  The controller code to pass that in would sit with the openshift kmm repo, not here, but allowing allSecretFiles to be a list here means that the signimage code will be common across both the k8s and ocp, and we'll have a single signing image rather than 2 subtly different ones.

Its also useful functionality to have if you want to use signimage within kubernetes but outside of KMM (e.g. an image build pipeline that just publishes to a registry without wanting to modprobe), so its worth supporting here, even if its not used by Kubernetes KMM.  (I also think we should allow ImageRepoSecret support multiple secrets, but that's an argument for another day)

cmd/signimage/signimage.go
```diff
@@ -51,38 +51,88 @@ func signFile(filename string, publickey string, privatekey string) error {
 	return nil
 }
 
-func getAuthFromFile(configfile string, repo string) (authn.Authenticator, error) {
+// allSecretFiles  is a comma seperated list of files and directories to search for creds for repo
+// so we need to search through them for the right one.
+// if we get an error we just log it and move on to the next file/dir i
+// and if we find nothing try anonymous auth
+// we're tolerant in our inputs and precise in our inputs. Its the UNIX way!
+func getAuthFromFileList(allSecretFiles string, repo string) authn.Authenticator {
```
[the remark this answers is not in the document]
chr15p: I looked at WalkDir and I think I rejected it because passing data too and from the WalkDirFunc function is a bit of a pain (iirc you can only do it using global variables) so it makes life easier in some ways and harder in others. 
Plus they are mounted secrets so there should only really be one level in the directory tree so the ability to walk down an arbitrarily deep dirtree isn't providing a huge benefit

cmd/signimage/signimage.go
chr15p: Would decoder not read the file into memory?  Or am I missing your point?

I'm opening and closing the file, then (potentially) opening and closing it again, which is slower then opening it once and seeking, but performance isn't really an issue here, (the whole signing job runs in a few seconds). I'm also quitting the search when I find the first authenticator that matches the repo whcih speeds things up but means we only loop through as many files as we need to, so on the second iteration we can't be sure if a file has been opened previously (so needed seek(0) ) and what hadn't (so needs open() ). 
It sounds more complicated and I'm not sure I understand the benefit.

Also the max size of a k8s secret is 1mb so thats all holding the file in memory could potentially need.

ybettan:
> You're missing a weird edge case interaction between k8s and crane used in a pod.

If you map a secret in as a volume the pod sees it as a directory named $SECRETNAME/ and then each secret key is a file within that ($SECRETNAME/$KEY). We might need two secrets (one for pull, one for push, assuming we're pulling from one registry, and pushing to another), so two keys in the secret, so we get a directory with two files in it.

Crane's default keyring looks for credentials in a single file called config.json (the name is hardcoded) which is obviously a single file so if we used crane we could map one secret in, b ut not two, so we could pull, or push, but not both.

Unless I am missing something obviously! (it would make me happy if I was!)

(I could implement my own version of the keyring interface, but at that point I have to do pretty much the same thing anyway i.e. get an authn.Authenticator and wrap it in a keyring, rather than just get an Authenticator and use it directly, so I don't see the benefit)

What happens if you use the `spec.imagePullSecrets` field instead of `volumtes`?
See https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-pod-that-uses-your-secret.

chr15p:
I have rewritten the directory walking code based around Walkdir, I'm not sure its simpler but it is more idiomatic (or go-ey!), and avoids the need for global variables by running it inside a class.  It also means we only walk the directory once which is nicer.

It does mean all secrets need to be mounted in the same directory and we pass in a single "-secretsdir" rather than separate "-pullsecret" and "-pushsecret", which is maybe not quite as nice but (I think) worth it.  In OCP we will then need to mount the builder sa account secrets into the same dir (but thats a change for midstream not here)

cmd/signimage/signimage.go
```diff
@@ -3,21 +3,141 @@ package main
...
+		// we could return the error and die, but we want to be tolerant of secrets in the wrong format etc
+		return
+	}
+}
+
+func (r *repoAuth) authDirFileHandler(path string, d fs.DirEntry, err error) error {
+	if path == r.configDir {
+		return nil
+	}
+	if strings.Contains(path, "..") {
```
qbarrand, line 75: Do we need special handling for that case? I don't think `Walk` will go back to any parent dir.
What do you think about skipping all directories anyway?

cmd/signimage/signimage.go
qbarrand: > Would decoder not read the file into memory?

The marshalled contents would, but not the bytes themselves - now, I believe we have both.
However I agree that the suggestion is probably premature optimization (if it optimizes anything at all) and doesn't deserve more time now.

cmd/signimage/signimage.go
```diff
@@ -3,21 +3,141 @@ package main
...
 	"k8s.io/klog/v2/klogr"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
 )
 
+// configDir should contain all the secrets available as individual files named for their keys
+// so we need to search through for the secrets we need for our pull and push repos.
+// and if we find nothingi default anonymous auth
```
qbarrand, line 23:
```suggestion
// If we do not find any authn, default to anonymous.
```

cmd/signimage/signimage.go
```diff
@@ -3,21 +3,141 @@ package main
...
+		logger.Info("no secrets for push found, default to Anonymous")
+	}
+	return r
+}
+
+func (r *repoAuth) PopulateAuthFromFileList() {
+	// not giving any secrets should short cuircuit the whole process
+	if r.configDir == "" {
+		return
+	}
```
qbarrand, line 60: How about setting the anonymous authn as default? Then we can remove some ifs in the calling function.
```suggestion
	r.PullAuth = authn.Anonymous
	r.PushAuth = authn.Anonymous

	// not giving any secrets should short cuircuit the whole process
	if r.configDir == "" {
		return
	}
```

cmd/signimage/signimage.go
chr15p: mounting a secret as a volume doesn't just give you the keys but also gives you some extra directories called things like "..data" , they all have ".." in their names and don't contain anything we care about so that test is to make sure we skip over them.

(its Contains() rather than HasPrefix() or a slice because path includes the full directory path so "mydir/..data" )

cmd/signimage/signimage.go
chr15p: because I wanted to be able to log that its using anonymous auth, so when it fails to push or pull you it gives you a clue that its because you didn't give it the right secrets. It might be that's what you want so its a message not an error, but the user needs to know so we would still need the if its just a question of `if nil` or `if authn.Anonymous` and the former feels more like its catching an error to me  

signimage is deliberately quite aggressive at logging what its doing so when it goes wrong its (hopefully) obvious what needs fixing

cmd/signimage/signimage.go
```diff
@@ -3,21 +3,141 @@ package main
...
+		return nil
+	}
+	if d.IsDir() {
+		return nil
+	}
+	if err != nil {
+		logger.Info("error walking tree", "filename", path, "error", err)
+		//theres an derror in WalkDir not much we can do except skip and move on
+		return nil
+	}
```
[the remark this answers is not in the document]
chr15p: err in this case is coming from Walkdir() and implies either fs.Stat() or ReadDir() has failed.  The only way I can think of that we could get these is if there is something wrong with the volumemount and that feels like a k8s problem not something the user can fix,   so I think "skip and moving on" is the only thing we can do, erroring out the program just leaves an issue the user cant fix.

If we carry on we either find the secrets in a directory that doesn't error and it doesn't matter or we don't, we try `authn.anonymous`, and we've made our best effort to run.

I'm not wedded to that opinion though if you have a better idea.

cmd/signimage/signimage.go
chr15p: vim needs a "random 'i' detected, have you forgotten you're already in insert mode?"  warning :)
Changed.

cmd/signimage/signimage.go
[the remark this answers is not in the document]
chr15p: I've moved this to the top and fixed the comment's spelling...

ybettan:
After talking to @chr15p this is my understanding:
* The sign job is using 2 pieces: the `spec.imageRepoSecret` and the `build.Secret` of the module.
* KMM is starting the sign job and mounting the `spec.ImageRepoSecret` and `build.Secret` to the job.
* the sign job is then combining all those files (secrets keys) into a single dockerconfigjson (or other format) and using it.

I would like to understand if it is possible and if it is, change it in the future (to not hold this PR) to:
* KMM will start the sign job and will mount the `build.secret` but instead of mounting `spec.imageRepoSecret` it will pass it as a runtime arg to the sign binary (the secret name only) and the cluster context if needed.
* The sign code will then get the keyChain based on the secret same way we are doing in KMM and pass it to crane as an option. 
This will probably remove that whole parsing code no?!

WDYT? @chr15p @qbarrand @yevgeny-shnaidman

chr15p:
> WDYT? @chr15p 

The standard answer to "how do I access a secret within a job?"  is "mount it as a volume" and thats simple so that's how this started out, however things have evolved and its now got quite complicated (two formats for pull secrets doesn't help!). 
To just pass in the secret name the signimage.go binary would need to act as a k8s client and so need the settings and creds to do that which has its own issues, but definitely worth a look.  But thats all for a future PR!

PR #218, opened 2022-12-20
Build two bundles

qbarrand:
Build one bundle for the standard edition of KMM and another one for the Hub version.

Hub edition owns the `ManagedClusterModule` CRD and is requires the `OPERATOR_NAMESPACE` environment variable.

Standard edition owns the `Module` and `PreflightValidation` CRDs. When started with `KMM_MANAGED=1`, the `NodeKernelClusterClaim` reconciler is started ("spoke mode").

Move the `ManagedClusterModule` CRD to the `hub.kmm.sigs.x-k8s.io` API sub-group to avoid having it included in the standard bundle.
Split `main.go` into two entrypoints under `cmd`.
Rework the `config` kustomize layout to accomodate two bundles.
Add all new images to `cloudbuild.yaml`.


Makefile
```diff
@@ -127,17 +134,19 @@ lint: golangci-lint ## Run golangci-lint against code.
...
+	go build -o $@ ./cmd/manager-hub
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./main.go
 
 .PHONY: docker-build
-docker-build: unit-test ## Build docker image with the manager.
-	docker build -t $(IMG) .
+docker-build: ## Build docker image with the manager.
```
yevgeny-shnaidman, line 148: why we no longer want to run unit-test for docker-build

Makefile
```diff
@@ -127,17 +134,19 @@ lint: golangci-lint ## Run golangci-lint against code.
...
+	docker build -t $(IMG) --build-arg TARGET=manager .
```
yevgeny-shnaidman, line 149: is there a target for building the manager-hub image?

cmd/manager/main.go
```diff
@@ -37,173 +52,156 @@ import (
...
-	setupLogger := logger.WithName("setup")
-
-	commit, err := gitCommit()
+	commit, err := cmd.GitCommit()
 	if err != nil {
 		setupLogger.Error(err, "Could not get the git commit; using <undefined>")
 		commit = "<undefined>"
 	}
 
+	managed, err := GetBoolEnv("KMM_MANAGED")
```
yevgeny-shnaidman, line 87: Can we set this flag into the ConfigMap, somehow?

internal/cmd/cmdutils.go
```diff
@@ -0,0 +1,32 @@
+package cmd
```
yevgeny-shnaidman, line 1: we already have utils package, maybe this file can be moved there?

Makefile
qbarrand: It slowed my development workflow and did not bring any benefit. Do you insist on keeping it?

Makefile
qbarrand: Not yet, I did not want to add too many targets. Do you think there should be one?

cmd/manager/main.go
qbarrand: This is much easier to override in the [`SubscriptionConfig`](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/subscription-config.md). If we use the `ConfigMap,` we need the admin to first createit with the appropriate value and then to replace the `volumeMount` entry appropriately. This looked more complex to me for now real benefit.
BTW the custom type for settings is gone in this PR.

internal/cmd/cmdutils.go
qbarrand: [Like Dave Cheney](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common), I think the package names should be as precise as possible. `utils` might contain anything - `cmd` will contain CLI-related stuff.

Makefile
yevgeny-shnaidman: i think it is a good benchguard. I don't mind that it will be moved to a different target, as long as that target will be called during the image build by CI.

Makefile
yevgeny-shnaidman: i think yes. We are all using docker-build target to build images, so it is better to have one for hub also. Maybe define 2 targets:
docker-build-manager
docker-build-hub

Makefile
qbarrand: Running tests is quite different from building the image though. It involves downloading `controller-gen` because we may need to regenerate mocks.
Neither upstream nor midstream use the target to build images for that reason. But the unit tests are ran in a different job.

Makefile
qbarrand: PTAL

Makefile
ybettan: +1 for removing it.

UT are tested in CI regardless of building the image so we have no risk here.

Makefile
```diff
@@ -127,17 +134,23 @@ lint: golangci-lint ## Run golangci-lint against code.
...
+
+.PHONY: docker-build-hub
+docker-build-hub: ## Build docker image with the hub manager.
+	docker build -t $(IMG) --build-arg TARGET=manager-hub .
```
ybettan, line 153: I don't think `IMG` should be use here. Don't we want a different image name for the hub?

Makefile
```diff
@@ -215,11 +228,24 @@ operator-sdk:
...
+	rm -fr ./bundle
+	${OPERATOR_SDK} generate kustomize manifests --apis-dir api
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	kubectl kustomize config/manifests | ${OPERATOR_SDK} generate bundle $(BUNDLE_GEN_FLAGS)
 	${OPERATOR_SDK} bundle validate ./bundle
 
+.PHONY: bundle-hub
+bundle-hub: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
+	rm -fr ./bundle
+	operator-sdk generate kustomize manifests \
```
ybettan, line 240:
```suggestion
	${OPERATOR_SDK} generate kustomize manifests \
```

Makefile
```diff
@@ -215,11 +228,24 @@ operator-sdk:
...
 bundle: operator-sdk manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
-	${OPERATOR_SDK} generate kustomize manifests -q
+	rm -fr ./bundle
+	${OPERATOR_SDK} generate kustomize manifests --apis-dir api
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
 	kubectl kustomize config/manifests | ${OPERATOR_SDK} generate bundle $(BUNDLE_GEN_FLAGS)
 	${OPERATOR_SDK} bundle validate ./bundle
 
+.PHONY: bundle-hub
+bundle-hub: manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
```
ybettan, line 238:
```suggestion
bundle-hub: operator-sdk manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
```

Makefile
```diff
@@ -215,11 +228,24 @@ operator-sdk:
...
+	rm -fr ./bundle
+	operator-sdk generate kustomize manifests \
+		--apis-dir api-hub \
+		--output-dir config/manifests-hub \
+		--package kernel-module-management-hub \
+		--input-dir config/manifests-hub
+	cd config/manager-hub && $(KUSTOMIZE) edit set image controller=$(HUB_IMG)
+	kubectl kustomize config/manifests-hub | operator-sdk generate bundle --package kernel-module-management-hub $(BUNDLE_GEN_FLAGS)
```
ybettan, line 246:
```suggestion
	kubectl kustomize config/manifests-hub | ${OPERATOR_SDK} generate bundle --package kernel-module-management-hub $(BUNDLE_GEN_FLAGS)
```

Makefile
```diff
@@ -215,11 +228,24 @@ operator-sdk:
...
+	operator-sdk bundle validate ./bundle
```
ybettan, line 247:
```suggestion
bundle-hub: operator-sdk manifests kustomize ## Generate bundle manifests and metadata, then validate generated files.
	rm -fr ./bundle
	${OPERATOR_DK} generate kustomize manifests \
		--apis-dir api-hub \
		--output-dir config/manifests-hub \
		--package kernel-module-management-hub \
		--input-dir config/manifests-hub
	cd config/manager-hub && $(KUSTOMIZE) edit set image controller=$(HUB_IMG)
	kubectl kustomize config/manifests-hub | ${OPERATOR_DK} generate bundle --package kernel-module-management-hub $(BUNDLE_GEN_FLAGS)
	${OPERATOR_DK} bundle validate ./bundle
```

qbarrand: Fixed, thanks!

Makefile
qbarrand: Fixed, thanks!

ybettan:
/hold
@yevgeny-shnaidman Can unhold when he is happy

PR #219, opened 2022-12-21
Installing `operator-skd` via Makefile.

ybettan:
To make sure u/s and m/s are always using the same version without the need of manually installing and maintaining `operator-sdk` with the same version on u/s infra and on m/s infra.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -208,12 +208,21 @@ GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
 }
 endef
 
+OPERATOR_SDK = $(shell pwd)/bin/operator-sdk
+.PHONY: operator-sdk
+operator-sdk:
+	@if [ ! -f ${OPERATOR_SDK} ]; then \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -208,12 +208,18 @@ GOBIN=$(PROJECT_DIR)/bin go install $(2) ;\
 }
 endef
 
+OPERATOR_SDK = $(shell pwd)/bin/operator-sdk
+.PHONY: operator-sdk
+operator-sdk:
+	curl -Lo ${OPERATOR_SDK} 'https://github.com/operator-framework/operator-sdk/releases/download/v1.25.2/operator-sdk_linux_amd64'
+	chmod +x ${OPERATOR_SDK}
```
qbarrand, line 215: This makes it always download the binary, doesn't it? It should skip the download if the binary is already in `bin`.

ybettan: Thanks. Fixed.

PR #220, opened 2022-12-22
Making the e2e-incluster-build job to actually build in cluster.

ybettan:
Until now, we were just building the driver container in the Github
runner, loading it into minikube and then the Dockerfile loaded as a
ConfigMap was just downloading that image, adding some files and pushing
instead of actually build the driver container in cluster.

This commit is fixing that flow.

### DIFF
--- a/ci/prow/e2e-incluster-build
+++ b/ci/prow/e2e-incluster-build
@@ -5,38 +5,31 @@ set -euxo pipefail
 echo "Deploy KMMO..."
 make deploy
 
-echo "Wait until the KMMO Deployment is Available"
-timeout 1m kubectl wait --for condition=Available deployments.apps -n kmm-operator-system kmm-operator-controller-manager 
-
 echo "Create a build secret..."
 kubectl create secret generic build-secret --from-literal=ci-build-secret=super-secret-value
+
 echo "Add a configmap that contain the kernel module build dockerfile..."
 kubectl apply -f ci/kmm-kmod-dockerfile.yaml
 
 echo "Add an kmm-ci Module that contains a valid mapping..."
-sed -e "s/KVER_CHANGEME/$(uname -r)/g" ci/module-kmm-ci-build.template.yaml | tee module-kmm-ci.yaml
+sed -e "s/KVER_CHANGEME/$(minikube ssh -- uname -r)/g" ci/module-kmm-ci-build.template.yaml | tee module-kmm-ci.yaml
 kubectl apply -f module-kmm-ci.yaml
 
-echo "Wait for the job to be created..."
-timeout 1m bash -c 'until kubectl get job -l kmm.node.kubernetes.io/module.name | grep kmm; do sleep 1; done'
+# Wait for the build pod to be created. `kubectl wait` doesn't support such option,
+# see https://github.com/kubernetes/kubernetes/issues/83242.
+echo "Waiting for the build pod to be created..."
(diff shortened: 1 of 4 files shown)


ybettan:
@qbarrand @yevgeny-shnaidman @enriquebelarte 
Not completely done yet but it is ready for an early review.

Something that still needs to be done is to make the `pre-build module` job to run after the `in-cluster-build` one and use the image that already exists in the `minikube` registry.

ybettan:
https://github.com/kubernetes-sigs/kernel-module-management/pull/220#issuecomment-1369493525

Usage of that job in other job will be done in future PRs.

enriquebelarte (approved):

ci/prow/e2e-incluster-build
```diff
@@ -5,38 +5,34 @@ set -euxo pipefail
...
-timeout 1m bash -c 'until kubectl get job -l kmm.node.kubernetes.io/module.name | grep kmm; do sleep 1; done'
+# Wait for the build pod to be created. `kubectl wait` doesn't support such option,
+# see https://github.com/kubernetes/kubernetes/issues/83242.
+echo "Waiting for the build pod to be created..."
+timeout 1m bash -c 'until kubectl get pods -o json | jq -er ".items[].metadata.name | select(.? | match(\"build\"))"; do sleep 1; done'
+POD_NAME=$(kubectl get pods -o json | jq -r '.items[].metadata.name | select(.? | match("build"))')
 
-echo "Check that the module gets loaded on the node..."
-timeout 1m bash -c 'until minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+echo "Check that the build secret is mounted to the build pod"
```
yevgeny-shnaidman, line 24: why do we need this check?

ci/prow/e2e-incluster-build
```diff
@@ -5,38 +5,34 @@ set -euxo pipefail
...
 
-echo "Check that the DriverContainer prints the secret's value to the standard output..."
-POD_NAME=$(kubectl get pod -l kmm.node.kubernetes.io/module.name --template='{{ (index .items 0).metadata.name }}')
-KERNEL_VERSION=$(uname -r)
-echo "Looking for the build secret"
-timeout 1m bash -c "until kubectl exec $POD_NAME -- grep super-secret-value /ci-build-secret; do sleep 3; done"
+# The build job/pod is deleted once done so we won't be able to get this info later on in the troubleshooting section.
+echo "Print the build logs..."
+# we can't get the logs on a pod that isn't `Running` yet.
+kubectl wait pod/${POD_NAME} --for jsonpath='{.status.phase}'=Running --timeout=60s
```
yevgeny-shnaidman, line 30: isn't it a little bit problematic? We can fail it, since build pod may have been deleted before we check

ci/prow/e2e-incluster-build
ybettan: This was already in the test but basically, when we have a `build` section in the CR, we may have a `build` secret as well.

The test is also making sure the build secret is correctly mounted to the build job/pod.

ci/prow/e2e-incluster-build
ybettan: The build is pretty long because it is downloading the kernel-headers.

ci/prow/e2e-incluster-build
ybettan: After chatting, the job is not really using that build secret in the CR, it would use it if the `minikube`  registry was protected by authZ but it isn't.

There is no reason to check if a secret was mounted in a flow in which we are not using that secret (this is more of a UT flow).

Removing that check.

PR #222, opened 2023-01-03
Fixing CI to use the KMM image build from the PR.

ybettan:
ATM, we are deploying the default image in all our e2e tests and completely ignoring the image build with the src code of a specific PR.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -4,6 +4,7 @@ on: [pull_request]
 
 env:
   GO_VERSION: 1.18
+  IMG: kmm:local
 
 jobs:
   build-operator-image:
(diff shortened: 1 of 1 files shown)


ybettan:
https://github.com/kubernetes-sigs/kernel-module-management/pull/222#issuecomment-1369665990

/unhold
My understanding is that we will always pull but from the minikube registry so as long as we load the KMM image there we are fine.

ybettan:
FYI @enriquebelarte

ybettan:
I was missing the overlay in https://github.com/kubernetes-sigs/kernel-module-management/blob/81a48e22997128ab33e35f07595bbafbd4a47702/.github/workflows/e2e.yaml#L82.

Closing the PR
/close.

PR #224, opened 2023-01-04
Making the pre-built e2e test to use the in-cluster-build output.

ybettan:
Instead of building the driver-container on the github runner and then
pulling it in the flow, we will now use the `in-cluster-build` flow to
build the driver container in-cluster and then pull it in the next step
with a new CR.

This detach the running cluster from its backend, for example running
any other cluster than a cluster using the host kernel (such as
`minikube` with the `docker` driver or `kind`) and allow us to re-use
those tests in other cluster envs.

The following jobs are removed:

* `e2e-two-nodes`
* `e2e-crd-variable`

The reasons they were removed are:

1. It is hard to have different jobs using the output of the
in-cluster-build job's output because those jobs won't have access to
the same `minikube` cluster and its registry (building on the runner is
problematic when trying to reproduce on different clusters). The other
option is to append it to the same job as another step but then we are
just making the CI running for longer period of time (x2 time) and not
gaining much coverage (see section 2).

2. Those jobs are not testing anything special, one is just testing the
selector of the `ds` and the other the regex in the `kernelMapping` which
IMO should be both tested in UT rather than having their own e2e
workflows.

---

####################################################
**This PR is removing some previous work so it is open for discussion**
####################################################


ybettan:
/uncc @qbarrand 
/uncc @yevgeny-shnaidman 
No need to spam you with notifications at this point.

ybettan:
/cc @qbarrand 
/cc @yevgeny-shnaidman 
FYI @enriquebelarte

ybettan:
/override Container image CRD variable
/override In-cluster build
/override Prebuilt kernel module
/override Prebuilt kernel module - two nodes
Those jobs are still enforced by the prow branch protection but the tests doesn't exist anymore. They will be removed from the branch protection in this PR: https://github.com/kubernetes/test-infra/pull/28350 later on.

ybettan:
After talking with @qbarrand we agreed that we will only keep one job which will be the `in-cluster-build` job which tests both the build and the loading of a now "pre-built" driver container.

Later on, when we have a `signing` job, this job will apply a module which will contain a `sign` section as well as a `build` section so we don't have to deal with pre-built driver containers.

PR #225, opened 2023-01-04
changing the short name of the preflightvalidation CRD

yevgeny-shnaidman:


ybettan:
Much better than the previous confusing name :)
/lgtm

PR #226, opened 2023-01-05
Configure related images via environment variables

qbarrand:
This allows us to patch those variables with the digest equivalent of the images.
Unfortunately `operator-sdk` does not seem to identify them properly for an unknown reason; that means that the images are not listed under the CSV's `relatedImages` field, nor are they automatically replaced with their digest equivalent.


qbarrand:
FYI @chr15p

PR #229, opened 2023-01-09
Adding preflight documentation

yevgeny-shnaidman:


ybettan:
I think it worth adding a `Preflight` CR example to the doc.

docs/mkdocs/preflight/preflight-validation.md
```diff
@@ -0,0 +1,100 @@
...
+Before executing upgrade on the cluster with applied KMM Modules, admin needs to verify that installed kernel modules (via KMM)
+will be able to be installed on the nodes after the cluster upgrade and possible kernel upgrade. 
+Preflight will try to validate every Module loaded in the cluster, in parallel (it does not wait for validation of one Module to complete, before starting
+validation of another Module).
+
+### Validation kick-off
+Preflight validation is triggered by uploading PreflightValidation CR into the cluster. This CR's Spec contains 2 fields:
+```
+type PreflightValidationSpec struct {
+        // KernelImage describes the kernel image that all Modules need to be checked against.
```
ybettan, line 12: Nit: We should probably update the doc to fit the `KernelVersion` instead of `KernelImage` in a later PR.

docs/mkdocs/preflight/preflight-validation.md
```diff
@@ -0,0 +1,100 @@
...
+        // +kubebuilder:validation:Required
+        KernelVersion string `json:"kernelVersion"`
+
+        // Boolean flag that determines whether images build during preflight must also
+        // be pushed to a defined repository
+        // +optional
+        PushBuiltImage bool `json:"pushBuiltImage"`
+}
```
ybettan, line 20:
```suggestion
  spec:
    kernelVersion: 6.0.15 # the version of the kernel that the cluster will be upgraded to. Manadatory field
    pushBuiltImage: true # if true, then the images created during the Build and Sign validation will be pushed to their repositories (false by default).
```

WDYT?

docs/mkdocs/preflight/preflight-validation.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+1. KernelVersion - the version of the kernel that the cluster will be upgraded to. Manadatory field
+2. PushBuiltImage - if true, then the images created during the Build and Sign validation will be pushed to their repositories (false by default).
+
+### Validation lifecycle
+Preflight validation will try to validation every module loaded in the cluster. Preflight will stop running validation on a module, once its validation
```
ybettan, line 27:
```suggestion
Preflight validation will try to validate every module loaded in the cluster. Preflight will stop running validation on a module, once its validation
```

docs/mkdocs/preflight/preflight-validation.md
```diff
@@ -0,0 +1,100 @@
...
+2  verify the presence of the kernel module defined in the Module in the correct path for future modprobe execution. If this validation is successfull
+   it probably means that the kernel module was compiled with the correct linux headers. The correct path is:
+   <DirName>/lib/modules/<UpgradedKernel>/  
+
+### Build validation stage
+
+Build validation is executed only in case image validation has failed, and there is a Build section in the Module that is relevent for the
+upgraded kernel. Build validation will try to run build job and validate that it finishes successfully. In case "Push" flag is defined in the Preflight CR,
+it will also try to push the resulting image into its repo. The resulting image is take from the definition of the ContainerImage field of the Module CR
+Note: if Sign section is defined for the upgraded kernel, then the resulting image will not be ContainerImage field of the Module CR, but a temporary image name,
```
ybettan, line 96: How does `then the resulting image will not be ContainerImage field of the Module CR` goes with...

docs/mkdocs/preflight/preflight-validation.md
```diff
@@ -0,0 +1,100 @@
...
+since the resulting image should be the product of Sign flow
+
+### Sign validation stage
+Sign validation is executed only in case image validation has failed, there is a Sign section in the Module that is relevent for the upgrade kernel, and build validation finished successfully in case there was a Build section in the Module relevent for the upgraded kernel. Sign validation will try to run sign job and validate that it finishes successfully. In case "Push" flag is defined in the Preflight CR, it will also try to push the resulting image into its repo.The result image is always the image defined in the ContainerImage field of the Module. The input image is either the output of the Build stage, or an image defined in the UnsignedImage field
```
ybettan, line 100: ...That `The result image is always the image defined in the ContainerImage field of the Module`?

ybettan, line 100: I will add a full CR example at the end/beginning of the doc.

docs/mkdocs/preflight/preflight-validation.md
yevgeny-shnaidman: fixed

docs/mkdocs/preflight/preflight-validation.md
yevgeny-shnaidman: in case of a build, when the sign section is defined, the build output will be some temporary image name, and not the final one. The final one can only be the result of a Sign success.

docs/mkdocs/preflight/preflight-validation.md
yevgeny-shnaidman: see the answer above

yevgeny-shnaidman: added example at the end of the file

docs/mkdocs/preflight/preflight-validation.md
ybettan: Thanks for the explanation. Make sense.

PR #230, opened 2023-01-09
Implement ManagedClusterModuleStatus for KMM-Hub

mresvanis:
This PR adds the implementation of the `ManagedClusterModuleStatus`. The latter reflects the number of desired, applied and degraded `ManifestWork` CRs owned by the respective `ManagedClusterModule` CR.

### DIFF
--- a/api-hub/v1beta1/managedclustermodule_types.go
+++ b/api-hub/v1beta1/managedclustermodule_types.go
@@ -38,6 +38,14 @@ type ManagedClusterModuleSpec struct {
 
 // ManagedClusterModuleStatus defines the observed state of ManagedClusterModule.
 type ManagedClusterModuleStatus struct {
+	// Number of ManifestWorks to be applied.
+	NumberDesired int32 `json:"numberDesired"`
+
+	// Number of ManifestWorks that have been successfully applied.
+	NumberApplied int32 `json:"numberApplied"`
+
+	// Number of ManifestWorks that could not be successfully applied.
+	NumberDegraded int32 `json:"numberDegraded"`
 }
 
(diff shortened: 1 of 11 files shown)


internal/statusupdater/statusupdater.go
```diff
@@ -84,6 +104,33 @@ func (m *moduleStatusUpdater) ModuleUpdateStatus(ctx context.Context,
 	return m.client.Status().Update(ctx, mod)
 }
 
+func (m *managedClusterModuleStatusUpdater) ManagedClusterModuleUpdateStatus(ctx context.Context,
+	mcm *hubv1beta1.ManagedClusterModule,
+	ownedManifestWorks []workv1.ManifestWork) error {
+
+	var numApplied int32
+	var numDegraded int32
```
ybettan, line 112: Why don't we just use `int` here?

api-hub/v1beta1/managedclustermodule_types.go
```diff
@@ -38,6 +38,14 @@ type ManagedClusterModuleSpec struct {
 
 // ManagedClusterModuleStatus defines the observed state of ManagedClusterModule.
 type ManagedClusterModuleStatus struct {
+	// Number of ManifestWorks to be applied.
+	NumberDesired int32 `json:"numberDesired"`
```
ybettan, line 42: Why does it needs to be `int32` and not just `int`?

qbarrand: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

PR #232, opened 2023-01-10
Adding back the `build-secret` to the e2e test.

ybettan:
The decision to remove it was a mistake. It was confused with `imageRepoSecret`.

The build arg is supposed to be available to the `build` Dockerfile at `/run/secrets/build-secret/filename`.


ybettan:
FYI @enriquebelarte

PR #234, opened 2023-01-11
Adjust ManagedClusterModuleReconciler ManagedCluster watch predicate

mresvanis:
This change updates the `ManagedClusterModule` reconciler's `ManagedCluster` watch predicate, in order to enqueue update events of changed KMM `ClusterClaim`s. This way whenever a new kernel version is available on a `ManagedCluster`, the `ManagedClusterModule` reconciler will reconcile the affected `ManagedClusterModule`s.

### DIFF
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -2,19 +2,23 @@ package filter
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/go-logr/logr"
-	hubv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api-hub/v1beta1"
-	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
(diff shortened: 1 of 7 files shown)


internal/filter/filter.go
```diff
@@ -27,6 +31,34 @@ var skipDeletions predicate.Predicate = predicate.Funcs{
 	DeleteFunc: func(_ event.DeleteEvent) bool { return false },
 }
 
+var kmmClusterClaimChanged predicate.Predicate = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		if e.ObjectOld == nil {
```
yevgeny-shnaidman, line 36: We don't need to check if the New and Old objects are nil, in the update Event they cannot be nil, i think 
we need to check that the casting is OK

mresvanis: Thank you for the comment! Updated accordingly.

PR #236, opened 2023-01-12
Updating the `in-cluster-build` job to correctly use build secrets.

ybettan:
Instead of checking that the secrets are available on the `kaniko` pod we should check that the secret files are available in the Dockerfile itself.

### DIFF
--- a/ci/kmm-kmod-dockerfile.yaml
+++ b/ci/kmm-kmod-dockerfile.yaml
@@ -17,6 +17,8 @@ data:
 
     WORKDIR /usr/src
 
+    RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret
+
     RUN git clone https://github.com/kubernetes-sigs/kernel-module-management.git
 
     WORKDIR /usr/src/kernel-module-management/ci/kmm-kmod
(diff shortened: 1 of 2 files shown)


ci/kmm-kmod-dockerfile.yaml
```diff
@@ -17,7 +17,8 @@ data:
 
     WORKDIR /usr/src
 
-    RUN git clone https://github.com/kubernetes-sigs/kernel-module-management.git
+    RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret && \
+        git clone https://github.com/kubernetes-sigs/kernel-module-management.git
```
qbarrand, line 21:
```suggestion
    RUN grep super-secret-value /run/secrets/build-secret/ci-build-secret
    RUN git clone https://github.com/kubernetes-sigs/kernel-module-management.git
```

ybettan: I don't mind by why? I would like to think about it as `git clone --auth ...` so it makes sense to me to put in in the same layer.

ybettan: Fixed.

PR #250, opened 2023-01-18
Configuring the build and sign job backoff and restart policies

yevgeny-shnaidman:
In the default configuration, once the failed container of the pod is restarted 6 times, and after the final failure it is deleted. This PR changes the policy to never restart, and the backoff limit to 0, meaning that after first failure, the pod/container won't be restarted, but the failed pod itself will be kept in the cluster for future debugging

### DIFF
--- a/internal/build/job/maker.go
+++ b/internal/build/job/maker.go
@@ -105,8 +105,9 @@ func (m *maker) MakeJobTemplate(
 			Annotations:  map[string]string{constants.JobHashAnnotation: fmt.Sprintf("%d", specTemplateHash)},
 		},
 		Spec: batchv1.JobSpec{
-			Completions: pointer.Int32(1),
-			Template:    specTemplate,
+			Completions:  pointer.Int32(1),
+			BackoffLimit: pointer.Int32(0),
+			Template:     specTemplate,
 		},
 	}
 
@@ -146,7 +147,7 @@ func (m *maker) specTemplate(
 				},
 			},
 			NodeSelector:  modSpec.Selector,
-			RestartPolicy: v1.RestartPolicyOnFailure,
+			RestartPolicy: v1.RestartPolicyNever,
 			Volumes:       volumes(modSpec, buildConfig),
 		},
 	}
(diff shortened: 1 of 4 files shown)


internal/build/job/maker.go
```diff
@@ -105,8 +105,9 @@ func (m *maker) MakeJobTemplate(
 			Annotations:  map[string]string{constants.JobHashAnnotation: fmt.Sprintf("%d", specTemplateHash)},
 		},
 		Spec: batchv1.JobSpec{
-			Completions: pointer.Int32(1),
-			Template:    specTemplate,
+			Completions:  pointer.Int32(1),
+			BackoffLimit: pointer.Int32(0),
```
qbarrand, line 109: Not sure about this. Signing jobs only depend on a few external components (key and cert secrets) but build jobs potentially clone code from the internet, which is a lot more prone to transient failures.

yevgeny-shnaidman: most of the transient errors will be due to networking issue for accessing the source code and , maybe, packages download. So, on one hand, if we don't change the backoff, it might resolve itself on subsequent re-runs, but if the issue is more serious, then user might not know what exactly failed

yevgeny-shnaidman: and, in case it failed on networking issue, user can see it and restart

ybettan:
I am happy if @qbarrand is.

internal/build/job/maker.go
yevgeny-shnaidman: if i understand correctly, the job will be clean up after timeout, regardless of it being successful or not, which is not what we want. We want the failed job to stay and the failed job's pod to stay, until user decides that he does not need them anymore

qbarrand: Isn't it ideal to keep the Job for a specified duration, e.g. 2 hours? We keep the retries in case that was something transient, and at the same time the user has an opportunity to must-gather with good troubleshooting data.

yevgeny-shnaidman: but TTL is per job, it is for job deletion. Without the PR, the failed pods will still get deleted, so keeping the job for 2 hours won't matter in that case

PR #255, opened 2023-01-19
Initial hub-spoke documentation

iranzo:
WIP for HUB-Spoke docs

Some issues with formatting as described in #257

### DIFF
--- a/docs/mkdocs/integration/hub-spoke.md
+++ b/docs/mkdocs/integration/hub-spoke.md
@@ -1,0 +1,363 @@
+# Hub-Spoke deployment
+
+Using [Open Cluster Management](https://open-cluster-management.io/), allows to configure and define policies in our control cluster, named `hub` that later can be applied to satellite clusters named `spoke`.
+
+We can leverage OCM to deploy KMM on the environment and have the different components orchestrated by OCM.
+
+!!! note "Using podman instead of docker"
+
+    If you're using podman like we'll do, we need to symlink the command to the `docker` name, as the `Makefile` does uses docker.
+
+    `ln -s /usr/bin/podman /usr/bin/docker`
+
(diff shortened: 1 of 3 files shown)


docs/mkdocs/integration/hub-spoke.md
```diff
@@ -0,0 +1,363 @@
...
+
+- `registry.ci.openshift.org` token - [Documentation](https://docs.ci.openshift.org/docs/how-tos/use-registries-in-build-farm/#how-do-i-log-in-to-pull-images-that-require-authentication)
+
+!!! warning "`quay.io` account"
+
+    if the repositories are not created beforehand, please make sure to convert them to public ones after pushing to them
+
+## Setup the Hub
+
+1. Build and push the KMM-Hub container image, the KMM-Hub OLM bundle and catalog from the midstream repository:
```
ybettan, line 25: Can't we just consume the operator directly from operatorhub.io? Why does a customer will need to build it from source?

docs/mkdocs/integration/hub-spoke.md
```diff
@@ -0,0 +1,363 @@
...
+
+5. Create a namespace on the Hub, in which the Build and Sign jobs will be created:
+
+```bash
+kubectl create ns kmm-jobs-test
+```
+
+## Setup the Spoke
+
+1. Build and push the KMM container image, the KMM OLM bundle and catalog from midstream repository:
```
ybettan, line 183: Same question. We have official builds - why not just use them? https://operatorhub.io/operator/kernel-module-management.

I also see now that we don't have any public subscription for kmm-hub. Should we?

iranzo: This was written before the docs were published and before 1.0 release, probably doesn't make sense anymore

PR #256, opened 2023-01-19
Workaround for SF.net plantuml URL missing

iranzo:
In my test of the docs pr, plantuml was failing to download and netlify complained of corrupt JAR file

In the website the version we were using has disappeared, and the reported 2023 one is not listed, using 2021 in the interim until it's fixed

### DIFF
--- a/docs.mk
+++ b/docs.mk
@@ -5,4 +5,5 @@ docs: bin/plantuml.jar
 
 bin/plantuml.jar:
 	mkdir -p bin
-	curl -Lo $@ https://sourceforge.net/projects/plantuml/files/1.2022.2/plantuml.1.2022.2.jar/download
+	# TODO: Seems that SF is showing updated one, but no download of the previous one we were using: curl -Lo $@ https://sourceforge.net/projects/plantuml/files/1.2022.2/plantuml.1.2022.2.jar/download
+	curl -Lo $@ https://sourceforge.net/projects/plantuml/files/1.2021.9/plantuml-nodot.1.2021.9.jar/download


qbarrand:
Wondering if we should remove the developer pages entirely, including the diagram. We explain in [the docs](https://kmm.sigs.k8s.io/documentation/deploy_kmod/#module-crd) what the reconciliation looks like, and the diagram is most likely outdated.

iranzo:
> Wondering if we should remove the developer pages entirely, including the diagram. We explain in [the docs](https://kmm.sigs.k8s.io/documentation/deploy_kmod/#module-crd) what the reconciliation looks like, and the diagram is most likely outdated.

Might make sense, but as the files are already in text form, we might just update them and keep them around for slides, the website itself, docs, etc.

docs.mk
```diff
@@ -5,4 +5,6 @@ docs: bin/plantuml.jar
 
 bin/plantuml.jar:
 	mkdir -p bin
-	curl -Lo $@ https://sourceforge.net/projects/plantuml/files/1.2022.2/plantuml.1.2022.2.jar/download
\ No newline at end of file
+	# Seems that SF is showing updated one, but no download of the previous one we were using
+	# curl -Lo $@ https://sourceforge.net/projects/plantuml/files/1.2022.2/plantuml.1.2022.2.jar/download
```
ybettan, line 9: Can we remove that line or at least add a `TODO` for it?

iranzo: done

qbarrand:
/hold
Seems like the JAR is back now. @iranzo OK to close?

PR #266, opened 2023-01-24
Do not create ServiceAccounts

qbarrand:
Do not create ServiceAccounts in user namespaces.
Use default ServiceAccounts shipped in the bundle when the Module is created in the operator's namespace.
Remove JobNamespace from the ManagedClusterModule spec and create all build / sign resources in the operator's namespace. Switch the Hub's installMode to OwnNamespace.


ybettan:
/cc @mresvanis
I don't think `/assign` will work because he is not an approver in this repo.

ybettan:
FYI @mresvanis

config/rbac/role.yaml
```diff
@@ -78,17 +78,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
```
yevgeny-shnaidman, line 81: don't we want to keep at least get,list,watch permissions for service accounts? IIRC we talked about examining them in users' namespaces

internal/daemonset/daemonset.go
```diff
@@ -191,7 +198,11 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(ctx context.Context, d
 
 	serviceAccountName := mod.Spec.ModuleLoader.ServiceAccountName
 	if serviceAccountName == "" {
-		serviceAccountName = rbac.GenerateModuleLoaderServiceAccountName(mod)
+		if useDefaultSA {
+			serviceAccountName = "kmm-operator-module-loader"
+		} else {
+			log.FromContext(ctx).Info("Warning: no ServiceAccount set for the ModuleLoader DaemonSet")
```
yevgeny-shnaidman, line 204: Maybe we should return an error here instead of allowing default namespace service account? It will be clearer for us and user to fail here then later

config/manifests-hub/bases/kernel-module-management-hub.clusterserviceversion.yaml
```diff
@@ -26,13 +26,13 @@ spec:
       deployments: null
     strategy: ""
   installModes:
-  - supported: false
+  - supported: true
```
ybettan, line 29: What is the difference between `OwnNamespace` and `SingleNamespace`?

internal/daemonset/daemonset.go
```diff
@@ -254,7 +270,11 @@ func (dc *daemonSetGenerator) SetDevicePluginAsDesired(ctx context.Context, ds *
 
 	serviceAccountName := mod.Spec.DevicePlugin.ServiceAccountName
 	if serviceAccountName == "" {
-		serviceAccountName = rbac.GenerateDevicePluginServiceAccountName(*mod)
+		if useDefaultSA {
+			serviceAccountName = "kmm-operator-device-plugin"
+		} else {
+			log.FromContext(ctx).Info("Warning: no ServiceAccount set for the device plugin DaemonSet")
```
yevgeny-shnaidman, line 276: Same note as above: maybe we want to fail here

config/manifests-hub/bases/kernel-module-management-hub.clusterserviceversion.yaml
qbarrand: [OLM Docs](https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#targetnamespaces-and-their-relationship-to-installmodes)

config/rbac/role.yaml
qbarrand: Then I think we should add them back when we implement that feature.

internal/daemonset/daemonset.go
qbarrand: I don't think it should break the flow: users may have configured their RBAC so that the default account has permission to run privileged pods.

internal/daemonset/daemonset.go
qbarrand: I don't think so.

internal/daemonset/daemonset.go
yevgeny-shnaidman: ok, then let's use util.WarnString API for warning messages

internal/daemonset/daemonset.go
yevgeny-shnaidman: ok, then let's use util.WarnString API for warning messages

internal/daemonset/daemonset.go
mresvanis: > users may have configured their RBAC so that the default account has permission to run privileged pods

I also agree that we have it more flexible and just log a warning. In the future, we could improve the UX by introducing a webhook and/or record an event for example.

PR #270, opened 2023-01-26
Fixing `check-commits-count` to work on other branches than `main`.

ybettan:

---

/cc @qbarrand


ybettan:
/hold
It might not fix the issue

mresvanis:
> /hold
> It might not fix the issue

After researching and internal discussions it seems that this change should work as expected. To sum up, the CI checks out the target branch, merges the PR commits on top of the latter and continues with the respective test.

Issue #274, opened 2023-01-31
The docs refer to older oot api version and dockerfile


cdvultur:
While checking docs on : **https://kmm.sigs.k8s.io/developer/** there are references to 

`apiVersion: ooto.sigs.k8s.io/v1alpha1`  and `dockerfile: |`

Those should be updated to 

`kmm.sigs.x-k8s.io/v1beta1` and  `dockerfileConfigMap:`

PR #275, opened 2023-01-31
Fixing documentation

yevgeny-shnaidman:
Removing section regarding pre-stop lifecycle that needs to be supplied by user and updatting the correct prefixes for labels and apiVersion


ybettan:
Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/274

PR #284, opened 2023-02-06
Refactor mapping and module unification and override

yevgeny-shnaidman:
Currently we have nultiple definitions (Build, Sign, ContainerImage, TLS options) that can be defined both in a mapping and also in the Module Container (more global). These are checked in mulitple places of the code ( during building, signing, ModuleLoader creation etc'), which make it very confusing to keep notice of,and in addition, makes it hard to add new variable both to KerneMapping and to ModuleLoader container This PR moves all the unifications/overrides into one place, at the beginning of the Reconcile loop, and stores the unified data in the KernelMapping. From that point on the code needs to work only with KernelMapping, without any need to check what is the data in ModuleLoadContainer Flow:
- find kernel mapping relevent for kernel version
- unifify data of the Container and Kernel Mapping into the kernel mapping
- all the sign/build/driver container code is now working only with kernel mapping


internal/module/kernelmapper.go
```diff
@@ -29,21 +31,60 @@ type NodeOSConfig struct {
...
+
+	// prepare TLS options
+	if mapping.RegistryTLS == nil {
+		mapping.RegistryTLS = &modSpec.ModuleLoader.Container.RegistryTLS
+	}
+
+	//prepare container image
+	if mapping.ContainerImage == "" {
+		mapping.ContainerImage = modSpec.ModuleLoader.Container.ContainerImage
+	}
```
ybettan, line 83: How about having a generic function that gets the field name as a string and access it using reflection?
I would think of something like:
* If the field is not-nil in the kernel mapping, take in from the mapping otherwise, take it from the global definition.

WDYT?

yevgeny-shnaidman: Currently i don't see the need for that. Using reflection makes code much less readable. In addition, we have fields (Sign, Build) that cannot just be overridden, but have a dedicated function that does the exact logic of overriding.  I would consider using reflect if we had a lot of fields, but in our case, we only have 2-3 fields that can use reflection for overriding

cmd/manager-hub/main.go
```diff
@@ -106,17 +106,19 @@ func main() {
 
 	registryAPI := registry.NewRegistry()
 	jobHelperAPI := utils.NewJobHelper(client)
+	buildHelper := build.NewHelper()
+	signHelper := sign.NewSignerHelper()
```
qbarrand, line 110: That variable is not strictly needed as it's used only once.

cmd/manager/main.go
```diff
@@ -114,23 +114,25 @@ func main() {
 
 	registryAPI := registry.NewRegistry()
 	jobHelperAPI := utils.NewJobHelper(client)
+	buildHelperAPI := build.NewHelper()
+	signHelperAPI := sign.NewSignerHelper()
```
qbarrand, line 118: That variable is not strictly needed as it's used only once.

internal/build/helper.go
```diff
@@ -44,22 +44,22 @@ func (m *helper) ApplyBuildArgOverrides(args []kmmv1beta1.BuildArg, overrides ..
...
-	if km.Build == nil {
-		return modSpec.ModuleLoader.Container.Build.DeepCopy()
+	if mappingBuild == nil {
+		return moduleBuild.DeepCopy()
 	}
 
-	buildConfig := modSpec.ModuleLoader.Container.Build.DeepCopy()
-	buildConfig.DockerfileConfigMap = km.Build.DockerfileConfigMap
+	buildConfig := moduleBuild.DeepCopy()
+	buildConfig.DockerfileConfigMap = mappingBuild.DockerfileConfigMap
```
qbarrand, line 57: What if the mapping's `DockerfileConfigMap` is nil and the `Module` isn't? in that situation, I think we should use the `Module`'s.

internal/module/helper.go
```diff
@@ -37,14 +30,14 @@ func IntermediateImageName(name, namespace, targetImage string) string {
...
-	return modSpec.ModuleLoader.Container.Build != nil || km.Build != nil
+func ShouldBeBuilt(km kmmv1beta1.KernelMapping) bool {
+	return km.Build != nil
 }
 
 // ShouldBeSigned indicates whether the specified KernelMapping of the
 // Module should be signed or not.
-func ShouldBeSigned(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping) bool {
-	return modSpec.ModuleLoader.Container.Sign != nil || km.Sign != nil
+func ShouldBeSigned(km kmmv1beta1.KernelMapping) bool {
```
qbarrand, line 39: If we agree that a `Module`'s sign section should be used as default when it's not defined in the kernel mapping, then should this function accept a [`*Sign`](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/api/v1beta1/module_types.go#L70) instead?

internal/module/helper.go
```diff
@@ -37,14 +30,14 @@ func IntermediateImageName(name, namespace, targetImage string) string {
 
 // ShouldBeBuilt indicates whether the specified KernelMapping of the
 // Module should be built or not.
-func ShouldBeBuilt(modSpec kmmv1beta1.ModuleSpec, km kmmv1beta1.KernelMapping) bool {
-	return modSpec.ModuleLoader.Container.Build != nil || km.Build != nil
+func ShouldBeBuilt(km kmmv1beta1.KernelMapping) bool {
```
qbarrand, line 33: If we agree that a Module's build section should be used as default when it's not defined in the kernel mapping, then should this function accept a [*Build](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/api/v1beta1/module_types.go#L47) instead?

cmd/manager-hub/main.go
yevgeny-shnaidman: fixed

cmd/manager/main.go
yevgeny-shnaidman: fixed

internal/build/helper.go
yevgeny-shnaidman: Nice catch! Fixed. BTW, this is one of the reason that i think mergo will be problematic: much easier to debug this cases in the code then in mergo' tags.

internal/module/helper.go
yevgeny-shnaidman: By the time we call this function ( i assume you are referring to ShouldBeSigned), KernelMapping should contain all unified configuration. So, in case origiinal KM sign section was not defined, the Sign section in the passed kernel mapping will contain the Module's sing section.  We can change the function to receive only Sign section, i just thought it would be more clear API that the decision is made based on the data in KernelMapping

internal/module/helper.go
yevgeny-shnaidman: same answer as above

PR #287, opened 2023-02-07
Add signing to the e2e CI

qbarrand:
/cc @ybettan


ci/module-kmm-ci-build.yaml
```diff
@@ -23,5 +23,14 @@ spec:
             # Optional. If kanikoParams.tag is empty, the default value will be: 'latest'
             kanikoParams:
               tag: "debug"
+          sign:
+            certSecret:
+              name: kmm-kmod-signing-cert
+            keySecret:
+              name: kmm-kmod-signing-key
```
yevgeny-shnaidman, line 30: don't we need public and private key for signing?

qbarrand: The certificate contains the public key.

yevgeny-shnaidman: Let's just change the name of the file to module-kmm-ci-build-sign.yaml

qbarrand: I would argue that it's more convenient to have the resource type in the file name?

yevgeny-shnaidman: but you have <build> in the name, it is because it contains build section, so why not add <sign> so that we know that it contains sign section also?

qbarrand: I get the suggestion now. PTAL.

PR #296, opened 2023-02-09
Adding warning message in case of Sign or Build failures

yevgeny-shnaidman:
Sign and Build  use resources besides those watched by module controller. In case it failed due to those resource (secret. configmaps), the fix involves updating the secrets or configmaps in the cluster. In this case KMMO won't be aware of any changes, and new jobs won't be created. This PR adds a logging message that instructs to delete the Build or Sign job in case of changes in secrets/configmaps


controllers/module_reconciler.go
```diff
@@ -302,6 +304,9 @@ func (r *ModuleReconciler) handleSigning(ctx context.Context,
 		r.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, kernelVersion, metrics.SignStage, false)
 	case utils.StatusCompleted:
 		r.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, kernelVersion, metrics.SignStage, true)
+	case utils.StatusFailed:
+		logger := log.FromContext(ctx).WithValues("module", mod.Name, "image", km.ContainerImage)
+		logger.Info(utils.WarnString("Sign job has failed. If the fix is not in Module CR, then delete job after the fix in order to restart the job"))
```
qbarrand, line 309: `logger` is already declared line 294
```suggestion
		logger.Info(utils.WarnString("Sign job has failed. If the fix is not in Module CR, then delete job after the fix in order to restart the job"))
```

internal/build/job/manager.go
```diff
@@ -138,7 +138,7 @@ func (jbm *jobManager) Sync(
 	case job.Status.Active == 1:
 		return build.Result{Status: build.StatusInProgress, Requeue: true}, nil
 	case job.Status.Failed == 1:
-		return build.Result{}, fmt.Errorf("job failed: %v", err)
+		return build.Result{Status: build.StatusFailed}, fmt.Errorf("job failed: %v", err)
```
qbarrand, line 141: I don't think we should do this. If the returned error is not nil, no one should use the first returned value.
I think we should rather create an error type with the desired error message, return that, and use `if errors.Is()` in the reconciler.

yevgeny-shnaidman: fixed

controllers/module_reconciler.go
yevgeny-shnaidman: fixed

controllers/module_reconciler.go
```diff
@@ -265,6 +265,8 @@ func (r *ModuleReconciler) handleBuild(ctx context.Context,
 		r.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, kernelVersion, metrics.BuildStage, false)
 	case build.StatusCompleted:
 		r.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, kernelVersion, metrics.BuildStage, true)
+	case build.StatusFailed:
+		logger.Info(utils.WarnString("Build job has failed. If the fix is not in Module CR, then delete job after the fix in order to restart the job"))
```
qbarrand, line 269: Should we return an error here? Otherwise I believe we'll proceed with the DaemonSet creation like if the image was successfully built.

internal/cluster/cluster.go
```diff
@@ -86,14 +87,17 @@ func (c *clusterAPI) BuildAndSign(
...
-
 	modSpec := mcm.Spec.ModuleSpec
 	mappings, err := c.kernelMappingsByKernelVersion(ctx, &modSpec, cluster)
 	if err != nil {
 		return false, err
 	}
 
+	// if no mappings were found, return not completed
+	if len(mappings) == 0 {
+		return false, nil
```
ybettan, line 98: 1. At this point we assume that the mapping contains everything needed already?
2. Why don't we return `true` here? If we have nothing to do then we are done aren't we? (same logic that you wrote for `handleSigning`)

controllers/module_reconciler.go
yevgeny-shnaidman: @ybettan , Quentin's comment was based on the previous version of the PR. In the new version, we either return if Build/Sign has completed successfully or not, and based on that proceed to DaemonSet or not.

internal/cluster/cluster.go
```diff
@@ -102,29 +106,30 @@ func (c *clusterAPI) BuildAndSign(
...
-		buildRequeue, err := c.build(ctx, mod, &mcm, m, kernelVersion)
+		buildCompleted, err := c.build(ctx, mod, &mcm, m, kernelVersion)
 		if err != nil {
 			return false, err
 		}
 
-		if buildRequeue {
-			requeue = true
+		if !buildCompleted {
+			completedSuccessfully = false
```
ybettan, line 117: Should we add a log here mentioning we are skipping sign?

internal/cluster/cluster.go
yevgeny-shnaidman: 1. Yes, we assume that all the needed data is stored in the KernelMapping
2. If no mappings are found, we don't want to continue to create ManifestWork, since we don't have suitable images to deploy

controllers/module_reconciler.go
ybettan: I am not sure it was. I can still see this code and there is no `outdated` badge on this comment.

internal/cluster/cluster.go
ybettan: Makes sense.

controllers/module_reconciler.go
yevgeny-shnaidman: but the code surrounding this function has changed and the meaning of the boolean returned by this function has changed

ybettan:
/lgtm
/hold
To make sure @qbarrand is happy with the changes as well.

PR #299, opened 2023-02-13
Fix CSV metadata

qbarrand:
Add display names to CRDs.
Use the Kubernetes logo as image.
Update bundle maintainers and links.


config/samples-hub/hub.kmm.sigs.x-k8s.io_managedclustermodules.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: hub.kmm.sigs.x-k8s.io/v1beta1
+kind: ManagedClusterModule
+metadata:
+  name: managedclustermodule-sample
+spec:
+  jobNamespace: default
```
ybettan, line 6: Didn't we remove that field from the API?

qbarrand: We did indeed. Good catch @ybettan, fixed - thanks!

PR #301, opened 2023-02-14
Creating ModuleLoaderData sturcture for Build/Sign/ModuleLoader workflow

yevgeny-shnaidman:
This PR add the creation of the ModuleLoaderData structure. This struct contains all the data needed for processing the wholw flow of Build/Sign/Module loader per specific Module/KernelMapping.
The flow ass following:
1) when starting processing the Reconcile request( either in Module controller
   or in ManagedCluster controller) all the needed data from Module and KernelMapping
   is merged into a newly created ModuleLoaderData struct
2) all the following layers are working only with this structure, no need to look
   at Module or KernelMapping again.
3) This does not affect the flow of DevicePlugin

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/api"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/daemonset"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
@@ -120,7 +121,7 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return res, fmt.Errorf("could get targeted nodes for module %s: %w", mod.Name, err)
 	}
 
-	mappings, nodesWithMapping, err := r.getRelevantKernelMappingsAndNodes(ctx, mod, targetedNodes)
+	mldMappings, nodesWithMapping, err := r.getRelevantKernelMappingsAndNodes(ctx, mod, targetedNodes)
 	if err != nil {
 		return res, fmt.Errorf("could get kernel mappings and nodes for modules %s: %w", mod.Name, err)
 	}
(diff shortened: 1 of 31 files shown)


controllers/module_reconciler.go
```diff
@@ -130,26 +131,26 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
-	for kernelVersion, m := range mappings {
-		completedSuccessfully, err := r.handleBuild(ctx, mod, m, kernelVersion)
+	for kernelVersion, mld := range mldMappings {
+		completedSuccessfully, err := r.handleBuild(ctx, mld, kernelVersion)
 		if err != nil {
 			return res, fmt.Errorf("failed to handle build for kernel version %s: %v", kernelVersion, err)
 		}
 		if !completedSuccessfully {
-			logger.Info("Build has not finished successfully yet:skipping handling signing and driver container for now", "kernelVersion", kernelVersion, "image", m)
+			logger.Info("Build has not finished successfully yet:skipping handling signing and driver container for now", "kernelVersion", kernelVersion, "mld", mld)
```
ybettan, line 140: How about adding `mld` to the logger instead of manually adding it everywhere?

internal/api/moduleloaderdata.go
```diff
@@ -0,0 +1,45 @@
+package api
+
+import (
+	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ModuleLoaderData struct {
```
ybettan, line 9: Is there a way to somehow generate that object based on the API? So we can regenerate on each API change?

controllers/module_reconciler.go

internal/api/moduleloaderdata.go
yevgeny-shnaidman: not that i know of right now. Mergo can probably help, but this is for later investigation

internal/daemonset/daemonset.go
```diff
@@ -32,7 +33,7 @@ const (
 type DaemonSetCreator interface {
 	GarbageCollect(ctx context.Context, existingDS map[string]*appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error)
 	ModuleDaemonSetsByKernelVersion(ctx context.Context, name, namespace string) (map[string]*appsv1.DaemonSet, error)
-	SetDriverContainerAsDesired(ctx context.Context, ds *appsv1.DaemonSet, image string, mod kmmv1beta1.Module, kernelVersion string) error
+	SetDriverContainerAsDesired(ctx context.Context, ds *appsv1.DaemonSet, image string, mld *api.ModuleLoaderData, kernelVersion string) error
```
ybettan, line 36: Why do we need `image` if it is already in `mld`?

ybettan:
I like this PR a lot. Awesome!

internal/daemonset/daemonset.go
yevgeny-shnaidman: Good catch, thanks! Fixed

internal/api/moduleloaderdata.go
qbarrand: Should this type also have a `KernelVersion` field?

qbarrand: If this is a type that we will be passing around everywhere, should we make it an interface? We could change the implementation as we please.

internal/api/moduleloaderdata.go
```diff
@@ -0,0 +1,45 @@
+package api
```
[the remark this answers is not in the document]
yevgeny-shnaidman: it is problematic, circular import

internal/api/moduleloaderdata.go
yevgeny-shnaidman: 1) comment added
2) kernelVersion added
3) i don't see the benefit: this struct just replaces the KernelMapping and the Module structures. Adding it as interface will just make code more complicated. Currently the only change that i foresee is adding a new field to the structure. If we rename the variable, then fixing it will be a simple matter of recompiling

PR #302, opened 2023-02-14
Add a script generating files for OperatorHub.io

qbarrand:
Make a few adjustements to the base CSVs.

/cc @ybettan

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,6 +48,26 @@ If several `It` sections are similar, consider refactoring them in a `DescribeTa
 - [Kubernetes Contributor Guide](https://git.k8s.io/community/contributors/guide) - Main contributor documentation, or you can just jump directly to the [contributing section](https://git.k8s.io/community/contributors/guide#contributing)
 - [Contributor Cheat Sheet](https://git.k8s.io/community/contributors/guide/contributor-cheatsheet) - Common resources for existing developers
 
+## Managing releases
+
+### OperatorHub.io
+
+After a new version is released, it must be added manually to the
+[k8s-operatorhub/community-operators](https://github.com/k8s-operatorhub/community-operators) repo to be available on
+OperatorHub.io.
+
+First, generate all manifests:
+
+```shell
+make operatorhub-release IMAGE_TAG=<RELEASE_TAG> VERSION=<VERSION_NUMBER>
+```
+
+This will create two directories, `kmm-operatorhub-<VERSION>` and `kmm-hub-operatorhub-<VERSION>`.
+
+Then, use the contents of those directories to create new versions in
+[k8s-operatorhub/community-operators](https://github.com/k8s-operatorhub/community-operators).
+Only one operator should be affected per pull request; create separate PRs if you want to update both KMM bundles.
+
 ## Mentorship
 
 - [Mentoring Initiatives](https://git.k8s.io/community/mentoring) - We have a diverse set of mentorship programs available that are always looking for volunteers!
--- a/Makefile
+++ b/Makefile
@@ -343,5 +343,8 @@ catalog-push: ## Push a catalog image.
 signimage-build: ## Build docker image with the signer.
 	docker build -f Dockerfile.signimage -t $(SIGNER_IMG) .
 
+operatorhub-release:
+	IMG=$(IMG) HUB_IMG=$(HUB_IMG) VERSION=$(VERSION) ./hack/release-operatorhub
+
 include docs.mk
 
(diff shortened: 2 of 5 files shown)


erusso7:
LGTM :+1:

CONTRIBUTING.md
```diff
@@ -42,6 +42,11 @@ Add one `Describe` section per method or function that you are unit testing:
 Within a `Describe`, use one `It` section per test case.
 If several `It` sections are similar, consider refactoring them in a `DescribeTable` construct with varying inputs.
 
+### Generating bundles for OperatorHub.io
+
+Use `hack/release-operator-hub` to generate bundles to be imported in
```
ybettan, line 47:
```suggestion
Use `hack/release-operatorhub` to generate bundles to be imported in
```

`operator-hub` can have different meanings and is not very clear.

ybettan: How about running the script directly in the Makefile in the `bundle` target instead of documenting it?

qbarrand: We could make a Makefile target, but what would be the benefit? You still to provide all the variables needed.

qbarrand:
/hold
Will make this a `Makefile` target.

qbarrand:
@yevgeny-shnaidman @ybettan PTAL

CONTRIBUTING.md
```diff
@@ -48,6 +48,26 @@ If several `It` sections are similar, consider refactoring them in a `DescribeTa
...
+### OperatorHub.io
+
+After a new version is released, it must be added manually to the
+[k8s-operatorhub/community-operators](https://github.com/k8s-operatorhub/community-operators) repo to be available on
+OperatorHub.io.
+
+First, generate all manifests:
+
+```shell
+make operator-hub-release IMAGE_TAG=<RELEASE_TAG> VERSION=<VERSION_NUMBER>
```
ybettan, line 62:
```suggestion
make operatorhub-release IMAGE_TAG=<RELEASE_TAG> VERSION=<VERSION_NUMBER>
```

Using `operator-hub` is confusing with the `kernel-module-management-operator-hub` terminology IMO.

CONTRIBUTING.md
```diff
@@ -48,6 +48,26 @@ If several `It` sections are similar, consider refactoring them in a `DescribeTa
...
+```
+
+This will create two directories, `kmm-operatorhub-<VERSION>` and `kmm-hub-operatorhub-<VERSION>`.
+
+Then, use the contents of those directories to create new versions in
+[k8s-operatorhub/community-operators](https://github.com/k8s-operatorhub/community-operators).
+Only one operator should be affected per pull request; create separate PRs if you want to update both KMM bundles.
```
ybettan, line 69: Why is that? Why can't we make a change to both KMM and KMM-hub in a single PR?
Adding a new feature to hub&spoke will probably modify both KMM and KMM-hub.

Makefile
```diff
@@ -343,5 +343,11 @@ catalog-push: ## Push a catalog image.
 signimage-build: ## Build docker image with the signer.
 	docker build -f Dockerfile.signimage -t $(SIGNER_IMG) .
 
+operator-hub-release:
```
ybettan, line 346:
```suggestion
operatorhub-release:
```

hack/release-operatorhub
```diff
@@ -0,0 +1,67 @@
...
+
+replace_csv () {
+  yq -i "
+    .metadata.annotations.containerImage = \"$container_image\" |
+    .metadata.annotations.createdAt = \"$created_at\"" \
+    "$csv_path"
+}
+
+: "$KMM_CONTAINER_IMG"
+: "$KMM_HUB_CONTAINER_IMG"
```
ybettan, line 30: Should we use the same names as we use in other places? `IMG` and `HUB_IMG`?

hack/release-operatorhub
```diff
@@ -0,0 +1,67 @@
...
+}
+
+created_at=$(date +'%Y-%m-%d %H:%M:%S')
+readonly created_at
+container_image=''
+csv_path=''
+
+replace_csv () {
+  yq -i "
+    .metadata.annotations.containerImage = \"$container_image\" |
```
ybettan, line 24: How about we just pass the image name as an argument to this function instead of using a global variable?
If we want to keep a meaningful name in the function we can always use `container_image=$1`. 

WDYT?

CONTRIBUTING.md
qbarrand: The checklist when creating a new PR mentions this limitation.

ybettan: You mean PRs to https://github.com/k8s-operatorhub/community-operators?

qbarrand: The checklist is the default PR template. See https://github.com/k8s-operatorhub/community-operators/pull/2858 for an example

PR #307, opened 2023-02-20
Changing the metrics flow

yevgeny-shnaidman:
Up to now, the metric were mainly refelcting the current progress of customer's Module, which was not very useful, since customers do not look at them. This PR changes the metrics so as to supply the info regarding KMMO usage by the customer. The following metrics are gathered: 1) number of Modules deployed on the cluster
2) number of modules with in-cluster build capabilty 3) number of module with in-cluster sign capability 4) number of modules deploying device plugin
5) if the preflight capability is used on the cluster 6) modules that are using Morprobe arg configuration ( raw and non-raw)
   and which configuration they are using

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -295,11 +295,8 @@ func (mrh *moduleReconcilerHelper) handleBuild(ctx context.Context, mld *api.Mod
 
 	completedSuccessfully := false
 	switch buildStatus {
-	case utils.StatusCreated:
-		mrh.metricsAPI.SetCompletedStage(mld.Name, mld.Namespace, mld.KernelVersion, metrics.BuildStage, false)
 	case utils.StatusCompleted:
 		completedSuccessfully = true
-		mrh.metricsAPI.SetCompletedStage(mld.Name, mld.Namespace, mld.KernelVersion, metrics.BuildStage, true)
 	case utils.StatusFailed:
 		logger.Info(utils.WarnString("Build job has failed. If the fix is not in Module CR, then delete job after the fix in order to restart the job"))
 	}
@@ -333,11 +330,8 @@ func (mrh *moduleReconcilerHelper) handleSigning(ctx context.Context, mld *api.M
 
 	completedSuccessfully := false
 	switch signStatus {
-	case utils.StatusCreated:
-		mrh.metricsAPI.SetCompletedStage(mld.Name, mld.Namespace, mld.KernelVersion, metrics.SignStage, false)
 	case utils.StatusCompleted:
 		completedSuccessfully = true
-		mrh.metricsAPI.SetCompletedStage(mld.Name, mld.Namespace, mld.KernelVersion, metrics.SignStage, true)
 	case utils.StatusFailed:
 		logger.Info(utils.WarnString("Sign job has failed. If the fix is not in Module CR, then delete job after the fix in order to restart the job"))
 	}
@@ -366,9 +360,6 @@ func (mrh *moduleReconcilerHelper) handleDriverContainer(ctx context.Context,
 	})
 
 	if err == nil {
-		if opRes == controllerutil.OperationResultCreated {
-			mrh.metricsAPI.SetCompletedStage(mld.Name, mld.Namespace, mld.KernelVersion, metrics.ModuleLoaderStage, false)
-		}
 		logger.Info("Reconciled Driver Container", "name", ds.Name, "result", opRes)
 	}
 
@@ -390,9 +381,6 @@ func (mrh *moduleReconcilerHelper) handleDevicePlugin(ctx context.Context, mod *
 	})
 
 	if err == nil {
-		if opRes == controllerutil.OperationResultCreated {
-			mrh.metricsAPI.SetCompletedStage(mod.Name, mod.Namespace, "", metrics.DevicePluginStage, false)
-		}
 		logger.Info("Reconciled Device Plugin", "name", ds.Name, "result", opRes)
 	}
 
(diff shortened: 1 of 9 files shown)


controllers/module_reconciler.go
```diff
@@ -440,9 +428,55 @@ func (mrh *moduleReconcilerHelper) setKMMOMetrics(ctx context.Context) {
...
+		buildFound := false
+		if mod.Spec.DevicePlugin != nil {
+			numModulesWithDevicePlugin += 1
+		}
+		if mod.Spec.ModuleLoader.Container.Build != nil {
+			buildFound = true
+		}
+		if mod.Spec.ModuleLoader.Container.Sign != nil {
+			signFound = true
+		}
```
ybettan, line 449: This is referring only to the global section of the build/sign isn't it? What if there is a build/sign specified only in the kernel mappings?

controllers/module_reconciler.go
```diff
@@ -440,9 +428,55 @@ func (mrh *moduleReconcilerHelper) setKMMOMetrics(ctx context.Context) {
...
+		if !buildFound || !signFound {
+			for _, mapping := range mod.Spec.ModuleLoader.Container.KernelMappings {
+				if mapping.Sign != nil {
+					signFound = true
+				}
+				if mapping.Build != nil {
+					buildFound = true
+				}
+			}
+		}
```
ybettan, line 459: How about encapsulating in a function/method?
We can have a method for build that leverages `ShouldBeBuilt` and another one for sign that uses `ShouldBeSigned`. 
WDYT?

controllers/module_reconciler.go
```diff
@@ -440,9 +428,55 @@ func (mrh *moduleReconcilerHelper) setKMMOMetrics(ctx context.Context) {
...
+		if buildFound {
+			numModulesWithBuild += 1
+		}
+		if signFound {
+			numModulesWithSign += 1
+		}
 
-	mrh.metricsAPI.SetExistingKMMOModules(len(mods.Items))
+		if mod.Spec.ModuleLoader.Container.Modprobe.Args != nil {
+			modprobeArgs := strings.Join(mod.Spec.ModuleLoader.Container.Modprobe.Args.Load, ",")
```
ybettan, line 468: Can we have a slice here instead? The current code will always end with `,`.

controllers/module_reconciler.go
```diff
@@ -440,9 +428,55 @@ func (mrh *moduleReconcilerHelper) setKMMOMetrics(ctx context.Context) {
...
+			mrh.metricsAPI.SetKMMOModprobeArgs(mod.Name, mod.Namespace, modprobeArgs)
+		}
+		if mod.Spec.ModuleLoader.Container.Modprobe.RawArgs != nil {
+			modprobeRawArgs := strings.Join(mod.Spec.ModuleLoader.Container.Modprobe.RawArgs.Load, ",")
```
ybettan, line 472: Can we have a slice here instead? The current code will always end with ,.

controllers/module_reconciler.go
```diff
@@ -440,9 +428,55 @@ func (mrh *moduleReconcilerHelper) setKMMOMetrics(ctx context.Context) {
...
+			mrh.metricsAPI.SetKMMOModprobeRawArgs(mod.Name, mod.Namespace, modprobeRawArgs)
+		}
+	}
+	mrh.metricsAPI.SetKMMOModulesNum(numModules)
+	mrh.metricsAPI.SetKMMOInClusterBuildNum(numModulesWithBuild)
+	mrh.metricsAPI.SetKMMOInClusterSignNum(numModulesWithSign)
+	mrh.metricsAPI.SetKMMODevicePluginNum(numModulesWithDevicePlugin)
```
ybettan, line 479: I would call `mrh.metricsAPI.SetKMMOModprobeArgs` here as well instead for code clarity.

internal/metrics/metrics.go
```diff
@@ -7,65 +7,138 @@ import (
...
+func (m *metrics) SetKMMOPreflightsNum(value int) {
+	m.kmmoPreflightResourceNum.Set(float64(value))
+}
+
+func (m *metrics) SetKMMOModprobeArgs(modName, namespace, modprobeArgs string) {
+	m.kmmoModprobeArgs.WithLabelValues(modName, namespace, modprobeArgs).Set(float64(1))
+}
+
+func (m *metrics) SetKMMOModprobeRawArgs(modName, namespace, modprobeRawArgs string) {
+	m.kmmoModprobeRawArgs.WithLabelValues(modName, namespace, modprobeRawArgs).Set(float64(1))
```
ybettan, line 143: Why do we need those functions? What's wrong with using the inner call directly?

internal/metrics/metrics.go
```diff
@@ -7,65 +7,138 @@ import (
...
-	SignStage                = "sign"
-	ModuleLoaderStage        = "module-loader"
-	DevicePluginStage        = "device-plugin"
+	kmmoModulesQuery         = "kmmo_module_num"
+	kmmoInClusterBuildQuery  = "kmmo_in_cluster_build_num"
+	kmmoInClusterSignQuery   = "kmmo_in_cluster_sign_num"
+	kmmoDevicePluginQuery    = "kmmo_device_plugin_num"
+	kmmoPreflightQuery       = "kmmo_preflight_num"
+	kmmoModprobeArgsQuery    = "kmmo_modprobe_args"
+	kmmoModprobeRawArgsQuery = "kmmo_modprobe_raw_args"
```
qbarrand, line 16: Shall we rename to `kmm_*`?

controllers/module_reconciler_test.go
```diff
@@ -728,3 +722,109 @@ var _ = Describe("ModuleReconciler_garbageCollect", func() {
...
+
+		mhr.setKMMOMetrics(ctx)
+		/*
+					clnt.EXPECT().List(context.Background(), gomock.Any(), gomock.Any()).DoAndReturn(
+			                        func(_ interface{}, list *v1.NodeList, _ ...interface{}) error {
+			                                list.Items = []v1.Node{node1, node2, node3}
+			                                return nil
+			                        },
+			                )
+		*/
```
qbarrand, line 754: Should we remove this comment?

controllers/module_reconciler.go
```diff
@@ -479,3 +496,25 @@ func isNodeSchedulable(node *v1.Node) bool {
...
+
+func isModuleBuildAndSignCapable(mod *kmmv1beta1.Module) (bool, bool) {
+	buildCapable := false
+	signCapable := false
+	if mod.Spec.ModuleLoader.Container.Build != nil {
+		buildCapable = true
+	}
+	if mod.Spec.ModuleLoader.Container.Sign != nil {
+		signCapable = true
+	}
```
qbarrand, line 508:
```suggestion
	buildCapable := mod.Spec.ModuleLoader.Container.Build != nil
	signCapable := mod.Spec.ModuleLoader.Container.Sign != nil
```

controllers/module_reconciler.go
```diff
@@ -479,3 +496,19 @@ func isNodeSchedulable(node *v1.Node) bool {
 	}
 	return true
 }
+
+func isModuleBuildAndSignCapable(mod *kmmv1beta1.Module) (bool, bool) {
+	buildCapable := mod.Spec.ModuleLoader.Container.Build != nil
+	signCapable := mod.Spec.ModuleLoader.Container.Sign != nil
+	if !buildCapable || !signCapable {
```
ybettan, line 503:
```suggestion

	if buildCapable && signCapable {
	    return true, true
	}
```

ybettan: No need to iterate if we got all we need already.

ybettan: nit

Issue #103, opened 2022-10-07
ER: Expose Module name as build environment variable


cdvultur:
I find helpful to have the module name exposed as an environment variable.
This is helpful in the build process where the module name could be needed in the build process. Also will help in case of yaml reuse.

qbarrand:
It would probably use useful to have access to the `Module`'s namespace as well. Those two values should be available both in the `containerImage` field and as build arguments when building in-cluster.

qbarrand:
Rephrasing this issue a bit to make it more clear.
Just like we have `KERNEL_FULL_VERSION` available for substitution [in the `Module`'s `containerImage` field](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/ci/module-kmm-ci-build-sign.yaml#L13) and `KERNEL_VERSION` [available as a build argument](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/ci/kmm-kmod-dockerfile.yaml#L9) when building, it would be useful to also have the `Module`'s name and namespace as new variables.

PR #310, opened 2023-02-23
Expose module's name and namespace

erusso7:
Expose module's name and namespace  through build args and use them in the `containerImage`

#103

### DIFF
--- a/internal/build/job/maker.go
+++ b/internal/build/job/maker.go
@@ -70,23 +70,20 @@ func (m *maker) MakeJobTemplate(
 	owner metav1.Object,
 	pushImage bool) (*batchv1.Job, error) {
 
-	buildConfig := mld.Build
-	containerImage := mld.ContainerImage
-
 	// if build AND sign are specified, then we will build an intermediate image
 	// and let sign produce the one specified in its targetImage
+	containerImage := mld.ContainerImage
 	if module.ShouldBeSigned(mld) {
 		containerImage = module.IntermediateImageName(mld.Name, mld.Namespace, containerImage)
 	}
 
-	specTemplate := m.specTemplate(
-		mld,
-		buildConfig,
-		containerImage,
-		mld.RegistryTLS,
-		pushImage)
-
-	specTemplateHash, err := m.getHashAnnotationValue(ctx, buildConfig.DockerfileConfigMap.Name, mld.Namespace, &specTemplate)
+	specTemplate := m.specTemplate(mld, containerImage, pushImage)
+	specTemplateHash, err := m.getHashAnnotationValue(
+		ctx,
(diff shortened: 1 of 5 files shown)


internal/build/job/maker.go
```diff
@@ -148,7 +148,7 @@ func (m *maker) specTemplate(
 
 func (m *maker) containerArgs(
 	buildConfig *kmmv1beta1.Build,
-	targetKernel string,
+	mld *api.ModuleLoaderData,
 	containerImage string,
 	registryTLS *kmmv1beta1.TLSOptions,
```
yevgeny-shnaidman, line 153: registryTLS can also be obtained from mld.  It is passed from specTemplate function, so it can be removed from specTemnplate's arguments and obtained directly from mld

ci/kmm-kmod-dockerfile.yaml
```diff
@@ -7,6 +7,8 @@ data:
     FROM ubuntu as builder
 
     ARG KERNEL_VERSION
+    ARG MOD_NAME
```
yevgeny-shnaidman, line 10: what is the use of those parameters during build?

qbarrand: We don't really have one in our CI example, but I think we should have all substitutions in the `containerImage` also available as default build arguments to avoid confusion.

internal/build/job/maker.go
erusso7: correct, good catch!

internal/build/job/maker.go
```diff
@@ -70,23 +70,20 @@ func (m *maker) MakeJobTemplate(
...
-	specTemplate := m.specTemplate(
-		mld,
-		buildConfig,
-		containerImage,
-		mld.RegistryTLS,
-		pushImage)
-
-	specTemplateHash, err := m.getHashAnnotationValue(ctx, buildConfig.DockerfileConfigMap.Name, mld.Namespace, &specTemplate)
+	specTemplate := m.specTemplate(mld, containerImage, pushImage)
+	specTemplateHash, err := m.getHashAnnotationValue(
```
erusso7, line 81: @yevgeny-shnaidman: In this case, as we are not changing the `buildConfig` before calling the `m.specTemplate`, I decided to remove it as a parameter and use the `mlm.Build` pointer. 
Is that ok, for you?

ci/kmm-kmod-dockerfile.yaml
erusso7: Correct, if we are not using in our CI dockerfile, we can remove them :+1:

internal/build/job/maker.go
yevgeny-shnaidman: yes, great!

Issue #140, opened 2022-11-09
Separate node selector for build process


hershpa:
Feature Summary:
Separate node selector for build process.

Suggested Priority(P1-P3) & Urgent(Urgent, medium, Low):
P2 and medium

Feature Detail:
Currently, KMM uses node selector for both build and DaemonSet. It would be nice to have a separate node selector for build process since it is not always required that the build occur on the same node as that selected for the DaemonSet to modprobe the kernel module. This allows flexibility for the build to occur on any other node. 

Solution Proposal:
Provide a separate node selector for build process.

qbarrand:
If the build selector is empty or undefined, let us keep using the global one to remain compatible.

ybettan:
Should we use the same "new selector" for signing? or also add a selector for signing?

If we don't then we still have the same issue of waiting for a "special" node with some HW for signing a module that was already built on another node in case we both build an sign the module.

WDYT?

qbarrand:
In which situations would we need special hardware to sign the files?

ybettan:
In case there is no `signSelector` and the user has specified a sign section in the module (let's say without build). The sign job will only be scheduled on nodes with the selector of the module (aka the nodes that a customer will load his driver too) which I believe is what we are trying to avoid for build jobs in this issue.

ybettan:
Hello @hershpa.

After discussing this issue, we tough about a solution that won't modify the API at this point.

As we understand it, your main requirement, is to not *restrict* the build/sign jobs to the same node as the moduleLoader is going to run, but instead to just run on any available node. Is that correct?

If this is the case, we can just remove the node selector from the build/sign jobs and let k8s do his job and schedule the jobs on the best fitting nodes. The `nodeSelector` of the `ModuleLoader` will remain with no changes.

Will this solution be good enough for your usecase at this point?

hershpa:
Hi @ybettan, 

That is correct and sounds good.

ybettan:
As agreed in the community meeting we will proceed with the following:

We add a new selector for build. Regarding the default value we differentiate between v1 and v2.

v1:
If the build selector is not set, we use the selector from the module to maintain the same behavior that we had until now.

v2:
If the build selector is not set, we do not set any selector for the build job and let k8s make the scheduling.

PR #311, opened 2023-02-23
Define new selector for building image job

erusso7:
#140


internal/build/job/maker.go
```diff
@@ -129,6 +129,11 @@ func (m *maker) specTemplate(
 		kanikoImage += ":" + buildConfig.KanikoParams.Tag
 	}
 
+	nodeSelector := mld.Selector
```
yevgeny-shnaidman, line 132: why do we need to set the nodeSelector at all for Builds? Why not let kubernetes decide where to run the Build job?

ybettan:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_kernel-module-management/311/pull-kernel-module-management-check-api-changes/1630142470052384768#1:build-log.txt%3A201

api/v1beta1/module_types.go
```diff
@@ -65,6 +65,10 @@ type Build struct {
 	// +optional
 	// KanikoParams is used to customize the building process of the image.
 	KanikoParams *KanikoParams `json:"kanikoParams,omitempty"`
+
+	// +optional
+	// Selector describes on which nodes the Job should be created.
+	Selector map[string]string `json:"selector,omitempty"`
```
ybettan, line 71: How about a `Sign` selector? What happens if we need to build and sign? We still have the same issue as we had before aren't we? The user will still need a "special node" with some HW in order to sign.

internal/build/job/maker.go
ybettan: This is a good point. Removing it completely will also solve https://github.com/kubernetes-sigs/kernel-module-management/issues/140#issuecomment-1446030028

qbarrand: > Why not let kubernetes decide where to run the Build job?

This comes from [that issue](https://github.com/kubernetes-sigs/kernel-module-management/issues/140). You may want to reserve nodes equipped with expensive hardware for some workloads only, for example.

yevgeny-shnaidman: I don't see in that issue any mention of nodes with expensive hardware. They just say that it would be nice if build/sign can also run on other nodes, and not only those with specific hardware.  Also,  allowing kuberentes to pick the running node makes use of the schedule, which probably has much more data to decide which nodes are less "overloaded"

qbarrand: > Also, allowing kuberentes to pick the running node makes use of the schedule, which probably has much more data to decide which nodes are less "overloaded"

This change makes a user able to do just that, with `selector: {}` in the `build` section.

qbarrand:
/hold
This change may brake flows for users building for a specific architecture in multi-arch clusters.
Let us consider a cluster with `amd64` and `arm64` nodes.
Before this change, if you created a `Module` targeting `amd64` nodes, the builds would be automatically scheduled on `amd64` nodes.
With this change, the build may run on an `arm64` node, which by default will produce a module for `arm64`, not `amd64`.

cc @hershpa @yevgeny-shnaidman @ybettan

yevgeny-shnaidman:
> /hold This change may brake flows for users building for a specific architecture in multi-arch clusters. Let us consider a cluster with `amd64` and `arm64` nodes. Before this change, if you created a `Module` targeting `amd64` nodes, the builds would be automatically scheduled on `amd64` nodes. With this change, the build may run on an `arm64` node, which by default will produce a module for `arm64`, not `amd64`.
> 
> cc @hershpa @yevgeny-shnaidman @ybettan

@qbarrand maybe all we need to add  a label to the Build/Sign job node selector is architecture label. I think that all the nodes are label with it anyways, and it can be derived from the KernelMapping

ybettan:
> This change may brake flows for users building for a specific architecture in multi-arch clusters.
Let us consider a cluster with amd64 and arm64 nodes.
Before this change, if you created a Module targeting amd64 nodes, the builds would be automatically scheduled on amd64 nodes.
With this change, the build may run on an arm64 node, which by default will produce a module for arm64, not amd64.

@qbarrand @yevgeny-shnaidman I would instead just ask kaniko to build for a specific arch (regardless of the host) using https://github.com/GoogleContainerTools/kaniko#flag---customplatform.

yevgeny-shnaidman:
> > This change may brake flows for users building for a specific architecture in multi-arch clusters.
> > Let us consider a cluster with amd64 and arm64 nodes.
> > Before this change, if you created a Module targeting amd64 nodes, the builds would be automatically scheduled on amd64 nodes.
> > With this change, the build may run on an arm64 node, which by default will produce a module for arm64, not amd64.
> 
> @qbarrand @yevgeny-shnaidman I would instead just ask `kaniko` to build for a specific arch (regardless of the host) using https://github.com/GoogleContainerTools/kaniko#flag---customplatform.
> 
> This will also allow to build for ARM in a x86 cluster or any other combination. This is especially handy in the hub-spoke topology.
> 
> This will requires extending the `build` section in our CRD though.

@ybettan does kaniko support something like that? build Dockerfile will probably include Makefile that needs to know how to cross-compile, not sure if customer will want to support it.

ybettan:
This PR is blocked by https://github.com/kubernetes-sigs/kernel-module-management/issues/325 unless we modify the CRD to contain a `nodeSelector` for build/sign which is not ideal as we wish not to extend the API unless there is a requirement for it.

qbarrand:
> we wish not to extend the API unless there is a requirement for it

Sounds like it's the only way to properly address #140 though.

ybettan:
@qbarrand 
We have `kubernetes.io/arch: amd64` on the nodes, how about we just set the inner selector of the build to match the same arch as of the module's selector node?

ybettan:
As agreed in the community meeting we will proceed with the following:

We add a new selector for build

v1:
If the build selector is not set, we use the selector from the module to maintain the same behavior that we had until now.

v2:
If the build selector is not set, we do not set any selector for the build job and let k8s make the scheduling.

api/v1beta1/module_types.go
erusso7: @ybettan I will close this comment because based on the community meeting and the notes, this is no longer valid, if so, we can just re-open it again :+1:

internal/build/job/maker.go
```diff
@@ -122,6 +122,14 @@ func (m *maker) specTemplate(mld *api.ModuleLoaderData, containerImage string, p
 		kanikoImage += ":" + buildConfig.KanikoParams.Tag
 	}
 
+	var selector map[string]string
+	if len(mld.Selector) != 0 {
+		selector = mld.Selector
+	}
```
ybettan, line 128:
```suggestion
    selector = mld.Selector
```

internal/sign/job/signer.go
```diff
@@ -193,7 +193,6 @@ func (s *signer) MakeJobTemplate(
 			},
 			RestartPolicy: v1.RestartPolicyNever,
 			Volumes:       volumes,
-			NodeSelector:  mld.Selector,
```
ybettan, line 196: Why are we removing it? This is breaking backward compatibility.

PR #313, opened 2023-02-27
Use kaniko for signing

qbarrand:
Replace the custom implementation with a kaniko-based setup where we generate the Dockerfile dynamically from a template.
Remove methods from the registry package that are now unused.

This code should be refactored in a subsequent PR to extract the template rendering and find more commonality between the build and signing code.

/hold
Giving time to @chr15p to review since he authored most of that code.


internal/build/job/maker.go
```diff
@@ -205,18 +201,18 @@ func (m *maker) getHashAnnotationValue(ctx context.Context, configMapName, names
 }
 
 func volumes(imageRepoSecret *v1.LocalObjectReference, buildConfig *kmmv1beta1.Build) []v1.Volume {
-	volumes := []v1.Volume{dockerfileVolume(dockerfileVolumeName, buildConfig.DockerfileConfigMap)}
+	volumes := []v1.Volume{dockerfileVolume(DockerfileVolumeName, buildConfig.DockerfileConfigMap)}
```
yevgeny-shnaidman, line 204: why do we need those changes for this PR? we don't need to change build at all, just duplicate what is needed into sign, and the do the real refactoring in the next PR. IMHO it will make this PR more clear

internal/sign/job/templates/Dockerfile.gotmpl
```diff
@@ -0,0 +1,20 @@
+{{- /*gotype: github.com/rh-ecosystem-edge/kernel-module-management/internal/sign/job.TemplateData */ -}}
+FROM {{ .UnsignedImage }} as source
+
+FROM alpine:3.17 AS signimage
+
+RUN ["apk", "add", "linux-virt-dev"]
```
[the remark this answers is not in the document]
qbarrand: It only takes seconds. That does not justify maintaining an additional image IMO.

internal/build/job/maker.go
qbarrand: Removed the changes to `internal/build` from this PR.

chr15p:
/lgtm
there's not a lot of my code left!

PR #316, opened 2023-02-28
Adding the namespace to the `<module>.ready` label.

ybettan:
This is handy when there are multiple modules on each node and make it easier to determine where is each module.

---

This is how the new label looks like:
```
kmm.node.kubernetes.io/default.kmm-ci.ready: ""
```

### DIFF
--- a/internal/daemonset/daemonset_test.go
+++ b/internal/daemonset/daemonset_test.go
@@ -496,7 +496,7 @@ var _ = Describe("SetDevicePluginAsDesired", func() {
 						},
 						ImagePullSecrets: []v1.LocalObjectReference{repoSecret},
 						NodeSelector: map[string]string{
-							getDriverContainerNodeLabel(mod.Name): "",
+							getDriverContainerNodeLabel(mod.Namespace, mod.Name): "",
 						},
 						PriorityClassName:  "system-node-critical",
 						ServiceAccountName: serviceAccountName,
@@ -762,6 +762,8 @@ var _ = Describe("OverrideLabels", func() {
 })
 
 var _ = Describe("GetNodeLabelFromPod", func() {
+
+	const namespace = "some-namespace"
 	var dc DaemonSetCreator
 
 	BeforeEach(func() {
@@ -775,10 +777,11 @@ var _ = Describe("GetNodeLabelFromPod", func() {
 					constants.ModuleNameLabel: moduleName,
 					kernelLabel:               "some kernel",
 				},
+				Namespace: namespace,
 			},
 		}
 		res := dc.GetNodeLabelFromPod(&pod, "module-name")
-		Expect(res).To(Equal(getDriverContainerNodeLabel("module-name")))
+		Expect(res).To(Equal(getDriverContainerNodeLabel(namespace, "module-name")))
 	})
 
 	It("should return a device plugin label", func() {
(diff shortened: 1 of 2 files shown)


internal/daemonset/daemonset_test.go
```diff
@@ -775,10 +777,11 @@ var _ = Describe("GetNodeLabelFromPod", func() {
 					constants.ModuleNameLabel: moduleName,
 					kernelLabel:               "some kernel",
 				},
+				Namespace: namespace,
 			},
 		}
 		res := dc.GetNodeLabelFromPod(&pod, "module-name")
-		Expect(res).To(Equal(getDriverContainerNodeLabel("module-name")))
+		Expect(res).To(Equal(getDriverContainerNodeLabel(namespace, "module-name")))
```
qbarrand, line 784: I realize that this change replicates what we were previously doing, but we're never actually verifying that `getDriverContainerNodeLabel` or `getDevicePluginNodeLabel` do things correctly.
```suggestion
		Expect(res).To(Equal("kmm.node.kubernetes.io/some-namespace.module-name.ready"))
```

internal/daemonset/daemonset_test.go
```diff
@@ -787,10 +790,11 @@ var _ = Describe("GetNodeLabelFromPod", func() {
 				Labels: map[string]string{
 					constants.ModuleNameLabel: moduleName,
 				},
+				Namespace: namespace,
 			},
 		}
 		res := dc.GetNodeLabelFromPod(&pod, "module-name")
-		Expect(res).To(Equal(getDevicePluginNodeLabel("module-name")))
+		Expect(res).To(Equal(getDevicePluginNodeLabel(namespace, "module-name")))
```
qbarrand, line 797:
```suggestion
		Expect(res).To(Equal("kmm.node.kubernetes.io/some-namespace.module-name.device-plugin-ready"))
```

qbarrand (commented):
I think we should add the new namespaced label, but also leave the existing one for compatibility purposes.

ybettan:
This is how the new label looks like:
```
kmm.node.kubernetes.io/default.kmm-ci.ready: ""
```

ybettan:
> I think we should add the new namespaced label, but also leave the existing one for compatibility purposes.

So you want to have both for each module?
```
kmm.node.kubernetes.io/default.kmm-ci.ready: ""
kmm.node.kubernetes.io/kmm-ci.ready: ""
```

What compatibility issue we are having here? Isn't the only consumer of that label is the DevicePlugin container?

internal/daemonset/daemonset_test.go
ybettan: If they fails then the test for `GetNodeLabelFromPod` will fail as well. Did I misunderstood your comment?

qbarrand:
> Isn't the only consumer of that label is the DevicePlugin container?

The label is usable by anyone. If we remove it, we break the solution of customers that were potentially using it.

yevgeny-shnaidman:
> > Isn't the only consumer of that label is the DevicePlugin container?
> 
> The label is usable by anyone. If we remove it, we break the solution of customers that were potentially using it.

Those labels are KMM internal implementation. It is not an API that we are supporting. So i think we can change them as we see fit , as long as they don't break the code once a new operator is loaded while a Module is in the cluster

qbarrand:
Let's wait for feedback until the 2023-03-08 community sync.
/hold

PR #324, opened 2023-03-02
Fix signing Dockerfile

qbarrand:
Fix the path to the sign-file binary.
Build the signing image in CI.

### DIFF
--- a/.github/actions/collect-troubleshooting/action.yml
+++ b/.github/actions/collect-troubleshooting/action.yml
@@ -26,6 +26,10 @@ runs:
       run: kubectl describe pod -l kmm.node.kubernetes.io/module.name
       shell: bash
 
+    - name: Describe Jobs
+      run: kubectl describe job.batch -l kmm.node.kubernetes.io/module.name
+      shell: bash
+
     - name: Collect dmesg
       run: sudo dmesg
       shell: bash
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -30,11 +30,34 @@ jobs:
           path: kmm_local.tar
           retention-days: 1
 
+  build-signing-image:
+    runs-on: ubuntu-latest
+
+    name: Build the signing image
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build the image
+        run:  make signimage-build SIGNER_IMG=kmm-signimage:local
+
+      - name: Export the image
+        run: docker save -o kmm-signimage_local.tar kmm-signimage:local
+
+      - name: Upload the image
+        uses: actions/upload-artifact@v3
+        with:
+          name: ci-images
+          if-no-files-found: error
+          path: kmm-signimage_local.tar
+          retention-days: 1
+
   in-cluster-build:
 
     runs-on: ubuntu-latest
     name: in-cluster-build
-    needs: [build-operator-image]
+    needs: [build-operator-image, build-signing-image]
 
     services:
       registry:
@@ -56,8 +79,14 @@ jobs:
         with:
           name: ci-images
 
-      - name: Import the KMMO image into minikube
+      - name: Import the KMM operator image into minikube
         run: minikube image load kmm_local.tar
+
+      - name: Copy the signing image to the registry
+        run: |
+          docker load -i kmm-signimage_local.tar
+          docker tag kmm-signimage:local localhost:5000/kmm/signimage:local
+          docker push localhost:5000/kmm/signimage:local
 
       - name: Cache binaries needed by Makefile
         id: cache-bin
--- a/Makefile
+++ b/Makefile
@@ -312,7 +312,7 @@ catalog-push: ## Push a catalog image.
 
 .PHONY: signimage-build 
 signimage-build: ## Build docker image with the signer.
-	docker build -f Dockerfile.signimage -t $(SIGNER_IMG)
+	docker build -f Dockerfile.signimage -t $(SIGNER_IMG) .
 
 include docs.mk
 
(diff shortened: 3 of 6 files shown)


.github/actions/collect-troubleshooting/action.yml
```diff
@@ -26,6 +26,10 @@ runs:
       run: kubectl describe pod -l kmm.node.kubernetes.io/module.name
       shell: bash
 
+    - name: Describe Jobs
+      run: kubectl describe job.batch -l kmm.node.kubernetes.io/module.name
+      shell: bash
```
ybettan, line 31: I don't think this will work. I think that the GC will remove it before we reach this stage.
This is why we have https://github.com/kubernetes-sigs/kernel-module-management/blob/107d0b054ef6d5a66fc07c5a321fdddc75f7353c/ci/prow/e2e-incluster-build#L48.

.github/workflows/e2e.yaml
```diff
@@ -30,11 +30,34 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    name: Build the signing image
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build the image
+        run:  make signimage-build SIGNER_IMG=kmm-signimage
```
ybettan, line 43:
```suggestion
        run:  make signimage-build SIGNER_IMG=kmm-signimage:local
```

.github/workflows/e2e.yaml
```diff
@@ -30,11 +30,34 @@ jobs:
...
+
+      - name: Export the image
+        run: docker save -o kmm-signimage.tar kmm-signimage
```
ybettan, line 46:
```suggestion
        run: docker save -o kmm-signimage.tar kmm-signimage:local
```

.github/workflows/e2e.yaml
```diff
@@ -30,11 +30,34 @@ jobs:
...
+
+      - name: Upload the image
+        uses: actions/upload-artifact@v3
+        with:
+          name: ci-images
+          if-no-files-found: error
+          path: kmm-signimage.tar
```
ybettan, line 53:
```suggestion
          path: kmm_signimage_local.tar
```

Makefile
```diff
@@ -312,7 +312,7 @@ catalog-push: ## Push a catalog image.
 
 .PHONY: signimage-build 
 signimage-build: ## Build docker image with the signer.
-	docker build -f Dockerfile.signimage -t $(SIGNER_IMG)
+	docker build -f Dockerfile.signimage -t $(SIGNER_IMG) .
```
ybettan, line 315: How does using `.` and also `-f Dockerfile.signimage` works?

ci/install-ci/kustomization.yaml
```diff
@@ -22,3 +22,6 @@ patchesStrategicMerge:
           containers:
             - name: manager
               imagePullPolicy: Never
+              env:
+                - name: RELATED_IMAGES_SIGN
+                  value: host.minikube.internal:5000/kmm/signimage:latest
```
ybettan, line 27: Can we extend `ci/install-ci/kustomization.yaml` instead?

.github/workflows/e2e.yaml
```diff
@@ -56,9 +79,12 @@ jobs:
         with:
           name: ci-images
 
-      - name: Import the KMMO image into minikube
+      - name: Import the KMM operator image into minikube
         run: minikube image load kmm_local.tar
 
+      - name: Copy the signing image to the registry
+        run: skopeo copy --dest-tls-verify=false tarball:kmm-signimage.tar docker://localhost:5000/kmm/signimage:latest
```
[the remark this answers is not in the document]
qbarrand: Because that would insert the image into the minikube node, and the signing job can't pull it - it needs to be in a registry accessible from the pods.

.github/workflows/e2e.yaml
qbarrand: Not sure if we want to append `local`, I initially used it to signal that it's inserted into the minikube nodes - not the case here.

Makefile
qbarrand: `.` is for the [build context](https://docs.docker.com/build/building/context/) and `-f` tells Docker which Dockerfile to use.

ci/install-ci/kustomization.yaml
qbarrand: Can you be more explicit?

ybettan: Sorry. I though you have added it as an env as part of `.github/workflow/e2e.yaml` and I suggested to add the env variable to `ci/install-ci/kustomization.yaml` instead but this is exactly what you did!

You can close this thread.

.github/workflows/e2e.yaml
ybettan: I think that "local" means implies that this image was build locally in CI (unlike using remote images unrelated to the changes in a PR) regardless of it is in the node registry or in `minikube`'s registry.

As long as it is built as part of the PR and not pushed anywhere remotely, I think we should call it `local` or else just remove the `local` from KMM as well - because there is no *logical* difference between the 2. It is just confusing IMO.

.github/workflows/e2e.yaml
```diff
@@ -30,11 +30,34 @@ jobs:
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build the image
+        run:  make signimage-build SIGNER_IMG=kmm-signimage:local
+
+      - name: Export the image
+        run: docker save -o kmm-signimage_local.tar kmm-signimage:local
```
ybettan, line 46: How about
```suggestion
        run: docker save -o kmm_signimage_local.tar kmm-signimage:local
```
?

qbarrand: Let's keep the dash like in the original image name.

PR #327, opened 2023-03-06
Adding Version field to Module CRD

yevgeny-shnaidman:
Version field will be used in the ordered upgrade scenario. It signifies the current version of the kernel module.
It should be updated together with the new image in the appropriate kernel mapping. Its update will start the process of the ordered upgrade

### DIFF
--- a/api/v1beta1/module_types.go
+++ b/api/v1beta1/module_types.go
@@ -165,6 +165,11 @@ type ModuleLoaderContainerSpec struct {
 	// Sign provides default kmod signing settings
 	Sign *Sign `json:"sign,omitempty"`
 
+	// Version defines the current version of the kernel module being used
+	// Used for upgrading the currently loaded kernel module to a new version
+	// +optional
+	Version string `json:"version,omitempty"`
+
 	// ContainerImage is a top-level field
 	// +optional
 	ContainerImage string `json:"containerImage,omitempty"`
--- a/config/crd-hub/bases/hub.kmm.sigs.x-k8s.io_managedclustermodules.yaml
+++ b/config/crd-hub/bases/hub.kmm.sigs.x-k8s.io_managedclustermodules.yaml
@@ -2442,6 +2442,11 @@ spec:
                             - certSecret
                             - keySecret
                             type: object
+                          version:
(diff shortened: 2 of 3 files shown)


api/v1beta1/module_types.go
```diff
@@ -165,6 +165,11 @@ type ModuleLoaderContainerSpec struct {
 	// Sign provides default kmod signing settings
 	Sign *Sign `json:"sign,omitempty"`
 
+	// KernelModuleVersion defines the current version of the kernel module being used
+	// Used for upgrading the currently loaded kernel module to a new version
+	// +kubebuilder:default="1.0"
+	KernelModuleVersion string `json:"kernelModuleVersion"`
```
qbarrand, line 171: Why `kernelModuleVersion` and not just `moduleVersion`? Or even `version`?

yevgeny-shnaidman: I am opened to whatever naming you deem suitable. Initially i named it ModuleVersion, but then i thought that KernelModuleVersion is more accurate.

ybettan: I like the `version` suggestions as well.

yevgeny-shnaidman: fixed

api/v1beta1/module_types.go
```diff
@@ -165,6 +165,11 @@ type ModuleLoaderContainerSpec struct {
 	// Sign provides default kmod signing settings
 	Sign *Sign `json:"sign,omitempty"`
 
+	// Version defines the current version of the kernel module being used
+	// Used for upgrading the currently loaded kernel module to a new version
+	// +kubebuilder:default="1.0"
```
qbarrand, line 170: Why do we want a default value? I think we want this value to be completely optional (with the appropriate marker).

yevgeny-shnaidman: in case user does not set the value of the version, we do want it to have value, even logically. User does not need to update it, if he does not want to use the ordered upgrade scenario. In addition, it is good to have a value in that field, since we will start differentiate DaemonSets of the ModuleLoader and DevicePlugins based not only on kernel mappings , but also on Version

Issue #315, opened 2023-02-28
`kmm.node.kubernetes.io/<module-name>.ready` can be removed by accident in some cases


ybettan:
We are removing the label in case a pod is in terminating state or deleted but there are some cases in which a pod restarts because it crashed for example. In this case for a short period of time we are going to have 2 pods:

* one in terminating state
*  a new one that has been created an is now in running state.
Once the terminating pod will be deleted, KMM will delete the label because it doesn't differentiate between different pods of the same module.

Steps to reproduce:

1. Create a module
2. Delete the module and directly recreate it (you will now have 2 pods as explained)
3. Wait for the terminating pod to finish and then check the labels
4. The label will be gone although we have a module and a running pod

ybettan:
@yevgeny-shnaidman has suggested one possible solution to check if we have a module in the cluster before deleting the label.

ybettan:
Thoughts @qbarrand @mresvanis ?

qbarrand:
I do not see any good way of solving this before v2, unfortunately.

yevgeny-shnaidman:
@ybettan   actually we can use the key=value option, where value is the pod name (as you suggested). There is an option for set-based selectors, meaning that device-plugin needs to just specify the key without value ( if i understand it correctly).  Please see [](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors)

yevgeny-shnaidman:
[set-based](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors)

ybettan:
I have tried to use the `key=value` label and then use `nodeAffinity` in the devicePlugin daemonset but seems to be useless as we can't label the same node with 2 labels that contains the same `key` but different values.

qbarrand:
How about we check if there is a newer pod for the same `Module` on the node when we remove the label?

yevgeny-shnaidman:
it means adding query for all the pods of the Module.  why do we want to go that way? the way @ybettan  is cleaner, in my opinion

qbarrand:
It merely queries the cache - no request going to the API server. And we can fix this bug without changing the existing label or introducing a new one.

yevgeny-shnaidman:
we are just changing our own internal label. And it seems that the code will be a little bit cleaner

PR #328, opened 2023-03-06
Adding the namespace to the `<module>.ready` label.

ybettan:
Adding the namespace to the `<module>.ready` label.

This is handy when there are multiple modules on each node and make it
easier to determine where is each module.

Also, solving a race condition in a new pod was created while the old
pod of a `moduleLoader` is being terminated which resulted in the
`<module>.ready` label to be deleted.

---

Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/315

### DIFF
--- a/controllers/pod_node_module_reconciler.go
+++ b/controllers/pod_node_module_reconciler.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/kubectl/pkg/util/podutils"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
@@ -53,7 +54,8 @@ func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		return ctrl.Result{}, fmt.Errorf("pod %s has no %q label", podNamespacedName, constants.ModuleNameLabel)
 	}
 
-	labelName := pnmr.daemonAPI.GetNodeLabelFromPod(&pod, moduleName)
+	labelName := pnmr.daemonAPI.GetNodeLabelFromPod(&pod, moduleName, false)
+	deprecatedLabelName := pnmr.daemonAPI.GetNodeLabelFromPod(&pod, moduleName, true)
 
 	logger = logger.WithValues(
 		"node name", nodeName,
@@ -70,8 +72,26 @@ func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if !podutils.IsPodReady(&pod) || !pod.DeletionTimestamp.IsZero() {
 		logger.Info("Unlabeling node")
 
-		if err := pnmr.deleteLabel(ctx, nodeName, labelName); err != nil {
-			return ctrl.Result{}, fmt.Errorf("could not unlabel node %s: %v", nodeName, err)
+		// Make sure we don't already have a new running pod before unlabeling the node
+		labelSelector := client.MatchingLabels{constants.ModuleNameLabel: moduleName}
+		fieldSelector := client.MatchingFields{"spec.nodeName": nodeName}
+		var modulePodsList v1.PodList
+		err := pnmr.client.List(ctx, &modulePodsList, labelSelector, fieldSelector)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get list of all pods for module %s on node %s: %v", moduleName, nodeName, err)
+		}
+		var foundRunningPod bool
+		for _, p := range modulePodsList.Items {
+			if podutils.IsPodReady(&p) && p.DeletionTimestamp.IsZero() {
+				foundRunningPod = true
+				break
+			}
+		}
+		if !foundRunningPod {
+			if err := pnmr.deleteLabel(ctx, nodeName, labelName, deprecatedLabelName); err != nil {
+				return ctrl.Result{}, fmt.Errorf("could not unlabel node %s with label {%q, %q}: %v",
+					nodeName, labelName, deprecatedLabelName, err)
+			}
 		}
 
 		if !pod.DeletionTimestamp.IsZero() {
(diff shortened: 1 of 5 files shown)


qbarrand:
As a user, how can I use this label in a selector?

yevgeny-shnaidman:
> As a user, how can I use this label in a selector?

I think that we should discourage customers to use this label.  We need to update our ModuleStatus that will reflect all the data needed by the users. In case customers chooses to use this label, it is up to him to deal with the changes

ybettan:
@qbarrand I agree with @yevgeny-shnaidman, this label should not be part of the API IMO.
You can see how this labeled is used by the device plugin DS using `nodeAffinity` in this PR.

ybettan:
`nodeAffinity` was removed from the solution and I have also merged the adding of the namespace to the label to this PR.

@yevgeny-shnaidman @qbarrand PTAL.

controllers/pod_node_module_reconciler.go
```diff
@@ -70,8 +72,21 @@ func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if !podutils.IsPodReady(&pod) || !pod.DeletionTimestamp.IsZero() {
 		logger.Info("Unlabeling node")
 
-		if err := pnmr.deleteLabel(ctx, nodeName, labelName); err != nil {
-			return ctrl.Result{}, fmt.Errorf("could not unlabel node %s: %v", nodeName, err)
+		// Make sure we don't already have a new running pod before unlabeling the node
+		labelSelector := client.MatchingLabels{constants.ModuleNameLabel: moduleName}
```
yevgeny-shnaidman, line 76: we are supporting both label now, can't we add a pod name as a value to the new label? That way we won't need to query the pods on the node

controllers/pod_node_module_reconciler.go
```diff
@@ -70,8 +72,21 @@ func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		fieldSelector := client.MatchingFields{"spec.nodeName": nodeName}
+		var modulePodsList v1.PodList
+		err := pnmr.client.List(ctx, &modulePodsList, labelSelector, fieldSelector)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get list of all pods for module %s on node %s: %v", moduleName, nodeName, err)
+		}
+		if len(modulePodsList.Items) == 0 {
```
ybettan, line 83: We should also delete if there are non-ready pods or pods with deletionTimestamp set. Will change it.

controllers/pod_node_module_reconciler.go
ybettan: I prefer it as well but @qbarrand prefer to query the pods.

controllers/pod_node_module_reconciler.go
ybettan: Fixed.

controllers/pod_node_module_reconciler.go
```diff
@@ -70,8 +72,28 @@ func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+				break
+			}
+		}
+		if !foundRunningPod {
+			if err := pnmr.deleteLabel(ctx, nodeName, labelName); err != nil {
+				return ctrl.Result{}, fmt.Errorf("could not unlabel node %s with label %q: %v", nodeName, labelName, err)
+			}
+			if err := pnmr.deleteLabel(ctx, nodeName, deprecatedLabelName); err != nil {
+				return ctrl.Result{}, fmt.Errorf("could not unlabel node %s with label %q: %v", nodeName, deprecatedLabelName, err)
+			}
```
qbarrand, line 96: This applies two patches in a row. Could we apply only one with both labels?

controllers/pod_node_module_reconciler.go
```diff
@@ -94,6 +116,9 @@ func (pnmr *PodNodeModuleReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err := pnmr.addLabel(ctx, nodeName, labelName); err != nil {
 		return ctrl.Result{}, fmt.Errorf("could not label node %s with %q: %v", nodeName, labelName, err)
 	}
+	if err := pnmr.addLabel(ctx, nodeName, deprecatedLabelName); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not label node %s with %q: %v", nodeName, deprecatedLabelName, err)
+	}
```
qbarrand, line 121: Same comment about two patch operations where just one would be preferable.

controllers/pod_node_module_reconciler.go
qbarrand: IMO no need to expose the pod names, and I think the implementation in this PR looks great.

controllers/pod_node_module_reconciler.go
ybettan: Good point. Will try to merge them into one patch.

internal/daemonset/daemonset.go
```diff
@@ -36,7 +36,7 @@ type DaemonSetCreator interface {
 	GetModuleDaemonSets(ctx context.Context, name, namespace string) ([]appsv1.DaemonSet, error)
 	SetDriverContainerAsDesired(ctx context.Context, ds *appsv1.DaemonSet, mld *api.ModuleLoaderData) error
 	SetDevicePluginAsDesired(ctx context.Context, ds *appsv1.DaemonSet, mod *kmmv1beta1.Module) error
-	GetNodeLabelFromPod(pod *v1.Pod, moduleName string) string
+	GetNodeLabelFromPod(pod *v1.Pod, moduleName string, includeNamespace bool) string
```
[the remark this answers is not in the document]
ybettan: Fixed.

controllers/pod_node_module_reconciler.go
ybettan: Fixed.

controllers/pod_node_module_reconciler.go
ybettan: Fixed.

ybettan:
/unhold
We are also keeping the old label for backward compatibility.

ybettan:
/assign @ybettan 
(checking if prow bot is responding)

ybettan:
Looks like Prow's cluster has some issues - https://github.com/kubernetes/test-infra/issues/29003

Issue #263, opened 2023-01-23
Orderly kernel module version upgrade


yevgeny-shnaidman:
### **Issue summary**
Kernel module version upgrade must allow user to control the order of the pods(nodes) upgrade, and the timing of the upgrade of a specific pod (node). 

### **Current upgrade process**
Current upgrade process is as following:
1. User updates Module CR with the new image-container
2. KMM updates ModuleLoader DaemonSet with the new configuration
3. k8s ReplicaSets controller initiates a rolling upgrade of the DaemonSet, which means that the pods of the DaemonSet will get destroyed and recreated one by one, in the order defined by the controller.

Current flow causes the following difficulty:
1. User needs to terminate workload on a node prior to the ModuleLoader pod being destroyed (otherwise modprobe -r will fail)
2. User does not want to delete all the workloads on all the nodes ( downtime), but do it one by one
3. Since we don't know  the order of ModuleLoader rolling upgrade, user cannot  schedule the workload per node downtime correctly
4. In case part of the workload is running in the DevicePlugin,  the "modprobe -r" command will fail during Pod rollout, since the DevicePlugin pods are not rolled out

 ### **Proposed Solution**

1. Add a "Module Version" field to Module CRD. This field will be set by the user and updated whenever the ModuleLoader image is updated. Its value will be used by a customer to set _ModuleVersionLabel_ on the nodes ( i.e if the customer want to run version 2 of the Module on some node, he must label the node with _ModuleVersionLabel_=<_module version_>)
2. New labels: _ModuleLoaderVersionLabel_ and _DevicePluginVersionLabel_ will be derived from the "Module Version" field. Those labels will be used by KMMO exclusively for adding/removal of ModuleLoader/DevicePlugin pods on nodes.  
3. When the Module CR is updated with new "ModuleVersion" and the new "Image" fields, it will not update the existing ModuleLoader,  but will create a new ModuleLoader, with the new image and the new _ModuleLoaderVersionLabel_ as part of the NodeSelector.  The same goes for DevicePlugin: a new DaemonSet is created with the new _DevicePluginVersionLabel_ as part of Node Selector
4. Garbage collector will need to be updated to collect ModuleLoader DS with existing kernels but irrelevant module versions, and to collect the DevicePlugin DS with irrelevant module version

**Upgrade Flow example**
****Initiating Module****
1. User create a Module CR with image, module version and selector fields
[...]

PR #332, opened 2023-03-07
Refactoring of handleDriverContainer code

yevgeny-shnaidman:
This PR refactors the code of handleDrivercontainer code, and as a result also refactors the moduleReconciler loop, garbage collector and status module updator The main points:
1) change the name of the module loader DS name creation. Now the name consists of
   module.Name + hash(kernel version). This way we don't actually need a map of DS per
   kernel version, we can allow CreateOrPatch function to do its work, since now DS name
   is deterministic
2) as a result of 1), the API of handleDriverContainer, garbageCollector, statusUpdater
   and reconcile loop has been change
3) DaemonSet interface now returns just a list of DS for module 4) the differintiation between device-plugin and module-loader DS
   is now based on the role label
5) updating unit test

Part of [263](https://github.com/kubernetes-sigs/kernel-module-management/issues/263)


qbarrand:
@yevgeny-shnaidman the PR decreases the coverage, could you please fix that?

controllers/module_reconciler.go
```diff
@@ -340,21 +345,18 @@ func (mrh *moduleReconcilerHelper) handleSigning(ctx context.Context, mld *api.M
 }
 
 func (mrh *moduleReconcilerHelper) handleDriverContainer(ctx context.Context,
-	mld *api.ModuleLoaderData,
-	dsByKernelVersion map[string]*appsv1.DaemonSet) error {
+	mld *api.ModuleLoaderData) error {
+
+	hashValue, err := hashstructure.Hash(hashData{kernelVersion: mld.KernelVersion}, nil)
```
qbarrand, line 350: If we just hash one string, I don't see the benefit of using hashstructure. It is probably much more expensive than [fnv](https://pkg.go.dev/hash/fnv).

qbarrand:
What is the benefit of dropping the map and relying on a hash? I originally went with that map to eliminate the risk of any collision.

controllers/module_reconciler.go
yevgeny-shnaidman: i want to use hashstructure, since later on we will add a module Version to the hash

yevgeny-shnaidman:
> What is the benefit of dropping the map and relying on a hash? I originally went with that map to eliminate the risk of any collision.

hash is safe enough, especially since we are talking about DaemonSets  per module per namespace. The benefit is simplifying the code, eliminating double code, since CreateOrPatch check for DS existance and knows to take the existent DS and patch it. In addition, when we will work with ModuleVersion, we don't need to maintain map per kernel version per module version - we can just code module version as part of DS name and the same code will work

internal/daemonset/daemonset.go
```diff
@@ -52,12 +53,13 @@ func NewCreator(client client.Client, kernelLabel string, scheme *runtime.Scheme
...
-func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS map[string]*appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
+func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
 	deleted := make([]string, 0)
 
-	for kernelVersion, ds := range existingDS {
-		if !dc.isDevicePluginDaemonSet(ds) && !validKernels.Has(kernelVersion) {
-			if err := dc.client.Delete(ctx, ds); err != nil {
+	for _, ds := range existingDS {
+		dsKernelVersion := ds.Labels[dc.kernelLabel]
+		if !IsDevicePluginDS(&ds) && !validKernels.Has(dsKernelVersion) {
```
ybettan, line 61:
```suggestion
		
		if dsKernelVersion := ds.Labels[dc.kernelLabel]; !IsDevicePluginDS(&ds) && !validKernels.Has(dsKernelVersion) {
```

This variable is only scoped to the `if` block

qbarrand:
> In addition, when we will work with ModuleVersion, we don't need to maintain map per kernel version per module version

What is the cost of that? I think a struct with both the kernel and the module version can be a map key.

internal/daemonset/daemonset.go
yshnaidman: fixed

controllers/module_reconciler.go
qbarrand: Turns out hashstructure does use fnv.

controllers/module_reconciler.go
```diff
@@ -340,21 +345,18 @@ func (mrh *moduleReconcilerHelper) handleSigning(ctx context.Context, mld *api.M
...
+	if err != nil {
+		return fmt.Errorf("failed to hash kernel version %s for module loader daemonset name: %v", mld.KernelVersion, err)
+	}
+	name := mld.Name + "-" + fmt.Sprintf("%d", hashValue)
```
qbarrand, line 354: How about we use hex to same some characters? Also, let's do all the formatting in `fmt.Sprintf`:
```suggestion
	name := fmt.Sprintf("%s-%x", mld.Name, hashValue)
```

controllers/module_reconciler.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/sign"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/statusupdater"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
+	"github.com/mitchellh/hashstructure"
```
qbarrand, line 34: Let's use v2.

controllers/module_reconciler.go

controllers/module_reconciler.go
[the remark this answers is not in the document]

internal/statusupdater/statusupdater_test.go
```diff
@@ -256,25 +255,26 @@ var _ = Describe("preflight status updates", func() {
...
+	return ds
 }
 
-func prepareDsByKernel(dsConfigs []daemonSetConfig) map[string]*appsv1.DaemonSet {
-	dsMap := make(map[string]*appsv1.DaemonSet)
-	for i, dsConfig := range dsConfigs {
-		kernelVersion, ds := getDaemonSet(i, dsConfig)
-		dsMap[kernelVersion] = ds
+func prepareDsByKernel(dsConfigs []daemonSetConfig) []appsv1.DaemonSet {
+	dsSlice := []appsv1.DaemonSet{}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/module_reconciler_test.go
```diff
@@ -555,20 +556,18 @@ var _ = Describe("ModuleReconciler_handleDriverContainer", func() {
...
 			KernelVersion: "kernelVersion1",
 		}
-		existingDS := map[string]*appsv1.DaemonSet{
-			"kernelVersion2": &appsv1.DaemonSet{}, "kernelVersion3": &appsv1.DaemonSet{},
-		}
+		hashValue, err := hashstructure.Hash(hashData{kernelVersion: mld.KernelVersion}, nil)
+		Expect(err).NotTo(HaveOccurred())
 		newDS := &appsv1.DaemonSet{
-			ObjectMeta: metav1.ObjectMeta{Namespace: mld.Namespace},
+			ObjectMeta: metav1.ObjectMeta{Namespace: mld.Namespace, Name: mld.Name + "-" + fmt.Sprintf("%d", hashValue)},
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

PR #333, opened 2023-03-12
In module version label, default namespace should be explicit

yevgeny-shnaidman:
In case use does not define a namespace in its module, the Module and all its DaemonSets are created in the default namespace. When using module versioning labels, this namespace must be explictily stated with "default"

### DIFF
--- a/internal/utils/moduleversionlabels.go
+++ b/internal/utils/moduleversionlabels.go
@@ -8,15 +8,22 @@ import (
 )
 
 func GetModuleVersionLabelName(namespace, name string) string {
-	return fmt.Sprintf("%s.%s.%s", constants.ModuleVersionLabelPrefix, namespace, name)
+	return getVersionLabelName(constants.ModuleVersionLabelPrefix, namespace, name)
 }
 
 func GetModuleLoaderVersionLabelName(namespace, name string) string {
-	return fmt.Sprintf("%s.%s.%s", constants.ModuleLoaderVersionLabelPrefix, namespace, name)
+	return getVersionLabelName(constants.ModuleLoaderVersionLabelPrefix, namespace, name)
 }
 
 func GetDevicePluginVersionLabelName(namespace, name string) string {
-	return fmt.Sprintf("%s.%s.%s", constants.DevicePluginVersionLabelPrefix, namespace, name)
+	return getVersionLabelName(constants.DevicePluginVersionLabelPrefix, namespace, name)
+}
+
+func getVersionLabelName(labelPrefix, namespace, name string) string {
+	if namespace == "" {
+		namespace = "default"
+	}
+	return fmt.Sprintf("%s.%s.%s", labelPrefix, namespace, name)
 }
 
 func GetNamespaceNameFromVersionLabel(label string) (string, string, error) {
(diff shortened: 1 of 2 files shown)


internal/utils/moduleversionlabels.go
```diff
@@ -8,15 +8,22 @@ import (
...
 func GetDevicePluginVersionLabelName(namespace, name string) string {
-	return fmt.Sprintf("%s.%s.%s", constants.DevicePluginVersionLabelPrefix, namespace, name)
+	return getVersionLabelName(constants.DevicePluginVersionLabelPrefix, namespace, name)
+}
+
+func getVersionLabelName(labelPrefix, namespace, name string) string {
+	if namespace == "" {
+		namespace = "default"
+	}
+	return fmt.Sprintf("%s.%s.%s", labelPrefix, namespace, name)
```
qbarrand, line 26: WDYT about:
```suggestion
	return fmt.Sprintf("%s/%s.%s.%s", prefix, namespace, name, property)
```
Where `prefix` is `kmm.node.kubernetes.io`.
This groups all labels for a given module together then they are sorted alphabetically.
Currently we set `{prefix}/{moduleName}.{ready}`.

yevgeny-shnaidman: Not sure i understand the change you suggests: this module deals with module version labels. So, what do we get by setting the prefix explicitly? Currently it is part of the LabelPrefix defined in the constants.go.

qbarrand: I meant that I would have the property (e.g. `version-module`) after the `namespace.name` in the label.

ybettan:
@yevgeny-shnaidman This PR is redundant IMO. If a user doesn't specify the namespace k8s will set it to `default` automatically.

Example:
```
apiVersion: kmm.sigs.x-k8s.io/v1beta1
kind: Module
metadata:
  name: kmm-ci
spec:
  moduleLoader:
    container:
      modprobe:
        moduleName: kmm_ci_a
      kernelMappings:
        - regexp: '^.+$'
          containerImage: quay.io/ybettan/kmm-kmod:$KERNEL_FULL_VERSION
          registryTLS:
            insecure: true
  selector:
    kubernetes.io/hostname: kind-control-plane
```

And then:
```
ybettan kernel-module-management (pull-333) $ oc get module/kmm-ci -o yaml | yq '.metadata.namespace'
default

```

yevgeny-shnaidman:
> @yevgeny-shnaidman This PR is redundant IMO. If a user doesn't specify the namespace k8s will set it to `default` automatically.
> 
> Example:
> 
> ```
> apiVersion: kmm.sigs.x-k8s.io/v1beta1
> kind: Module
> metadata:
>   name: kmm-ci
> spec:
>   moduleLoader:
>     container:
>       modprobe:
>         moduleName: kmm_ci_a
>       kernelMappings:
>         - regexp: '^.+$'
>           containerImage: quay.io/ybettan/kmm-kmod:$KERNEL_FULL_VERSION
>           registryTLS:
>             insecure: true
>   selector:
>     kubernetes.io/hostname: kind-control-plane
> ```
> 
> And then:
> 
> ```
> ybettan kernel-module-management (pull-333) $ oc get module/kmm-ci -o yaml | yq '.metadata.namespace'
> default
> ```
> 
> So, if the consumer code gets the namespace from the user's label then the user would have set `default` in the label key. Otherwise, if we get the namespace from querying the module we will get `default` and if we get it from the `KernelMappingData` struct (which was not query by k8s) then I think we should set it to `default` there and get it for free in any other place in the code.

Ok, closing

PR #334, opened 2023-03-12
handling Version change in DevicePlugin and ModuleLoader DaemonSets

yevgeny-shnaidman:
This PR add handing of the Version of the module in the creation of DevicePlugin and ModuleLoader DaemonSets. The following was added: 
1) setting the Module's Version in the ModuleLoaderData
2) adding Version the hash of DevicePlugin/ModuleLoader daemonset's name
3) adding module version label to the labels of DevicePlugin/ModuleLoader daemonsets
4) changing DS garbage collection to remove DS that does not contain the current
   version label, and has no more available pods

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -192,6 +192,7 @@ type moduleReconcilerHelperAPI interface {
 
 type hashData struct {
 	KernelVersion string
+	ModuleVersion string
 }
 
 type moduleReconcilerHelper struct {
@@ -347,9 +348,13 @@ func (mrh *moduleReconcilerHelper) handleSigning(ctx context.Context, mld *api.M
 func (mrh *moduleReconcilerHelper) handleDriverContainer(ctx context.Context,
 	mld *api.ModuleLoaderData) error {
 
-	hashValue, err := hashstructure.Hash(hashData{KernelVersion: mld.KernelVersion}, hashstructure.FormatV2, nil)
+	dsNameData := hashData{
+		KernelVersion: mld.KernelVersion,
+		ModuleVersion: mld.ModuleVersion,
+	}
+	hashValue, err := hashstructure.Hash(dsNameData, hashstructure.FormatV2, nil)
 	if err != nil {
-		return fmt.Errorf("failed to hash kernel version %s for module loader daemonset name: %v", mld.KernelVersion, err)
+		return fmt.Errorf("failed to hash kernel and module versions (%s %s) and for module loader daemonset name: %v", mld.KernelVersion, mld.ModuleVersion, err)
 	}
 	name := fmt.Sprintf("%s-%x", mld.Name, hashValue)
 	ds := &appsv1.DaemonSet{
@@ -374,8 +379,13 @@ func (mrh *moduleReconcilerHelper) handleDevicePlugin(ctx context.Context, mod *
 	}
 
 	logger := log.FromContext(ctx)
+	hashValue, err := hashstructure.Hash(hashData{ModuleVersion: mod.Spec.ModuleLoader.Container.Version}, hashstructure.FormatV2, nil)
+	if err != nil {
+		return fmt.Errorf("failed to hash module version %s for device-plugin daemonset name: %v", mod.Spec.ModuleLoader.Container.Version, err)
+	}
+	name := fmt.Sprintf("%s-device-plugin-%x", mod.Name, hashValue)
 	ds := &appsv1.DaemonSet{
-		ObjectMeta: metav1.ObjectMeta{Name: mod.Name + "-device-plugin", Namespace: mod.Namespace},
+		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: mod.Namespace},
 	}
 
 	opRes, err := controllerutil.CreateOrPatch(ctx, mrh.client, ds, func() error {
@@ -397,7 +407,7 @@ func (mrh *moduleReconcilerHelper) garbageCollect(ctx context.Context,
 	// Garbage collect old DaemonSets for which there are no nodes.
 	validKernels := sets.KeySet[string](mldMappings)
 
-	deleted, err := mrh.daemonAPI.GarbageCollect(ctx, existingDS, validKernels)
+	deleted, err := mrh.daemonAPI.GarbageCollect(ctx, mod, existingDS, validKernels)
 	if err != nil {
 		return fmt.Errorf("could not garbage collect DaemonSets: %v", err)
 	}
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -9,6 +9,7 @@ import (
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/api"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -32,7 +33,7 @@ const (
 //go:generate mockgen -source=daemonset.go -package=daemonset -destination=mock_daemonset.go
 
 type DaemonSetCreator interface {
-	GarbageCollect(ctx context.Context, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error)
+	GarbageCollect(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error)
 	GetModuleDaemonSets(ctx context.Context, name, namespace string) ([]appsv1.DaemonSet, error)
 	SetDriverContainerAsDesired(ctx context.Context, ds *appsv1.DaemonSet, mld *api.ModuleLoaderData) error
 	SetDevicePluginAsDesired(ctx context.Context, ds *appsv1.DaemonSet, mod *kmmv1beta1.Module) error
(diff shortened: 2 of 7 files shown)


ybettan:
Part of https://github.com/kubernetes-sigs/kernel-module-management/issues/263

controllers/module_reconciler.go
```diff
@@ -374,8 +379,13 @@ func (mrh *moduleReconcilerHelper) handleDevicePlugin(ctx context.Context, mod *
 	}
 
 	logger := log.FromContext(ctx)
+	hashValue, err := hashstructure.Hash(hashData{ModuleVersion: mod.Spec.ModuleLoader.Container.Version}, hashstructure.FormatV2, nil)
```
ybettan, line 382: Why don't we hash the kernelVersion here as well?

controllers/module_reconciler.go
```diff
@@ -347,9 +348,13 @@ func (mrh *moduleReconcilerHelper) handleSigning(ctx context.Context, mld *api.M
 func (mrh *moduleReconcilerHelper) handleDriverContainer(ctx context.Context,
 	mld *api.ModuleLoaderData) error {
 
-	hashValue, err := hashstructure.Hash(hashData{KernelVersion: mld.KernelVersion}, hashstructure.FormatV2, nil)
+	dsNameData := hashData{
+		KernelVersion: mld.KernelVersion,
+		ModuleVersion: mld.ModuleVersion,
+	}
+	hashValue, err := hashstructure.Hash(dsNameData, hashstructure.FormatV2, nil)
```
ybettan, line 355: Why don't we garbage collect `moduleLoader`'s daemonsets as well?

internal/daemonset/daemonset.go
```diff
@@ -32,7 +33,7 @@ const (
 //go:generate mockgen -source=daemonset.go -package=daemonset -destination=mock_daemonset.go
 
 type DaemonSetCreator interface {
-	GarbageCollect(ctx context.Context, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error)
+	GarbageCollect(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error)
```
ybettan, line 36: If we also need to garbage collect moduleLoader's DS so how does it work? moduleLoader is using a `moduleLoaderData`.

controllers/module_reconciler.go
```diff
@@ -397,7 +407,7 @@ func (mrh *moduleReconcilerHelper) garbageCollect(ctx context.Context,
 	// Garbage collect old DaemonSets for which there are no nodes.
 	validKernels := sets.KeySet[string](mldMappings)
 
-	deleted, err := mrh.daemonAPI.GarbageCollect(ctx, existingDS, validKernels)
+	deleted, err := mrh.daemonAPI.GarbageCollect(ctx, mod, existingDS, validKernels)
```
ybettan, line 410: This is also deleting moduleLoader's DSs. I don't think we should do it inside `handelDevicePlugin` method.

internal/daemonset/daemonset.go
```diff
@@ -53,10 +54,20 @@ func NewCreator(client client.Client, kernelLabel string, scheme *runtime.Scheme
 	}
 }
 
-func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
+func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
```
ybettan, line 57: I would add a new `daemonSetType` with possible values `daemonSetTypeModuleLoader` and `daemonsetTypeDevicePlugin` to be passed to this function and add re-orgenize this function as a switch-case.

internal/daemonset/daemonset.go
```diff
@@ -92,6 +103,10 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(
 		dc.kernelLabel:            kernelVersion,
 		constants.DaemonSetRole:   moduleLoaderRoleLabelValue,
 	}
+	if mld.ModuleVersion != "" {
```
ybettan, line 106: Can we move this logic to https://github.com/kubernetes-sigs/kernel-module-management/blob/bc488304f926742c14b0837563ef63acfef2b158/internal/module/kernelmapper.go#L118? I think that the namespace should be set to `default` instead of ` ` as soon as possible in the code.

internal/daemonset/daemonset.go
```diff
@@ -230,6 +245,10 @@ func (dc *daemonSetGenerator) SetDevicePluginAsDesired(
 		constants.ModuleNameLabel: mod.Name,
 		constants.DaemonSetRole:   devicePluginRoleLabelValue,
 	}
+	if mod.Spec.ModuleLoader.Container.Version != "" {
```
ybettan, line 248: I am wondering if `moduleLoaderData` should implement a `moduleData` interface and then we can pass to this method (and others) a `moduleData` instead of using `Module` for devicePlugin and `modueDataLoader` for driverContainer.

internal/module/kernelmapper.go
```diff
@@ -120,6 +120,7 @@ func (kh *kernelMapperHelper) prepareModuleLoaderData(mapping *kmmv1beta1.Kernel
 	mld.Selector = mod.Spec.Selector
 	mld.ServiceAccountName = mod.Spec.ModuleLoader.ServiceAccountName
 	mld.Modprobe = mod.Spec.ModuleLoader.Container.Modprobe
+	mld.ModuleVersion = mod.Spec.ModuleLoader.Container.Version
```
ybettan, line 123: Should we set a default value in case it was not set?

controllers/module_reconciler.go
yevgeny-shnaidman: no need, device plugin is independent of kernel version

controllers/module_reconciler.go
yevgeny-shnaidman: This code does not deal with Garbage collection, only with creation of the Daemonsets

internal/daemonset/daemonset.go
yevgeny-shnaidman: moduleLoader is just a daemonset, it does not use moduleLoaderData.  All the data needed for collecting the device-plugin and module loader are present in the Module (name, namespace, version), so that's all we need.

controllers/module_reconciler.go
yevgeny-shnaidman: i think there is confusion here, we are doing it inside garbageCollection function

internal/daemonset/daemonset.go
yevgeny-shnaidman: There is a common case for DS deletion (former version and no running pods), and a private case (invalid kernel version). The common case is valid for both ModuleLoader and Device Plugin , and private case is valid only for ModuleLoader. So, adding a switch case will result in going over all DS and creating to slices (one for DevicePlugin and one for ModuleLoader. I don't think this will improve the current code

internal/daemonset/daemonset.go
yevgeny-shnaidman: Setting the Version to default 1, will eventually require us or user to create a label on the nodes. This way , customers who don't want/need  to use ordered upgrade functionality can entirely skip it

internal/daemonset/daemonset.go
yevgeny-shnaidman: i am not sure if it worth the effort. In any case, if we decide to do it , it should be outside of this PR. You can open a Issue on that and we can discuss it there

internal/module/kernelmapper.go
yevgeny-shnaidman: This is something that i discussed with @qbarrand in previous PR, but it seems that currently his suggestion is better, since it allows customer who does not want to use ordered upgrade to completely ignore it

controllers/module_reconciler.go
qbarrand: Device plugin is kernel-independent.

internal/daemonset/daemonset.go
qbarrand: Or two distinct methods.

internal/daemonset/daemonset.go
```diff
@@ -53,10 +54,20 @@ func NewCreator(client client.Client, kernelLabel string, scheme *runtime.Scheme
...
 }
 
-func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
+func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
 	deleted := make([]string, 0)
 
+	versionLabel := utils.GetModuleVersionLabelName(mod.Namespace, mod.Name)
 	for _, ds := range existingDS {
+		// delete inactive device-plugin or module-loader DS with former label
+		if ds.Labels[versionLabel] != mod.Spec.ModuleLoader.Container.Version && ds.Status.NumberAvailable == 0 {
```
qbarrand, line 63: Should we instead check the number of desired pods?

yevgeny-shnaidman: both are ok, but "desired" is better and cleaner.
Fixed

controllers/module_reconciler.go
ybettan: I don't know why I though it was added to `handelDevicePlugin`... Please ignore this comment.

internal/daemonset/daemonset.go
ybettan: This comment was based on a previous one (and misslooking this PR) - https://github.com/kubernetes-sigs/kernel-module-management/pull/334/files#r1135481519.

internal/daemonset/daemonset.go
yevgeny-shnaidman: added a more distinct functions instead of ifs

internal/daemonset/daemonset.go
```diff
@@ -53,11 +54,21 @@ func NewCreator(client client.Client, kernelLabel string, scheme *runtime.Scheme
...
+		// delete inactive device-plugin or module-loader DS with former label
+		if isOlderVersionUnusedDaemonset(&ds, versionLabel, mod.Spec.ModuleLoader.Container.Version) {
+			if err := dc.client.Delete(ctx, &ds); err != nil {
+				return nil, fmt.Errorf("could not delete DaemonSet %s: %v", ds.Name, err)
+			}
+			deleted = append(deleted, ds.Name)
+			continue
+		}
+		// delete module-loader ds with invalid kernel mapping
+		if isModuleLoaderDaemonsetWithInvalidKernel(&ds, dc.kernelLabel, validKernels) {
```
ybettan, line 71: Any reason not to use `else if`?
```suggestion
		if isOlderVersionUnusedDaemonset(&ds, versionLabel, mod.Spec.ModuleLoader.Container.Version) {
			if err := dc.client.Delete(ctx, &ds); err != nil {
				return nil, fmt.Errorf("could not delete DaemonSet %s: %v", ds.Name, err)
			}
			deleted = append(deleted, ds.Name)
		} else if isModuleLoaderDaemonsetWithInvalidKernel(&ds, dc.kernelLabel, validKernels) {
```

internal/daemonset/daemonset.go
```diff
@@ -53,11 +54,21 @@ func NewCreator(client client.Client, kernelLabel string, scheme *runtime.Scheme
...
 }
 
-func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
+func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet, validKernels sets.Set[string]) ([]string, error) {
 	deleted := make([]string, 0)
 
+	versionLabel := utils.GetModuleVersionLabelName(mod.Namespace, mod.Name)
 	for _, ds := range existingDS {
-		if !IsDevicePluginDS(&ds) && !validKernels.Has(ds.Labels[dc.kernelLabel]) {
+		// delete inactive device-plugin or module-loader DS with former label
```
ybettan, line 62: Code is self explanatory now, I don't think we need this comment anymore.

internal/daemonset/daemonset.go
yevgeny-shnaidman: basically i would prefer a code with continue, and without else, to me it seems cleaner

internal/daemonset/daemonset.go
yevgeny-shnaidman: fixed

ybettan:
/lgtm
/hold
@qbarrand unhold if you are happy with the changes.

PR #347, opened 2023-03-20
Module version nodes' label management

yevgeny-shnaidman:
This PR adds the management of the a module version labels management per node. It is done in a dedicated controller, which listens on the changes of the module version labels on the nodes. Once such a change is detected, the reconciliation flow is as following: 1) get all the module version labels of the node (module version,
   module_loader version, device-plugin version)
2) sort the labels per their module-name/namespace. Result: slice
   of structure (one structure per module), that holds the values
   of module/module-loader/device-lugin labels for that particular module
3) Per each module, based on a static action-table, determine what is the next
   action should be: none, adding label, deleting label. Per module it will always
   be only one label that will either be added or deleted
4) Gather all labels to be added/deleted for all the modules and update node in
   one operation

Chnages in code:
1) new controller
2) static action table for that controller
3) adding filter function to trigger the controller only on changes
   to module version labels
4) mocks and unit-tests


controllers/module_version_label_action_table.go
```diff
@@ -0,0 +1,66 @@
...
+import "github.com/kubernetes-sigs/kernel-module-management/internal/utils"
+
+const (
+	labelMissing   = "missing"
+	labelPresent   = "present"
+	labelDifferent = "different"
+
+	addAction    = "add"
+	deleteAction = "delete"
+	noneAction   = "none"
```
ybettan, line 12:
```suggestion
	moduleVersionLabelMissing   = "missing"
	moduleVersionLabelPresent   = "present"
	moduleVersionLabelDifferent = "different"

	moduleVersionActionAdd    = "add"
	moduleVersionActionDelete = "delete"
	moduleVersionActionNone   = "none"
```

ybettan: Also should we create new types for it?
```
type moduleVersionStatus string
type moduleVersionLabelAction string
```
?

yevgeny-shnaidman:
> Don't we need some code for the hub-spoke topology as well?
> 
> If there is a build section, then bumping the version in the module will trigger a new build in the hub while the node's labels are on the spoke.

bumping the version alone won't do anything. User needs to bump both version and image, and only then will the build be triggered.

controllers/module_version_label_action_table.go
```diff
@@ -0,0 +1,66 @@
...
+
+	labelActionKey{
+		module:       labelPresent,
+		moduleLoader: labelDifferent,
+		devicePlugin: labelDifferent}: labelAction{getLabelName: utils.GetDevicePluginVersionLabelName, action: deleteAction},
+
+	labelActionKey{
+		module:       labelPresent,
+		moduleLoader: labelDifferent,
+		devicePlugin: labelMissing}: labelAction{getLabelName: utils.GetModuleLoaderVersionLabelName, action: deleteAction},
```
ybettan, line 65: It means that the user has overwritten the version label right?! So the new DS will be deploy in parallel with removing the old labels?

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+const (
+	NodeLabelModuleVersionReconcilerName = "NodeLabelModuleVersion"
+)
+
+type reconcileLabelsResult struct {
+	labelsToAdd    map[string]string
+	labelsToDelete []string
+	requeue        bool
```
ybettan, line 36: Why do we need the `requeue`?

controllers/module_version_label_action_table.go
yevgeny-shnaidman: We have 3 type of version labels: module, moduleLoader and DevicePlugin.  This const just defines the state of any one of those labels. It might even be used for other labels also, if need be
I ma not sure that we need to redefine the string here, i don't see the benefits of doing so

controllers/module_version_label_action_table.go
yevgeny-shnaidman: yes, correct , user has set the new value to its version. So the action for this state is to just remove the label of the module-loader, which will cause the removal of the pod. new DS pod will not be deployed since we have not added the module-loader label with new version yet

controllers/module_version_label_action_table.go
ybettan: > This const just defines the state of any one of those labels

The prefix `moduleVersionLabel` isn't because it is for the module label but for all labels related to the versioning feature.

The convention in go in such cases is to (basically this is the golang way of using `enum`) define a new type, define consts to contain all the possible values of this type and you can even define methods to this type in the future if we want to extend it.

I would go with something like:
```
type moduleVersionLabelStatus int
const (
    moduleVersionLabelStatusMissing moduleVersionLabelStatus = iota
    moduleVersionLabelStatusPresent
    moduleVersionLabelStatusDifferent
)
```

And a similar flow for `moduleVersionLabelAction` type.

controllers/module_version_label_action_table.go
ybettan: > new DS pod will not be deployed since we have not added the module-loader label with new version yet

And this will automatically be triggered by the action in index 4 in the table once the labels are deleted right?!

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+func (nlmvr *NodeLabelModuleVersionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	node := v1.Node{}
+
+	if err := nlmvr.client.Get(ctx, types.NamespacedName{Name: req.Name}, &node); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get node: %v", err)
+	}
+
+	logger := log.FromContext(ctx).WithValues("node name", node.Name)
+
+	// get node's label data per module
```
ybettan, line 60: Can we remove this comment, the code is self explanatory.

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+	modulesVersionLabels := nlmvr.helperAPI.getLabelsPerModules(node.Labels)
+
+	// get module loader pods on the node
+	moduleLoaderPods, err := nlmvr.helperAPI.getModuleLoaderPods(ctx, node.Name)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("could get module loader pods for the node %s: %v", node.Name, err)
+	}
+
+	// get the labels to add and delete for this specific node
```
ybettan, line 69: Can we remove this comment, the code is self explanatory.

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+	reconLabelsRes := nlmvr.helperAPI.reconcileLabels(modulesVersionLabels, moduleLoaderPods)
+
+	logLabelsUpdateData(logger, reconLabelsRes)
+
+	// update the node with the labels
```
ybettan, line 74: Can we remove this comment, the code is self explanatory.

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+	return &nodeLabelModuleVersionHelper{client: client}
+}
+
+func (nlmvha *nodeLabelModuleVersionHelper) getLabelsPerModules(nodeLabels map[string]string) map[string]*modulesVersionLabels {
+	labelsPerModule := map[string]*modulesVersionLabels{}
+	for key, value := range nodeLabels {
+		if utils.IsVersionLabel(key) {
+			namespace, name, err := utils.GetNamespaceNameFromVersionLabel(key)
+			if err != nil {
+				continue
```
ybettan, line 106: Why are we skipping here? Why don't we return an error or at least a warning log?

controllers/module_version_label_action_table.go
yevgeny-shnaidman: yes, change in the module version labels triggers the reconciliation loop

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+
+			if utils.IsModuleVersionLabel(key) {
+				labelsPerModule[mapKey].moduleVersionLabel = value
+			}
+			if utils.IsModuleLoaderVersionLabel(key) {
+				labelsPerModule[mapKey].moduleLoaderVersionLabel = value
+			}
+			if utils.IsDevicePluginVersionLabel(key) {
+				labelsPerModule[mapKey].devicePluginVersionLabel = value
+			}
```
ybettan, line 121:
```suggestion
			if utils.IsModuleVersionLabel(key) {
				labelsPerModule[mapKey].moduleVersionLabel = value
			} else if utils.IsModuleLoaderVersionLabel(key) {
				labelsPerModule[mapKey].moduleLoaderVersionLabel = value
			} else if utils.IsDevicePluginVersionLabel(key) {
				labelsPerModule[mapKey].devicePluginVersionLabel = value
			}
```

Just for clarity to show that a label cannot be true on more than 1 entry.

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: fixed

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: fixed

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: fixed

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: IMHO "if" without "else" is a much cleaner and readable code. The names of the functions are a clear indication of which label we are checking.

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+	node := v1.Node{}
+
+	if err := nlmvha.client.Get(ctx, types.NamespacedName{Name: nodeName}, &node); err != nil {
+		return fmt.Errorf("could not get node %s: %v", nodeName, err)
+	}
+
+	nodeCopy := node.DeepCopy()
+
+	if node.Labels == nil && len(reconLabelsRes.labelsToAdd) > 0 {
+		node.Labels = make(map[string]string, 1)
```
ybettan, line 149:
```suggestion
		node.Labels = map[string]string{}
```

I don't see the point of setting the initial map size to 1 when we don't know how many labels we are going to add. There is no logical reason to set the value to 1.

Am I missing something?

controllers/module_version_label_action_table.go
yevgeny-shnaidman: 1) the problem is that we have moduleVersionLabel, so moduleVersionLabelAciotn... might imply that it is relevant only for moduleVersion and not for, say, ModuleLoaderVersion or DevicePluginVersion
2) I think that the main reason for type definition is to enforce type checking and add additional methods on the type. In this case we need neither, so why do it? We have additional places in the ocde where we have constants without type redefinition

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+	for _, pod := range moduleLoaderPods {
+		podLabels := pod.GetLabels()
+		if pod.Namespace == namespace && podLabels[constants.ModuleNameLabel] == name {
+			return true
+		}
+	}
+	return false
+}
+
+// returns the label, value of the label and action to execute on label (add or delete)
```
ybettan, line 197: I would remove this comment as well. At some point the code will change and the comment will be forgotten and left inaccurate.

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+func getLabelAndAction(moduleLabels *modulesVersionLabels) (string, string, string) {
+	labelActionKey := getModuleVersionLabelsState(moduleLabels)
+	labelAction, ok := labelActionTable[labelActionKey]
+	if !ok || labelAction.action == noneAction {
+		return "", "", noneAction
+	}
```
ybettan, line 203:
```suggestion
	if !ok {
		return "", "", noneAction
	}
```

The second condition is covered in the switch-case.

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: fixed

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: i would prefer to leave, since it states clearly what is the order of the returned strings. In case the function is changed, then the comment should be changed as well.

internal/daemonset/daemonset.go
```diff
@@ -27,7 +27,7 @@ const (
 	nodeVarLibFirmwarePath         = "/var/lib/firmware"
 	nodeVarLibFirmwareVolumeName   = "node-var-lib-firmware"
 	devicePluginRoleLabelValue     = "device-plugin"
-	moduleLoaderRoleLabelValue     = "module-loader"
+	//moduleLoaderRoleLabelValue     = "module-loader"
```
ybettan, line 30: Forgotten

controllers/module_version_label_action_table.go
ybettan: > I think that the main reason for type definition is to enforce type checking

Why not getting it for free? I won't insist but I just think it is much cleaner and you can't pass any string you want but well pre-defined values.

controllers/node_label_module_version_reconciler.go
ybettan: I may think that a label can be both a moduleVersionLabel and also a moduleLoaderVersionLabel. If I wish to understand if multiple cases are possible I will need to check the functions content while having an `else` make it clear that each of of them can only fit a single entry.

controllers/node_label_module_version_reconciler.go
```diff
@@ -0,0 +1,257 @@
...
+	if len(reconLabelsRes.labelsToAdd) != 0 {
+		logger.Info("found labels to add", "addLabels", reconLabelsRes.labelsToAdd)
+	}
+	if len(reconLabelsRes.labelsToDelete) != 0 {
+		logger.Info("found labels to delete", "deleteLabels", reconLabelsRes.labelsToDelete)
+	}
+}
+
+// SetupWithManager sets up the controller with the Manager.
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: changed to switch

ybettan:
/lgtm
/hold
@qbarrand Unhold if you are happy with the changes.

Issue #89, opened 2022-09-27
Add a validating webhook


qbarrand:
Some complex constraints cannot be expressed through kubebuilder / OpenAPI annotations.
To avoid errors at runtime, non-trivial validations are typically implemented in a [Validating Admission Webhook](https://book.kubebuilder.io/reference/admission-webhook.html). We should implement such a webhook to detect malformed `Modules` as early as possible.

**Checks:**
- [ ] all regexes in the `Module` are valid
- [ ] all kernel mappings should have a non-empty `containerImage` field, unless `.spec.moduleLoader.container.containerImage` is defined
- [ ] `.spec.moduleLoader.container.modprobe.moduleName` is not empty unless both `modprobe.rawArgs.load` and `modprobe.rawArgs.unload` are defined

yevgeny-shnaidman:
in case RawArgs are set, webhooks must make sure that -a option is not set ( -a option is not support by KMM)

qbarrand:
Merged in #348 #368 #383 #388

PR #348, opened 2023-03-21
Create a validation webhook

erusso7:
Job done:

- Generated all the required files using `operator-sdk` to enable the webhook for the operator. 
- Implemented the first validation using the `Create` and `Update` webhook
- Implement other validations to meet the #89 issue's requirements. 
- Squash commits before the final PR

Pending to do:
- Adapt this code to make it work downstream


cmd/manager/main.go
```diff
@@ -178,6 +178,10 @@ func main() {
 		}
 	}
 
+	if err = (&v1beta12.Module{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLogger.Error(err, "unable to create webhook", "webhook", "Module")
+		os.Exit(1)
```
ybettan, line 183: How about using `cmd.FatalError` here?

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,80 @@
...
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Module) ValidateCreate() error {
+	modulelog.Info("validate create", "name", r.Name)
+
+	if err := r.validateKernelMappingRegex(); err != nil {
+		return err
+	}
+
+	return nil
```
ybettan, line 49:
```suggestion
	return r.validateKernelMappingRegex()
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,80 @@
...
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Module) ValidateUpdate(old runtime.Object) error {
+	modulelog.Info("validate update", "name", r.Name)
+
+	if err := r.validateKernelMappingRegex(); err != nil {
+		return err
+	}
+
+	return nil
```
ybettan, line 60:
```suggestion
	return r.validateKernelMappingRegex()

```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,80 @@
...
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *Module) ValidateDelete() error {
+	return nil
+}
+
+func (r *Module) validateKernelMappingRegex() error {
+	if r.Spec.ModuleLoader.Container.KernelMappings == nil {
+		return nil
+	}
+
```
qbarrand, line 64:
```suggestion
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,80 @@
...
+	for _, km := range r.Spec.ModuleLoader.Container.KernelMappings {
+		if _, rErr := regexp.Compile(km.Regexp); rErr != nil {
+			return errors.New("invalid regex: " + rErr.Error())
+		}
+	}
```
qbarrand:
```suggestion
	for idx, km := range r.Spec.ModuleLoader.Container.KernelMappings {
		if km.Regexp == "" {
			continue
		}
	
		if _, rErr := regexp.Compile(km.Regexp); rErr != nil {
			return fmt.Errorf("invalid regex at index %d: %v", idx, err)
		}
	}
```

config/webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,25 @@
...
+    path: webhooks/clientConfig/service/name
+  - kind: ValidatingWebhookConfiguration
+    group: admissionregistration.k8s.io
+    path: webhooks/clientConfig/service/name
+
+namespace:
+- kind: MutatingWebhookConfiguration
+  group: admissionregistration.k8s.io
+  path: webhooks/clientConfig/service/namespace
+  create: true
```
qbarrand, line 18:
```suggestion
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (r *Module) validateKernelMappingRegex() error {
+	if r.Spec.ModuleLoader.Container.KernelMappings == nil {
+		return nil
+	}
+
+	for _, km := range r.Spec.ModuleLoader.Container.KernelMappings {
+		if _, rErr := regexp.Compile(km.Regexp); rErr != nil {
+			return errors.New("invalid regex: " + rErr.Error())
+		}
```
ybettan, line 76:
```suggestion
		if _, err := regexp.Compile(km.Regexp); err != nil {
			return fmt.Errorf("invalid regex: %v", err)
		}
```

api/v1beta1/module_webhook.go
erusso7: For the moment, I will reject this suggestion because the plan is to have a "pipeline" of at least 3 validations, and this refactor will be done at the last validation.

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,79 @@
...
+			Container: ModuleLoaderContainerSpec{
+				KernelMappings: nil,
+			},
+		},
+	},
+}
+
+var _ = BeforeEach(func() {
+	module.Spec.ModuleLoader.Container.KernelMappings = validKMs
+})
```
ybettan, line 50: I will suggest not having a global `BeforeEach`.

You can have 1 BeforeEach per spec (`It`) if you wish if it makes sense to have it. Otherwise I will just defined a module in each `Describe` or `It` separately.

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+var module = &Module{
+	Spec: ModuleSpec{
+		ModuleLoader: ModuleLoaderSpec{
+			Container: ModuleLoaderContainerSpec{
+				KernelMappings: nil,
+			},
+		},
+	},
+}
```
ybettan, line 46: Can we please define those in each `Describe`/`It` accordingly instead of having global variables?
ATM you have them globally and override them in some `It`. This can become very messy when the test file is starting to grow.

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,79 @@
...
+		e := module.ValidateCreate()
+		Expect(e).To(BeNil())
+	})
+
+	It("should fail when an invalid regex is found", func() {
+		m := module
+		m.Spec.ModuleLoader.Container.KernelMappings = invalidKMs
+		e := module.ValidateCreate()
+		Expect(e).To(HaveOccurred())
+		Expect(e.Error()).To(ContainSubstring("invalid regex:"))
```
ybettan, line 63:
```suggestion
		Expect(e.Error()).To(ContainSubstring("invalid regexp"))
```

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,79 @@
...
+	})
+})
+
+var _ = Describe("ValidateUpdate", func() {
+	It("should pass when all conditions are met", func() {
```
ybettan, line 68: We should also have a test in case one of the regexp isn't valid.

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,79 @@
...
+		e := module.ValidateUpdate(nil)
+		Expect(e).To(BeNil())
+	})
+})
+
+var _ = Describe("ValidateDelete", func() {
+	It("should do nothing and return always nil", func() {
+		e := module.ValidateDelete()
+		Expect(e).To(BeNil())
```
ybettan, line 77: Same comment as before.

api/v1beta1/module_webhook_test.go
erusso7: Done :heavy_check_mark: but it looks like very repetitive code because we are modifying a deep property of the `Module` struct.

qbarrand (commented):
I think we should also [declare a dependency](https://olm.operatorframework.io/docs/concepts/olm-architecture/dependency-resolution/#declaring-dependencies) on cert-manager.

api/v1beta1/module_webhook.go
erusso7: This check is to prevent the `KernelMappings` is a nil point, otherwise, it will panic. I understand that from the use-case point of view, it will never be `nil`, should we remove it in that case?

api/v1beta1/module_webhook.go
erusso7: An empty regexp is valid and will pass the `Compile` check, should we split that and do redundant checks?

I'll add the index to explain better the error to the user :+1:

api/v1beta1/module_webhook.go
ybettan: `KernelMappings` is defined as a list of `KernelMapping`s in the [API](https://github.com/kubernetes-sigs/kernel-module-management/blob/50a9d36c7ec21cad4d201af40d6d0a96e0db8386/api/v1beta1/module_types.go#L189). Which means that if it wasn't define in the Module, it will just be any empty list (and not `nil`).

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+//+kubebuilder:webhook:path=/validate-kmm-sigs-x-k8s-io-v1beta1-module,mutating=false,failurePolicy=fail,sideEffects=None,groups=kmm.sigs.x-k8s.io,resources=modules,verbs=create;update,versions=v1beta1,name=vmodule.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &Module{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Module) ValidateCreate() error {
+	modulelog.Info("validate create", "name", r.Name)
+
+	if err := r.validateKernelMappingRegex(); err != nil {
+		return err
```
ybettan, line 46:
```suggestion
		return fmt.Errorf("failed to validate the kernel mapping regexps: %v", err)
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return r.validateModuleName()
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Module) ValidateUpdate(_ runtime.Object) error {
+	modulelog.Info("validate update", "name", r.Name)
+
+	if err := r.validateKernelMappingRegex(); err != nil {
+		return err
```
ybettan, line 57:
```suggestion
		return fmt.Errorf("failed to validate the kernel mapping regexps: %v", err)
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	container := r.Spec.ModuleLoader.Container
+
+	for idx, km := range container.KernelMappings {
+		if _, err := regexp.Compile(km.Regexp); err != nil {
+			return fmt.Errorf("invalid regexp at index %d: %v", idx, err)
+		}
+
+		if len(container.ContainerImage) == 0 && len(km.ContainerImage) == 0 {
```
ybettan, line 80:
```suggestion
		if lcontainer.ContainerImage == "" && km.ContainerImage == "" {
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+			return fmt.Errorf("missing spec.moduleLoader.container.kernelMappings[].containerImage")
```
ybettan, line 81:
```suggestion
			return fmt.Errorf("missing spec.moduleLoader.container.kernelMappings[%d].containerImage", idx)
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *Module) validateModuleName() error {
```
ybettan, line 88:
```suggestion
func (r *Module) validateModprobe() error {
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+	"regexp"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var modulelog = logf.Log.WithName("module-resource")
+
+func (r *Module) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
ybettan, line 31:
```suggestion
func (m *Module) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
Was this auto-generated by `operator-sdk`? If not, can we please change all the modules names to `m` instead of `r`?

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+
+	return nil
+}
+
+func (r *Module) validateModuleName() error {
+	modprobe := r.Spec.ModuleLoader.Container.Modprobe
+
+	if len(modprobe.ModuleName) > 0 {
```
ybettan, line 91:
```suggestion
	if modprobe.ModuleName != "" {
```

Also, what happens if both `modprobe.ModuleName` and `modprobe.RawArgs` are set? We should return an error in this case aren't we?

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil
+	}
+
+	if modprobe.RawArgs == nil {
+		return fmt.Errorf("missing spec.moduleLoader.container.modprobe.moduleName")
```
ybettan, line 96:
```suggestion
		return errors.New("missing spec.moduleLoader.container.modprobe.rawArgs and spec.moduleLoader.container.modprobe.moduleName isn't set")
```

When you don't need any error formatting, it is better to use `errors.New` as it is much faster.

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	if len(modprobe.RawArgs.Load) == 0 {
+		return fmt.Errorf("missing spec.moduleLoader.container.modprobe.rawArgs.load")
```
ybettan, line 100:
```suggestion
		return errors.New("missing spec.moduleLoader.container.modprobe.rawArgs.load")
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	if len(modprobe.RawArgs.Unload) == 0 {
+		return fmt.Errorf("missing spec.moduleLoader.container.modprobe.rawArgs.unload")
+	}
```
ybettan, line 105:
```suggestion
	if len(modprobe.RawArgs.Load) == 0 || len(modprobe.RawArgs.Unload) == 0 {
		return fmt.Errorf("when using spec.moduleLoader.container.modprobe.rawArgs, both load and unload fields should be specified")
	}
```

I think that if you are using rawArgs you must specify both:
1. How to load
2. How to unload

@qbarrand Please keep me honest here.

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+	})
+
+	It("should fail when a kernel-mapping has invalid containerName", func() {
+		mod := &Module{
+			Spec: ModuleSpec{
+				ModuleLoader: ModuleLoaderSpec{
+					Container: ModuleLoaderContainerSpec{
+						Modprobe: ModprobeSpec{
+							ModuleName: "mod-name",
+						},
```
ybettan, line 91: Is this relevant to the test? I think the code will fail before reaching that check anyway.

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+		Expect(e).To(HaveOccurred())
+		Expect(e.Error()).To(
+			ContainSubstring(
+				"missing spec.moduleLoader.container.kernelMappings[].containerImage",
+			),
+		)
+	})
+})
+
+var _ = Describe("validateModuleName", func() {
```
ybettan, line 110:
```suggestion
var _ = Describe("validateModprobe", func() {
```

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+	It("should fail when moduleName is missing", func() {
+		mod := &Module{}
+
+		e := mod.validateModuleName()
```
ybettan, line 114:
```suggestion
		e := mod.validateModprobe()
```

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+	})
+
+	It("should fail when rawArgs.load is empty", func() {
+		mod := &Module{
+			Spec: ModuleSpec{
+				ModuleLoader: ModuleLoaderSpec{
+					Container: ModuleLoaderContainerSpec{
+						Modprobe: ModprobeSpec{
+							RawArgs: &ModprobeArgs{
+								Load: nil,
```
ybettan, line 126:
```suggestion
								Load: []string{},
```

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+	It("should fail when rawArgs.unload is empty", func() {
+		mod := &Module{
+			Spec: ModuleSpec{
+				ModuleLoader: ModuleLoaderSpec{
+					Container: ModuleLoaderContainerSpec{
+						Modprobe: ModprobeSpec{
+							RawArgs: &ModprobeArgs{
+								Load:   []string{"arg"},
+								Unload: nil,
```
ybettan, line 147:
```suggestion
								Unload: []string{},
```

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+				},
+			},
+		}
+
+		e := mod.validateModuleName()
+		Expect(e).To(HaveOccurred())
+		Expect(e.Error()).To(ContainSubstring("missing spec.moduleLoader.container.modprobe.rawArgs.unload"))
+	})
+
+	It("should pass when rawArgs has load and unload values", func() {
```
ybettan, line 160: We also need a test to validate that it passes when `spec.moduleLoader.container.modporbe.name` is set (with no rawargs).

cmd/manager/main.go
```diff
@@ -178,6 +178,9 @@ func main() {
 		}
 	}
 
+	if err = (&v1beta12.Module{}).SetupWebhookWithManager(mgr); err != nil {
+		cmd.FatalError(setupLogger, err, "unable to create webhook", "webhook", "Module")
+	}
```
ybettan, line 183:
```suggestion
	}
	
```

ybettan:
```
resource mapping not found for name: "kmm-operator-serving-cert" namespace: "kmm-operator-system" from "ci/install-ci": no matches for kind "Certificate" in version "cert-manager.io/v1"
ensure CRDs are installed first
resource mapping not found for name: "kmm-operator-selfsigned-issuer" namespace: "kmm-operator-system" from "ci/install-ci": no matches for kind "Issuer" in version "cert-manager.io/v1"
ensure CRDs are installed first
make: *** [Makefile:184: deploy] Error 1
```

api/v1beta1/module_webhook.go
erusso7: I prefer sub-sequent `if` statements instead of one with multiple conditions... The validation won't pass if there is one missing so it's checking the same, isn't it?

api/v1beta1/module_webhook_test.go
```diff
@@ -0,0 +1,263 @@
...
+		Expect(e).ToNot(HaveOccurred())
+	})
+
+	It("should pass when regexp and containerImage are valid", func() {
+		mod1 := &Module{
+			Spec: ModuleSpec{
+				ModuleLoader: ModuleLoaderSpec{
+					Container: ModuleLoaderContainerSpec{
+						KernelMappings: []KernelMapping{
+							{Regexp: "", ContainerImage: "image-url"},
```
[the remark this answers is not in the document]
erusso7: Not sure will node will pick, but the regex is valid as the `Compile` method is not complaining

api/v1beta1/module_webhook.go
erusso7: Not sure. I understood that the `modprobe.ModuleName` has a preference and if it's set, we can ignore the `RawArgs`, but maybe I misunderstood the behavior. 

@qbarrand can you confirm this?

api/v1beta1/module_webhook.go
ybettan: Ack. If it was generated that way then feel free to process as you wish. Just make sure it is consistent along the whole file.

api/v1beta1/module_webhook.go
ybettan: You are right, the validation will fail anyway, the error message is a bit less explicit though. I think it makes sense to group them together as they are treated as a pair (both are set or both unset) but if you insist on keeping 2 separate `if`s I won't object.

api/v1beta1/module_webhook_test.go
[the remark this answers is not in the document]
ybettan: Can we put in a real regexp? for example `"*"`? just to make sure we are testing a real usecase?

api/v1beta1/module_webhook_test.go
ybettan: Yup. Sorry, this is hard to follow when non-interface helper-function are used :)
If it is cover in another test then no need to duplicate. Sorry for the noise.

api/v1beta1/module_webhook_test.go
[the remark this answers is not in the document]
erusso7: actually, I think that's invalid because there isn't any char to repeat with `*`. I used a similar example in the invalid case, but anyway, I added a valid case as well :+1:

PROJECT
```diff
@@ -16,9 +16,11 @@ resources:
   kind: Module
   path: github.com/kubernetes-sigs/kernel-module-management/api/v1beta1
   version: v1beta1
+  webhooks:
+    validation: true
+    webhookVersion: v1
 - api:
     crdVersion: v1
-    namespaced: false
```
qbarrand, line 21: Please add this field back.

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,101 @@
...
+
+	return nil
+}
+
+func (m *Module) validateModprobe() error {
+	modprobe := m.Spec.ModuleLoader.Container.Modprobe
+
+	if modprobe.ModuleName != "" {
+		return nil
+	}
```
ybettan, line 90:
```suggestion
        if modprobe.ModuleName != "" && modprobe.RawArgs != nil {
            return errors.New("moduleName and rawArgs are mutually exclusive in spec.moduleLoader.container.modprobe")
        }

	if modprobe.ModuleName != "" {
		return nil
	}
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if modprobe.RawArgs == nil {
+		return errors.New("missing spec.moduleLoader.container.modprobe.moduleName")
+	}
+
+	if len(modprobe.RawArgs.Load) == 0 || len(modprobe.RawArgs.Unload) == 0 {
+		return errors.New("when using spec.moduleLoader.container.modprobe.rawArgs, both load and unload fields should be specified")
+	}
```
ybettan, line 98: I think that `moduleName` and `rawArgs` are mutually exclusive so I would go with:
* If both are not set - return an error
* If both are set - return an error

At this point we can assume that exactly one of them is set, then:
* if moduleName is set return nil

At this point we can assume that `rawArgs` is set
* make sure that both `load` and `unload` are set.

ybettan, line 98: I think there is an issue with this validation. According to the [code](https://github.com/kubernetes-sigs/kernel-module-management/blob/50a9d36c7ec21cad4d201af40d6d0a96e0db8386/internal/daemonset/daemonset.go#L452) `moduleName` must be set so I would start from there and then assume it is set.

qbarrand: I think it's fine to mandate that if raw args are used to load a module, they also need to be used to unload it.
I think the logic is:
```go
moduleNameDefined := modprobe.ModuleName
rawLoadArgsDefined := modprobe.RawArgs != nil && len(modprobe.RawArgs.Load) > 0
rawUnloadArgsDefined := modprobe.RawArgs != nil && len(modprobe.RawArgs.Unload) > 0

if moduleNameDefined && (rawLoadArgsDefined || rawUnloadArgsDefined) {
    return errors.New("rawArgs cannot be set when moduleName is set")
}

if !moduleNameDefined && (!rawLoadArgsDefined || !rawUnloadArgsDefined) {
    return errors.New("load and unload rawArgs must be set when moduleName is unset")
}
```

ci/prow/e2e-incluster-build
```diff
@@ -30,6 +33,14 @@ kubectl apply -f ci/kmm-kmod-dockerfile.yaml
 echo "Add Secrets containing the signing key and certificate..."
 kubectl apply -f ci/secret_kmm-kmod-signing.yaml
 
+# Wait controller-manager and cert-manager are configured before we apply the module because the validation webhook will be called.
+CM_NS=cert-manager
+POD_NAME=$(kubectl -n ${CM_NS} get pods -o json | jq -r '.items[].metadata.name | select(.? | match("cert-manager-webhook"))')
+kubectl wait -n ${CM_NS} "pod/${POD_NAME}" --for jsonpath='{.status.phase}'=Running --timeout=60s
+KMM_NS=kmm-operator-system
+POD_NAME=$(kubectl -n ${KMM_NS} get pods -o json | jq -r '.items[].metadata.name | select(.? | match("controller"))')
+kubectl wait -n ${KMM_NS} "pod/${POD_NAME}" --for jsonpath='{.status.phase}'=Running --timeout=60s
```
qbarrand, line 42: I think we could just wait for the KMM deployment here.
```suggestion
kubectl wait --for=condition=Available deployment/kmm-operator-controller-manager -n kmm-operator-system
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+			return fmt.Errorf("regexp and literal are mutually exclusive properties at kernelMappings[%d]", idx)
+		}
+
+		if km.Regexp == "" && km.Literal == "" {
+			return fmt.Errorf("regexp or literal must be set at kernelMappings[%d]", idx)
+		}
+
+		if _, err := regexp.Compile(km.Regexp); err != nil {
+			return fmt.Errorf("invalid regexp at index %d: %v", idx, err)
+		}
```
ybettan, line 83:
```suggestion
                if km.Literal != "" {
                    return nil
                }
                
		if _, err := regexp.Compile(km.Regexp); err != nil {
			return fmt.Errorf("invalid regexp at index %d: %v", idx, err)
		}
```

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+
+	return m.validateModprobe()
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (m *Module) ValidateUpdate(_ runtime.Object) error {
+	modulelog.Info("Validating Module update", "name", m.Name, "namespace", m.Namespace)
+
+	if err := m.validateKernelMapping(); err != nil {
+		return fmt.Errorf("failed to validate the kernel mapping regexps: %v", err)
```
ybettan, line 58:
```suggestion
		return fmt.Errorf("failed to validate kernel mappings: %v", err)
```
We are not only validating the regexps but also making sure the container image is set.

api/v1beta1/module_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+
+var _ webhook.Validator = &Module{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (m *Module) ValidateCreate() error {
+	modulelog.Info("Validating Module creation", "name", m.Name, "namespace", m.Namespace)
+
+	if err := m.validateKernelMapping(); err != nil {
+		return fmt.Errorf("failed to validate the kernel mapping regexps: %v", err)
+	}
```
ybettan, line 48:
```suggestion
		return fmt.Errorf("failed to validate kernel mappings: %v", err)
```
We are not only validating the regexps but also making sure the container image is set.

.github/workflows/e2e.yaml
```diff
@@ -74,6 +74,9 @@ jobs:
         with:
           start-args: --insecure-registry=host.minikube.internal:5000
 
+      - name: Install cert-manager
+        run: kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
+
```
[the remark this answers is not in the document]
ybettan: I have created 2 PRs to make sure we are not missing such changes in the future:
* https://github.com/kubernetes-sigs/kernel-module-management/pull/353
* https://github.com/kubernetes/test-infra/pull/29149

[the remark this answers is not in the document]
ybettan: Also, don't we need to install `cert-manager` and the validation webhook in `manager-hub` as well?
We can open an issue about it and do it in a later PR though.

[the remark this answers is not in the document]
ybettan: We may need to use https://github.com/kubernetes-sigs/kustomize/issues/880#issuecomment-967153917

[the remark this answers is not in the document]
ybettan: I tried to extract the installation of `certmanager`  (not the `Certificate` nor the `Issuer` but `certmanager itself) to another PR that should be used as a base for this PR IMO.
https://github.com/kubernetes-sigs/kernel-module-management/pull/355

api/v1beta1/module_webhook.go
erusso7: What I have talked about with @qbarrand is the `Literal` and `Regexp` are mutually exclusive. So, I the user is using `Literal`, then the `Regexp` must be empty, and checking that regexp will be very simple and fast, moreover if we `return nil` as in this suggestion, then we will miss the validation of the `containerImage` and that's a requirement. So, I would suggest keeping it this way instead of over-complicating the code with more `if` statements that seem redundant from my POV.
WDYT?

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
ybettan: Updated PR for installing `cert-manager`
* https://github.com/kubernetes-sigs/kernel-module-management/pull/359.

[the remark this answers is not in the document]
ybettan: @erusso7 Can you please rebase the PR on the new `main` branch?

api/v1beta1/module_webhook.go
ybettan: I agree. Your code is better than my suggestion. Let's close this thread and keep the code as it is.

config/cert-manager/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+bases:
+- https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
```
ybettan, line 2: We don't need this one anymore. It is now being installed as a pre-requisite by `make deploy`

Issue #104, opened 2022-10-07
remove in tree module as a pre-requisite.


mregmi:
Hello, Is there a way to tell the KMMO to remove a certain kernel module before we load ours? In a OOT driver we are currently working on, we need to remove the in tree driver before we load ours. Thanks
To elaborate, in order to insert our OOT graphics driver, we first need to remove in tree drm driver or else the insert will fail. if there no current functionality to do this before insert, a way to run a pre-requisite commands/scripts would be a good idea.

qbarrand:
Would an [InitContainer](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) that runs `modprobe -r <list of modules to unload>` be a good solution to this?

mregmi:
InitContainer for kmmo? we can give it a try. Thanks.

qbarrand:
I mean than KMM would generate a module loader DaemonSet that contains such an init container. That means that we would not be able to load the in-tree module again after we unload the OOT one, but at least that would fit your use case.

uMartinXu:
@qbarrand I can't understand the benefits to do it in init-container.  It is for security concern. And you want restrict user to do any operation they want in KMMO like SRO? If so, user also can do what ever they want to do in init-container also can bring the risk.

Before we talking about the security, we need to start from the user/role of the KMMO.  Please correct me. @qbarrand 
1. Cluster-administrators the user who make use of KMMO Module CR to manage drivers in the cluster. As the Cluster-administrators, they should be trusted in some degree. But we can't fully trust him, because they also can make some mistake by accident to impair the cluster. So we need to design KMMO to reduce the chance for the administrators to make this kind of mistake. Actually SeLinux and Linux Smack technology can be used to achieve it.
2. Driver developers/distributors: who develop, package and distribute the driver code and driver container images. Normally they are not in charge of manage the Moule CR in a specific cluster. Their code and packaged binary should be fully trust. Since the driver will be insmod into kernel, and in theory the code can do everything in the kernel. And they have the full knowledge how to build the driver, how to inmod and configure the driver, and also know how sanity check the driver.

So if we agree on these two roles, then, I think it should be Driver developers to provide the process which can be some program or script (static linked binary is preferred). And the program can be packaged into the driver container image for KMMO to insmod and configure the driver. Since this program is from driver developers/distributors, and can be fully trusted. And should not has the security vulnerability.

uMartinXu:
@qbarrand any comments? :)

[...]

PR #354, opened 2023-03-26
In-tree kernel module replacement

yevgeny-shnaidman:
This PR implements in-tree kernel module replacement per user request:
1) Add a boolean flag to KernelMapping and to Container definition
   In case KernelMapping's field is defined, it takes precedence over
   Container field
2) In case the boolean flag is defined, the PostStart command will now
   first try to remove in-tree kernel module, and only after that will try
   to insert the kernel module stored in the container image
3) unit-test


ybettan:
Last time we talked about it I remember we said we need to check if it is possible to load the in-tree driver back once the module is removed. Do we have more knowledge about it?

yevgeny-shnaidman:
> Last time we talked about it I remember we said we need to check if it is possible to load the in-tree driver back once the module is removed. Do we have more knowledge about it?

This is a very problematic thing to do: we don't know exactly where the in-tree kernel module is stored ( it can be under /lib/modules, but it can also be under a different path). We don't know if there is a modporbe.conf file  for the in-tree kernel module, and it may conflict with the modprobe.conf stored in the containerImage. All in all, it seems like a lot of work, a lot of new definitions, for very-very small benifit, which we aren't even sure if the customer wants

ybettan:
Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/104.

ybettan:
What happens if the in-tree module we would like to remove have a different name than the one we are inserting?

yevgeny-shnaidman:
> What happens if the in-tree module we would like to remove have a different name than the one we are inserting?

We don't support that option. The main goal of this issue is to allow user to replace the version of an existing driver. The in-tree is usually 2-3 years behind the current driver, so customers just want to replace to a newer version, which means the same driver. We may get back to this option once there is a request for that, but we have not seen such a request yet

ybettan:
SGTM.
I am OK to merge if @qbarrand is.

qbarrand:
What is the downside of making that property a string instead? We would address the use case of different names for free.

yevgeny-shnaidman:
> What is the downside of making that property a string instead? We would address the use case of different names for free.

I don't think we want to support the unloading of different names. 2 reason:
1) we have never seen a user-case for that
2) this option will allow one module to unload the kernel modules of another module (either intentionally or by mistake). It will be a hell to debug. 
I don't think we should go that way, unless we see a number of use-cases for that.

PR #355, opened 2023-03-27
Adding `cert-manager` to config/default.

ybettan:
It is required for adding validation webhooks for example.

This commit also contains side-effects required for this change:

1. `cert-manager` is installing some resources to the `cert-manager`
   namespace and some to the `kube-system` namespace, therefore we
   cannot use a single namespace override for all resources, so the
   global definition of `namespace:` in the `kustomization.yaml` was
   removed.

2. Added the `NamespaceTransformer` as a replacement for the global
   `namespace:` since it allows using the `unsetOnly` field which will
    now use the new "global namespace" only for resources that didn't
    explicitly specified their namespace.

3. Removed the `system` namespaces for all resources that expect to get
   the default namespace.

4. Added the `kind: Kustomization` type to all resources, it allows for
   k8s to run some type validation. I guess the reason it was not
   present until now is because it didn't exist in previous versions of
   `kustomize` and `operator-sdk` is using an old version.


ybettan:
FYI @erusso7

ybettan:
@yevgeny-shnaidman I have updated the commit message and the PR description. PTAL.

PR #360, opened 2023-03-28
Making sure that `cert-manager` is ready before deploying KMM.

ybettan:
KMM will soon start applying a `Certificate` and an `Issuer` which requires that `cert-manager` installation is done.

Also added some makefile targets to remove duplications and updated the installation instructions.


Makefile
```diff
@@ -175,31 +175,46 @@ install: manifests ## Install CRDs into the K8s cluster specified in ~/.kube/con
...
 	kubectl delete -k $(KUSTOMIZE_CONFIG_CRD) --ignore-not-found=$(ignore-not-found)
 
-KUSTOMIZE_CONFIG_DEFAULT ?= config/default
-KUSTOMIZE_CONFIG_HUB_DEFAULT ?= config/default-hub
+KUSTOMIZE_CONFIG_CERT_MANAGER ?= https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
+
+.PHONY: deploy-cert-manager
+deploy-cert-manager:
+	kubectl apply -f $(KUSTOMIZE_CONFIG_CERT_MANAGER)
+	kubectl -n cert-manager wait --for=condition=Available deployment \
```
qbarrand, line 183: We don't need to wait for the cert-manager operator to be ready - we just need to wait for the CRDs to have the `Established` condition.

README.md
```diff
@@ -7,8 +7,19 @@ The Kernel Module Management Operator manages out of tree kernel modules in Kube
...
+kubectl -n cert-manager wait --for=condition=Available deployment \
+	cert-manager \
+	cert-manager-cainjector \
+	cert-manager-webhook
+```
+
+Install KMM.
 ```shell
-kubectl apply -k https://github.com/kubernetes-sigs/kernel-module-management/config/default
+kubectl apply -k https://github.com/kubernetes-sigs/kernel-module-management/config/kmm
```
qbarrand, line 22: Should we revert the PR that creates this overlay? Since we cannot reliably install both cert-manager and KMM in one kustomize command, it's not really useful anymore.

Makefile
ybettan: I think there is also an http call done and if the pod/service isn't ready we get a timeout. I will need to validate it though.

qbarrand: `cert-manager` will only generate the `Secret` KMM needs once it's ready. Thus it may take some time for the KMM webhook to come online; in the meantime all `Module` creations and updates will fail. I think that is fine; if a user wants to wait for KMM (and thus `cert-manager`) to be ready, they can wait for the KMM `Deployment` to be `Available`.

ybettan: So the first error we may hit is the CRD not installed:
```
resource mapping not found for name: "kmm-operator-serving-cert" namespace: "kmm-operator-system" from "config/default": no matches for kind "Certificate" in version "cert-manager.io/v1"
ensure CRDs are installed first
resource mapping not found for name: "kmm-operator-selfsigned-issuer" namespace: "kmm-operator-system" from "config/default": no matches for kind "Issuer" in version "cert-manager.io/v1"
ensure CRDs are installed first
make: *** [Makefile:185: deploy] Error 1
```

And then you can get another (later) error:
```
Error from server (InternalError): error when creating "config/default": Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": dial tcp 10.96.229.242:443: connect: connection refused
Error from server (InternalError): error when creating "config/default": Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": dial tcp 10.96.229.242:443: connect: connection refused
make: *** [Makefile:185: deploy] Error 1
```

As I understand it is is the Certificate webhook validation that isn't ready yet.
So I don't see another option rather than waiting for the webhook pod to be ready (which is the slowest of all tree so there is no harm in waiting to all of the cert-manager deployments).

qbarrand: I did not see those in my deployment. Totally makes sense to wait for the deployment as well indeed.

ybettan:
/unhold
This PR was also tested together with
* https://github.com/kubernetes-sigs/kernel-module-management/pull/348
at
* https://github.com/kubernetes-sigs/kernel-module-management/pull/363

docs/mkdocs/documentation/install.md
```diff
@@ -2,8 +2,18 @@
...
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
+kubectl -n cert-manager wait --for=condition=Available deployment \
+	cert-manager \
+	cert-manager-cainjector \
+	cert-manager-webhook
+```
+
+Install KMM.
+```shell
+kubectl apply -k https://github.com/kubernetes-sigs/kernel-module-management/config/kmm
```
qbarrand, line 16: Why don't we revert to `config/default`?

docs/mkdocs/documentation/install.md
```diff
@@ -2,8 +2,18 @@
...
 ## Using `kubectl`
 
+First install [cert-manager](https://github.com/cert-manager/cert-manager) which is a dependency.
 ```shell
-kubectl apply -k https://github.com/kubernetes-sigs/kernel-module-management/config/default
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
+kubectl -n cert-manager wait --for=condition=Available deployment \
+	cert-manager \
+	cert-manager-cainjector \
+	cert-manager-webhook
```
qbarrand, line 11: Why do we need to wait for all 3 deployments?
```suggestion
kubectl -n cert-manager wait --for=condition=Available deployment cert-manager-webhook
```

qbarrand: Also, we should wait for the `Issuer` and `Certificate` CRDs to be `Established`.

docs/mkdocs/documentation/uninstall.md
```diff
@@ -3,7 +3,8 @@
 ## Using `kubectl`
 
 ```shell
-kubectl delete -k https://github.com/kubernetes-sigs/kernel-module-management/config/default
+kubectl delete -k https://github.com/kubernetes-sigs/kernel-module-management/config/kmm
+kubectl delete -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
```
qbarrand, line 7: Can we do both in one command?
```suggestion
kubectl delete -k https://github.com/kubernetes-sigs/kernel-module-management/config/kmm -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
```

bundle.Dockerfile
qbarrand, line 1: Let's not commit bundle files if we don't have to.

docs/mkdocs/documentation/install.md
ybettan: Because now we have 2 operators installing, `cert-manager` and KMM so I though it make more sense to call it `kmm`.

docs/mkdocs/documentation/install.md
ybettan: > Why do we need to wait for all 3 deployments?

We need to wait for the webhook cause it is validating the `Certificate` and the `Issuer` objects. I added the other one just in case cuase this is what the [docs](https://cert-manager.io/docs/installation/verify/) says. I prefer to be on the safe side rather than having another unknown issue in the near future. (The slowest is the webhook anyway).

bundle.Dockerfile
ybettan: Removed.

docs/mkdocs/documentation/install.md
ybettan: > Also, we should wait for the Issuer and Certificate CRDs to be Established.

Is it possible for an operator to be ready if it's CRDs aren't present yet? I see nothing about it in the docs.

qbarrand: > Is it possible for an operator to be ready if it's CRDs aren't present yet?

Probably not since it watches them.

ybettan:
/unhold
Both PRs works together

PR #361, opened 2023-03-28
Refactor build / sign managers

qbarrand:
Create a new `internal/imgbuild` package that brings a new `Manager` type covering the needs of both features, as well as a `JobMaker` interface.
Implement that interface in `internal/{build,sign}` packages.
Consolidate `ModuleLoaderData` generation code into the `internal/api` package.
Move `utils` functions into their own package and remove duplicate code.


qbarrand:
/hold
Some unit tests are probably missing.

yevgeny-shnaidman:
IMHO, this PR can be split into 3: imgbuild changes, ModuleLoaderData changes, utils changes.

internal/imgbuild/jobmanager.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+type JobType string
+
+const (
+	JobTypeBuild JobType = "build"
+	JobTypeSign  JobType = "sign"
+)
+
+type BaseJobManager struct {
```
yevgeny-shnaidman, line 24: why do we need this struct public? It is used only in this package, no?

internal/imgbuild/jobmanager.go
```diff
@@ -0,0 +1,165 @@
...
+				return nil, fmt.Errorf("failed to delete build job %s: %v", job.Name, err)
+			}
+			deleteNames = append(deleteNames, job.Name)
+		}
+	}
+	return deleteNames, nil
+}
+
+func (bjm *BaseJobManager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushImage bool, owner metav1.Object) (Status, error) {
+	logger := log.FromContext(ctx).WithValues("type", bjm.jobType)
```
yevgeny-shnaidman, line 52: lets add at least Name from mld to the WithValues, so that we can find in the log which module are we talking about here

qbarrand:
Extracted `internal/{api,module}` changes into https://github.com/kubernetes-sigs/kernel-module-management/pull/364.

PR #368, opened 2023-04-04
Adding a validation webhook to the `ManagedClusterModule` CRD.

ybettan:
Following the webhook added for the `Module` CRD we now also need a validation webhook to validate the `ManagedClusterModule` which is used in the hub-spoke topology.


ybettan:
@erusso7 Your review will be appreciated as well.

cmd/manager-hub/main.go
```diff
@@ -139,6 +139,10 @@ func main() {
 		cmd.FatalError(ctrlLogger, err, "unable to create controller")
 	}
 
+	if err = (&v1beta1.ManagedClusterModule{}).SetupWebhookWithManager(mgr); err != nil {
```
yevgeny-shnaidman, line 142: WDYT about adding a comment regarding the validation path explanation ( what we found out today during debug) to some place in a code

ybettan: Added the comment inside `SetupWebhookWithManager` for both `Module` and `ManagedClusterModule` webhook files.

api-hub/v1beta1/managedclustermodule_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
```
mresvanis, line 27: I think this comment is not necessary, but if we would like to keep it let's start with `// modulelog is for ...` . WDYT?

api-hub/v1beta1/managedclustermodule_webhook.go
```diff
@@ -0,0 +1,69 @@
...
+
+	module := &kmmv1beta1.Module{
+		Spec: mcm.Spec.ModuleSpec,
+	}
+	return module.ValidateUpdate(obj)
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (mcm *ManagedClusterModule) ValidateDelete() error {
+
```
mresvanis, line 64: Do you we maybe need a log info here as well? Like what we have [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/368/files#diff-fe0b4a96e1db25686e8c4c6b7b46b7c12afb5bbf530ea4f7a63d50aaad4db06eR44) and [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/368/files#diff-fe0b4a96e1db25686e8c4c6b7b46b7c12afb5bbf530ea4f7a63d50aaad4db06eR54) ?

api-hub/v1beta1/managedclustermodule_webhook.go
ybettan: I didn't like it either but kept is because it was copy-pasted from a file that was generated using the `operator-sdk`. Will remove the comment.

api-hub/v1beta1/managedclustermodule_webhook.go
ybettan: The reason we don't have a log is because the validation is "empty" we are not checking anything, which means it is basically like there was no webhook at all when removing. (we are not having a log on deleting `Module`s either in the `Module`'s webhook.

Do you still think we should have a log here?

api-hub/v1beta1/managedclustermodule_webhook.go
ybettan: Fixed.

api-hub/v1beta1/managedclustermodule_webhook.go
ybettan: Added the log anyway cause it is calling an internal function and it is not clear that it is not doing anything.

mresvanis (approved):

ybettan:
/hold
Making sure @yevgeny-shnaidman approves the changes as well

ybettan:
/unhold
@yevgeny-shnaidman has approved offline.

Issue #126, opened 2022-10-26
Create a base runtime image for Module Loader


qbarrand:
When creating the Module Loader DaemonSet, KMM currently sets `sleep infinity` as the container's `command`.
`sleep` runs as the main process in the container (PID 1) and is therefore [unkillable](https://raby.sh/sigterm-and-pid-1-why-does-a-container-linger-after-receiving-a-sigterm.html). That makes the container hang when terminated, because running as PID 1 means that the `TERM` signal that Kubernetes sends is ignored unless the binary specifically implements a `TERM` handler (which `sleep` doesn't).
The following solutions would mitigate that problem:
1. set the Module Loader pod's [`shareProcessNamespace`](https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/) property to `true`, but that implies that other containers in the pod may be able to interact with the Module Loader (athough there is a single container in that pod now);
2. introduce a requirement that [`pause`](https://github.com/kubernetes/kubernetes/blob/master/build/pause/linux/pause.c) be present in the Module Loader image - it explicitly handles the `TERM` signal which is the desired behaviour.

To facilitate option 2, we could release a slim image that includes both `modprobe` and `pause` - the two components that KMM requires in the image.

mresvanis:
We should probably add `find` and `xargs` (`findutils` [package](https://www.gnu.org/software/findutils/)) to the requirements, with regards to the firmware [cleanup](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/internal/daemonset/daemonset.go#L422). 

```go
fwUnloadCommand = fmt.Sprintf(" && cd %s && find |sort -r |xargs -I{} rm -d %s/$(basename %s)/{}", fw, nodeVarLibFirmwarePath, fw)
```

qbarrand:
We will not run ModuleLoader images in v2 but rather mount them and load kmods ourselves. @mresvanis @yevgeny-shnaidman OK to close this?

yevgeny-shnaidman:
Maybe we should implement option 1 for V1?  We already have another bug that is probably dependent on this one

PR #379, opened 2023-04-18
Fixing termination timeout for module loader pods

yevgeny-shnaidman:
Setting the SharedProcessNamespace to true, which will cause the PID 1 process of the pod container to be pause. Puase process is trapping SIGTERM, which will allow kubelet to destroy the container process immediately, without 30 seconds timeout

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/go-openapi/swag"
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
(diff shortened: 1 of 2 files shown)


ybettan:
Isn't `SIGINT` stands for "interrupt from keyboard"?
What are we trying to solve with this PR?

yevgeny-shnaidman:
> Isn't `SIGINT` stands for "interrupt from keyboard"? What are we trying to solve with this PR?

Sorry, SIGTERM. What we are trying to solve is issue that it takes usually ~30 seconds and more for the pods of the ModuleLoader to be deleted when we delete Module

yevgeny-shnaidman:
Fixes [https://github.com/kubernetes-sigs/kernel-module-management/issues/126]

PR #380, opened 2023-04-24
Adding api for ordered multiple kernel modules

yevgeny-shnaidman:
Adding list entry under Modprobe spec that defines
the sequence of loading multiple independent kernel modules
that are stored in the Module image. The entry is a list with
the modules that depends on all other modules at the head of the list.
Example: if moduleA depends on first loading moduleB, and moduleB depends on first loading moduleC
then the list will look:
ModulesLoadingOrder:
    - moduleA
    - moduleB
    - moduleC
moduleC shoudl be loaded first, then module B, and then module C.
moduleA should be set in the ModuleName entry of the ModprobeSpec


api/v1beta1/module_types.go
```diff
@@ -163,6 +163,14 @@ type ModprobeSpec struct {
...
 	// +optional
 	FirmwarePath string `json:"firmwarePath,omitempty"`
+
+	// ModulesLoadingOrder defines the dependency between modules loading, in case
+	// it was not created by depmod (independent modules). Format is:
+	// moduleA => moduleB => moduleC
+	// meaning: loading moduleA will require loading moduleB, and loading moduleB will require loadig moduleC.
+	// In order to load all 3 modules, moduleA shoud be defined in the ModuleName parameter of this struct
+	// +optional
+	ModulesLoadingOrder string `json:"modulesLoadingOrder,omitempty"`
```
qbarrand, line 173: Could this be a list instead? I do not see the point of this special syntax.

yevgeny-shnaidman: I think that special syntax make its more clear regarding the ordering.

yevgeny-shnaidman: fixed

config/crd-hub/bases/hub.kmm.sigs.x-k8s.io_managedclustermodules.yaml
```diff
@@ -2362,6 +2362,18 @@ spec:
                                 description: ModuleName is the name of the Module
                                   to be loaded.
                                 type: string
+                              modulesLoadingOrder:
+                                description: 'ModulesLoadingOrder defines the dependency
```
qbarrand, line 2366: Please update the CRD description to match the comment.

yevgeny-shnaidman: fixed

PR #381, opened 2023-04-24
Loading multiple independent kernel modules in one KMM Module

yevgeny-shnaidman:
This PR allows defining the load order of multiple independent kernel modules in one KMM Module. The order is defined as a list in the ModulesLoadingOrder field of the ModprobeSpec. The first module will be loaded last, the last one in the list will be loaded first. It is implement by using DownwardAPI: module loader pod template contains is added an annotation that contains the /etc/modprobe.d/softdep format, and then this annotation is added as a file to module loader's pod filesystem using Downward API

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -57,19 +57,12 @@ func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, mod *kmmv1beta
 	deleted := make([]string, 0)
 
 	for _, ds := range existingDS {
-		if isOlderVersionUnusedDaemonset(&ds, mod.Spec.ModuleLoader.Container.Version) {
+		if isOlderVersionUnusedDaemonset(&ds, mod.Spec.ModuleLoader.Container.Version) ||
+			isModuleLoaderDaemonsetWithInvalidKernel(&ds, dc.kernelLabel, validKernels) {
+			deleted = append(deleted, ds.Name)
 			if err := dc.client.Delete(ctx, &ds); err != nil {
 				return nil, fmt.Errorf("could not delete DaemonSet %s: %v", ds.Name, err)
 			}
-			deleted = append(deleted, ds.Name)
-			continue
-		}
-		if isModuleLoaderDaemonsetWithInvalidKernel(&ds, dc.kernelLabel, validKernels) {
-			if err := dc.client.Delete(ctx, &ds); err != nil {
-				return nil, fmt.Errorf("could not delete DaemonSet %s: %v", ds.Name, err)
-			}
-
-			deleted = append(deleted, ds.Name)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/module_reconciler.go
```diff
@@ -367,12 +381,27 @@ func (mrh *moduleReconcilerHelper) handleDriverContainer(ctx context.Context,
...
 
 	if err == nil {
-		logger.Info("Reconciled Driver Container", "name", ds.Name, "result", opRes)
+		logger.Info("Reconciled Module Loader Container", "name", ds.Name, "result", opRes)
 	}
 
 	return err
 }
 
+func (mrh *moduleReconcilerHelper) handleModuleOrderConfigMap(ctx context.Context, mld *api.ModuleLoaderData) error {
```
qbarrand, line 390: Do we need a ConfigMap? The config could be a Pod annotation that we could mount using the [downward API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).

yevgeny-shnaidman: That is not an optimal solution, since you cannot map each label/annotation individually. So you will get a file with label=<value>. I think it is easier to do it with ConfigMap, and will be easier to debug. The only change is that we create one configmap per Module.

yevgeny-shnaidman: Fixed

internal/daemonset/daemonset.go
```diff
@@ -472,3 +493,20 @@ func MakeUnloadCommand(spec kmmv1beta1.ModprobeSpec, modName string) []string {
...
+		}
+		line := prepareSoftDepLine(modulesNames[i], modulesNames[i+1])
+		softDepData = softDepData + line
+	}
+	return softDepData
+}
+
+func prepareSoftDepLine(dependendModuleName, moduleName string) string {
+	return fmt.Sprintf("softdep %s pre: %s\n", dependendModuleName, moduleName)
+}
```
qbarrand, line 512: Could we use a Go template here instead?

internal/daemonset/daemonset.go
```diff
@@ -57,19 +57,12 @@ func (dc *daemonSetGenerator) GarbageCollect(ctx context.Context, mod *kmmv1beta
 	deleted := make([]string, 0)
 
 	for _, ds := range existingDS {
-		if isOlderVersionUnusedDaemonset(&ds, mod.Spec.ModuleLoader.Container.Version) {
-			if err := dc.client.Delete(ctx, &ds); err != nil {
-				return nil, fmt.Errorf("could not delete DaemonSet %s: %v", ds.Name, err)
-			}
+		if isOlderVersionUnusedDaemonset(&ds, mod.Spec.ModuleLoader.Container.Version) ||
+			isModuleLoaderDaemonsetWithInvalidKernel(&ds, dc.kernelLabel, validKernels) {
```
qbarrand, line 61: This looks unrelated to the rest of the PR.

yevgeny-shnaidman: This is just a better refactor of the GarbageCollection

yevgeny-shnaidman: added it here, in order not to open a different PR just for that

internal/daemonset/daemonset.go
yevgeny-shnaidman: We can, but i am not sure if that will male the code more readable , or reduce the number of code line. We will need to use range inside template definition, and we will still need to use the loop over the modulesName to prepare the input data for templates

PR #383, opened 2023-04-27
Prefix the webhook secret name with kmm-operator-

### LINKED ISSUE #382: kustomize: prefix the webhook secret name
The Secret name generated by cert-manager should have the `kmm-operator-` prefix like all other resources created by kustomize. This should be achievable by using a namePrefix directive in the kustomize config.
Certificate object: https://github.com/kubernetes-sigs/kernel-module-management/blob/main/config/certificate/certificate.yaml#L39

qbarrand:
Fixes #382

/cc @ybettan

### DIFF
--- a/config/certificate/kustomizeconfig.yaml
+++ b/config/certificate/kustomizeconfig.yaml
@@ -14,3 +14,9 @@ varReference:
 - kind: Certificate
   group: cert-manager.io
   path: spec/dnsNames
+
+namePrefix:
+- kind: Certificate
+  group: cert-manager.io
+  name: serving-cert
+  path: spec/secretName
--- a/config/default/kustomizeconfig.yaml
+++ b/config/default/kustomizeconfig.yaml
@@ -1,0 +1,4 @@
+namePrefix:
+- kind: Deployment
+  name: controller-manager
+  path: spec/template/spec/volumes/secret/secretName
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -57,3 +57,6 @@ vars:
       kind: Service
       version: v1
       name: webhook-service
+
+configurations:
+  - kustomizeconfig.yaml


config/certificate/kustomizeconfig.yaml
```diff
@@ -14,3 +14,9 @@ varReference:
 - kind: Certificate
   group: cert-manager.io
   path: spec/dnsNames
+
+namePrefix:
+- kind: Certificate
+  group: cert-manager.io
+  name: serving-cert
+  path: spec/secretName
```
ybettan, line 22: This one is already prefixed by kustomize because it is a applied by kustomize:
```
ybettan kernel-module-management (e2e-hub) $ oc get certificate --all-namespaces
NAMESPACE             NAME                            READY   SECRET                    AGE
kmm-operator-system   kmm-operator-hub-serving-cert   True    webhook-hub-server-cert   2d
kmm-operator-system   kmm-operator-serving-cert       True    webhook-server-cert       2d
```

config/default/kustomizeconfig.yaml
```diff
@@ -0,0 +1,4 @@
+namePrefix:
+- kind: Deployment
+  name: controller-manager
+  path: spec/template/spec/volumes/secret/secretName
```
ybettan, line 4: What is the difference between this `namePrefix` and the one in `config/certificate/kustomization.yaml`?

qbarrand: The directive in `kustomization.yaml` specifies the value of the prefix. This specifies which fields need to be specified.

config/certificate/kustomizeconfig.yaml
qbarrand: The PR prefixes the Secret name generated by cert-manager as configured in the `Certificate` resource.

PR #385, opened 2023-05-01
Adding an e2e workflow for the hub-spoke topology.

ybettan:
Adding an e2e job for the hub-spoke topology.

This workflow is using a single cluster which is both the hub and also
the spoke and test the main flow of applying a `ManagedClusterModule` to
the hub and validating that all the artifacts are created on the spoke
cluster.


.github/workflows/e2e.yaml
```diff
@@ -108,3 +131,65 @@ jobs:
...
+
+    services:
+      registry:
+        image: registry:2
+        ports: ['5000:5000/tcp']
+
+    steps:
+
+      - name: Install curl on the machine
+        run: sudo apt update -y && sudo apt upgrade -y && sudo apt install -y curl
```
qbarrand, line 149: Why is this needed?

ybettan: We need it in order to install the `clusteradm` command in the e2e script:
```
# Install the `clusteradm` command
curl -L https://raw.githubusercontent.com/open-cluster-management-io/clusteradm/main/install.sh | bash
```

Makefile
```diff
@@ -181,7 +181,7 @@ KUSTOMIZE_CONFIG_CERT_MANAGER ?= https://github.com/cert-manager/cert-manager/re
 .PHONY: deploy-cert-manager
 deploy-cert-manager:
 	kubectl apply -f $(KUSTOMIZE_CONFIG_CERT_MANAGER)
-	kubectl -n cert-manager wait --for=condition=Available deployment \
+	kubectl -n cert-manager wait --timeout=2m --for=condition=Available deployment \
```
yevgeny-shnaidman, line 184: why do we need 2m timeouts?

ybettan: Because sometime 1m isn't enough and cert-manager isn't ready.

ybettan:
In case you want to see the job demo: https://asciinema.org/a/583299

PR #388, opened 2023-05-02
Add webhook checks for the softdep feature

qbarrand:
Reject the Module if:
- modulesLoadingOrder has a length < 2
- modulesLoadingOrder is defined but not moduleName
- the first element of modulesLoadingOrder is not moduleName
- moduleName is present in modulesLoadingOrder but not as its first element

/cc @yevgeny-shnaidman


api/v1beta1/module_webhook.go
```diff
@@ -95,19 +95,37 @@ func (m *Module) validateKernelMapping() error {
...
 	}
 
-	if !rawLoadArgsDefined || !rawUnloadArgsDefined {
-		return errors.New("load and unload rawArgs must be set when moduleName is unset")
+	if modprobe.ModulesLoadingOrder != nil {
+		if len(modprobe.ModulesLoadingOrder) < 2 {
+			return errors.New("if a loading order is defined, at least two values must be defined")
+		}
+
+		if !moduleNameDefined {
```
yevgeny-shnaidman, line 116: is there any sense to have a Module without moduleName defined? I think that verification can be moved to line 99

api/v1beta1/module_webhook.go
```diff
@@ -95,19 +95,37 @@ func (m *Module) validateKernelMapping() error {
...
+			return errors.New("if a loading order is defined, moduleName must be set")
+		}
+
+		if modprobe.ModulesLoadingOrder[0] != moduleName {
+			return errors.New("if a loading order is defined, the first element must be moduleName")
+		}
+
+		for idx, modName := range modprobe.ModulesLoadingOrder[1:] {
```
yevgeny-shnaidman, line 124: I think that we need to verify that the whole list is unique, and not moduleName. I mean, that no name can appear in it more then once

qbarrand: Absolutely. Now looking for duplicates.

api/v1beta1/module_webhook.go
yevgeny-shnaidman: in case user uses raw and modules order list, then we need to make sure that the first module on the raw command is also the first on the list. Maybe simplier just to block the modules order list in case user uses raw?

qbarrand: I would in fact not check anything related to the module order list if raw args are used.

PR #389, opened 2023-05-02
Changing InTree kernel module removal

yevgeny-shnaidman:
Currently we support removing the old in-tree version of the kernel module that customer wants to insert. This PR changes the boolean flag to string, allowing user to specify the exact name of the in-tree kernel module to remove. It can be the same kernel module name as to be inserted, or a completely different kernel module name Changes:
1) Module API now has InTreeModuleTo remove string instead of
   InTreeRemoval boolean
2) changing the code and unit-test


hershpa:
> @hershpa this PR makes it possible to configure the name of the module to remove, instead of re-using the `moduleName`. Does this work for you?

Yes, thanks!

PR #392, opened 2023-05-02
Hub & Spoke: Expand and replace Module kernel mappings with literal ones before applying the ManifestWork

mresvanis:
The PR adds a `ModuleSpec` processing step before creating the respective `ManifestWork` CR for a `ManagedCluster`, which in turn leads to the respective `Module` CR getting applied on the Spokes.

This step includes parsing the `ManagedClusterModule.Spec.ModuleSpec.ModuleLoader.Container.KernelMappings`, in order to replace them with literal ones and define the respective container images with their digests (instead of their tags). The latter is necessary for the Spokes to be able to pull the `ModuleLoader` container images also in a disconnected environment (where registry images are mirrored and digests are necessary to fetch the mirrored images successfully).

This change also ensures that kernel mappings are processed in sorted order every time (by sorting the kernel versions found in its Spoke cluster's `ClusterClaims`), in order to not trigger additional reconciliation loops on the Spokes just because the kernel mappings changed ordering and not in any other way. 

In order to ensure that we don't overwhelm the registry when fetching the image digests, a small cache was added. For example, if we have 100 targeted Spokes, all with the same kernel version, KMM Hub will fetch the image digest once and cache it, so that the rest 99 `ManifestWork`s will leverage the latter. This cache items expire after 10 minutes and the expired items are automatically cleaned up every 10 minutes.


internal/cluster/cluster.go
```diff
@@ -140,6 +145,97 @@ func (c *clusterAPI) BuildAndSign(
...
+		)
+
+		mapping := c.mappingFromModuleLoaderData(mld)
+		mappings = append(mappings, mapping)
+
+		digest, err := c.imageDigestForModuleLoaderData(ctx, mld)
+		if err != nil {
+			kernelVersionLogger.Info("skipping image tag replacement: %v", err)
+			continue
+		}
```
qbarrand, line 195: Should we do this before appending? We would not have to modify the latest mapping with `mappings[len(mappings)-1]` which feels unnecessary to me.

mresvanis: I have updated it accordingly, I hope it's more readable now :)

controllers/hub/managedclustermodule_reconciler.go
```diff
@@ -113,6 +113,12 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, req ctrl
 			continue
 		}
 
+		err = r.clusterAPI.ExpandAndReplaceKernelMappings(clusterCtx, mcm, cluster)
+		if err != nil {
+			logger.Error(err, "failed to expand and replace kernel mappings; skipping ManiefstWork reconciliation")
```
qbarrand, line 118:
```suggestion
			logger.Error(err, "failed to expand and replace kernel mappings; skipping ManifestWork reconciliation")
```

internal/cluster/cluster.go
```diff
@@ -140,6 +147,99 @@ func (c *clusterAPI) BuildAndSign(
...
+	}
+
+	logger := log.FromContext(ctx)
+
+	kernelVersions, err := c.kernelVersions(cluster)
+	if err != nil {
+		return err
+	}
+
+	mappings := make([]kmmv1beta1.KernelMapping, 0, len(kernelVersions))
```
qbarrand, line 171: Why do we need this intermediate variable?

internal/module/helper.go
```diff
@@ -40,6 +40,31 @@ func ShouldBeSigned(mld *api.ModuleLoaderData) bool {
...
+	var registryAuthGetter auth.RegistryAuthGetter
+	if mld.ImageRepoSecret != nil {
+		registryAuthGetter = auth.NewRegistryAuthGetter(client, types.NamespacedName{
+			Name:      mld.ImageRepoSecret.Name,
+			Namespace: namespace,
+		})
+	}
+
+	tlsOptions := mld.RegistryTLS
+	digest, err := reg.GetDigest(ctx, imageName, tlsOptions, registryAuthGetter)
```
qbarrand, line 60:
```suggestion
	digest, err := reg.GetDigest(ctx, imageName, mld.RegistryTLS, registryAuthGetter)
```

internal/registry/registry.go
```diff
@@ -91,6 +92,19 @@ func (r *registry) VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion,
...
 
+func (r *registry) GetDigest(ctx context.Context, image string, tlsOptions *kmmv1beta1.TLSOptions, registryAuthGetter auth.RegistryAuthGetter) (string, error) {
+	pullConfig, err := r.getPullOptions(ctx, image, tlsOptions, registryAuthGetter)
+	if err != nil {
+		return "", fmt.Errorf("failed to get pull options for image %s: %w", image, err)
+	}
+
+	digest, err := crane.Digest(image, pullConfig.authOptions...)
+	if err != nil {
+		return "", fmt.Errorf("failed to get digest for image %s: %w", image, err)
```
qbarrand, line 103: Do we really need `%w` here?
```suggestion
		return "", fmt.Errorf("failed to get digest for image %s: %v", image, err)
```

internal/registry/registry.go
```diff
@@ -91,6 +92,19 @@ func (r *registry) VerifyModuleExists(layer v1.Layer, pathPrefix, kernelVersion,
 	return err == nil
 }
 
+func (r *registry) GetDigest(ctx context.Context, image string, tlsOptions *kmmv1beta1.TLSOptions, registryAuthGetter auth.RegistryAuthGetter) (string, error) {
+	pullConfig, err := r.getPullOptions(ctx, image, tlsOptions, registryAuthGetter)
+	if err != nil {
+		return "", fmt.Errorf("failed to get pull options for image %s: %w", image, err)
```
qbarrand, line 98: Do we really need %w here?
```suggestion
		return "", fmt.Errorf("failed to get pull options for image %s: %v", image, err)
```

internal/cluster/cluster.go
```diff
@@ -140,6 +147,99 @@ func (c *clusterAPI) BuildAndSign(
...
+func (c *clusterAPI) ExpandAndReplaceKernelMappings(
+	ctx context.Context,
+	mcm *hubv1beta1.ManagedClusterModule,
+	cluster clusterv1.ManagedCluster) error {
+
+	modSpec := mcm.Spec.ModuleSpec
+	mod := &kmmv1beta1.Module{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      mcm.Name,
+			Namespace: c.namespace,
```
qbarrand, line 159: Could we avoid setting this?

qbarrand (commented):
Awesome work, thank you!

internal/registry/registry.go
mresvanis: Hm, you're probably right, we currently don't and most probably won't leverage the error wrapping for the callers of this API. I was just following the rest of this API's source code to be consistent. Should I change only these occurrences I've added here?

qbarrand: I think so. I would not make unrelated changes to the existing API in this PR.

controllers/hub/managedclustermodule_reconciler.go
```diff
@@ -113,6 +113,12 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, req ctrl
 			continue
 		}
 
+		kernelVersions, err := r.clusterAPI.KernelVersions(cluster)
+		if err != nil {
```
ybettan, line 117:
```suggestion
		if kernelVersions, err := r.clusterAPI.KernelVersions(cluster); err != nil {
```

If `err` isn't used later on, can we reduce its scope?

internal/cache/cache.go
```diff
@@ -0,0 +1,111 @@
...
+	if !found {
+		c.mu.RUnlock()
+		return nil, false
+	}
+	if item.Expiration > 0 {
+		if time.Now().UnixNano() > item.Expiration {
+			c.mu.RUnlock()
+			return nil, false
+		}
+	}
```
ybettan, line 95: Wondering why do we need this "double check". Shouldn't we treat all available items as valid until `DeleteExpired` is called?

internal/cluster/cluster.go
```diff
@@ -153,12 +155,26 @@ func (c *clusterAPI) GarbageCollectBuildsAndSigns(ctx context.Context, mcm hubv1
...
+		if clusterClaim.Name != constants.KernelVersionsClusterClaimName {
+			continue
+		}
+
+		kernelVersions := strings.Split(clusterClaim.Value, "\n")
+		sort.Strings(kernelVersions)
+
+		return kernelVersions, nil
+	}
+	return nil, errors.New("KMM ClusterClaim not found")
```
ybettan, line 169:
```suggestion
	return nil, fmt.Errorf("%s ClusterClaim not found", constants.KernelVersionsClusterClaimName)
```

internal/cache/cache.go
```diff
@@ -0,0 +1,111 @@
...
+//go:generate mockgen -source=cache.go -package=cache -destination=mock_cache.go
+
+type Cache interface {
+	DeleteExpired()
+	Get(key string) (interface{}, bool)
+	Set(key string, value interface{})
+	StopCleanup()
+}
+
+type Item struct {
```
qbarrand, line 17: Does this type need to be exported?

internal/manifestwork/manifestwork.go
```diff
@@ -175,3 +194,92 @@ func (mwg *manifestWorkGenerator) getOwnedManifestWorks(ctx context.Context, mcm
...
+			Namespace: mcm.Spec.SpokeNamespace,
+		},
+		Spec: mcm.Spec.ModuleSpec,
+	}
+
+	logger := log.FromContext(ctx)
+
+	mappings := make([]kmmv1beta1.KernelMapping, 0, len(kernelVersions))
+
+	for _, kernelVersion := range kernelVersions {
```
ybettan, line 215: Same question as before. Should we iterate over `mcm.Spec.ModuleSpec.ModuleLoader.Container.KernelMappings[]` instead?

internal/cache/cache.go
```diff
@@ -0,0 +1,111 @@
...
+	Object interface{}
+
+	Expiration int64
```
qbarrand, line 20: How about making this a `time.Time`? It would simplify some manipulations that currently use `UnixNano()`.

internal/manifestwork/manifestwork.go
```diff
@@ -175,3 +194,92 @@ func (mwg *manifestWorkGenerator) getOwnedManifestWorks(ctx context.Context, mcm
...
+		kernelVersion := strings.TrimSuffix(kernelVersion, "+")
+
+		kernelVersionLogger := logger.WithValues(
+			"kernel version", kernelVersion,
+		)
+
+		mld, err := mwg.kernelAPI.GetModuleLoaderDataForKernel(mod, kernelVersion)
+		if err != nil {
+			kernelVersionLogger.Info("no suitable container image found; skipping kernel version")
```
ybettan, line 224: Should we consider this as a warning log? If we wish to consider this as a warning we can use `utils.WarnString()`

internal/cache/suite_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/kubernetes-sigs/kernel-module-management/internal/test"
+)
+
+var scheme *runtime.Scheme
```
qbarrand, line 14: I don't think this variable is needed in the `cache` package?

internal/cluster/cluster.go
qbarrand: I think the KMM ClusterClaim name can be considered an implementation detail; not sure why we should log it.

internal/manifestwork/manifestwork.go
```diff
@@ -175,3 +194,92 @@ func (mwg *manifestWorkGenerator) getOwnedManifestWorks(ctx context.Context, mcm
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      mcm.Name,
+			Namespace: mcm.Spec.SpokeNamespace,
+		},
+		Spec: mcm.Spec.ModuleSpec,
+	}
+
+	logger := log.FromContext(ctx)
+
+	mappings := make([]kmmv1beta1.KernelMapping, 0, len(kernelVersions))
```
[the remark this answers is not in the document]
qbarrand: We are doing this to avoid a situation where several mappings would be added to the `ManifestWork` with the same `regex` field.

> What happens if there is an upgrade on the spoke but the mapping wasn't added yet to the module by the hub?

We can make it clear in the docs that it should be added to the `ManagedClusterModule` first.

internal/manifestwork/manifestwork.go
```diff
@@ -175,3 +194,92 @@ func (mwg *manifestWorkGenerator) getOwnedManifestWorks(ctx context.Context, mcm
...
+			"image", mld.ContainerImage,
+		)
+
+		mapping := mwg.mappingFromModuleLoaderData(mld)
+
+		digest, err := mwg.imageDigestForModuleLoaderData(ctx, mld)
+		if err != nil {
+			kernelVersionLogger.Error(err, "skipping image tag replacement")
+		}
+		if digest != "" {
```
ybettan, line 238: What happens if `digest == ""`? We keep the tag?

internal/manifestwork/manifestwork.go
```diff
@@ -175,3 +194,92 @@ func (mwg *manifestWorkGenerator) getOwnedManifestWorks(ctx context.Context, mcm
...
+	}
+
+	return mappings
+}
+
+func (mwg *manifestWorkGenerator) mappingFromModuleLoaderData(mld *api.ModuleLoaderData) kmmv1beta1.KernelMapping {
+	return kmmv1beta1.KernelMapping{
+		ContainerImage:       mld.ContainerImage,
+		Literal:              mld.KernelVersion,
+		InTreeModuleToRemove: mld.InTreeModuleToRemove,
```
ybettan, line 257: Don't we need all the other fields such as `RegistryTLS`?

internal/manifestwork/manifestwork.go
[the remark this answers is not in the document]
mresvanis: If there is a spoke OCP upgrade (i.e. a new kernel is available in one or more of its nodes), the following happens:

* the spoke KMM `ClusterClaim` on the `ManagedCluster.Status` is updated by KMM Managed (on the spoke) 
* [as a result](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/controllers/hub/managedclustermodule_reconciler.go#L164-L170) KMM Hub will reconcile the `ManagedClusterModule` targeting that spoke
* this reconciliation will add the new literal `KernelMapping`(s) to the `Module` that will be deployed to the spoke 

Does this make sense?

internal/cache/cache.go
mresvanis: No, we don't need it exported. It's just a leftover while testing this code, I'll change it to an non-exported field. Thanks for spotting this :+1:

internal/cache/suite_test.go
mresvanis: Absolutely not, too much copy-paste ... I'll remove it. Thank you :+1:

controllers/hub/managedclustermodule_reconciler.go
ybettan: I see that you are using this structure everywhere else in the PR so if this is personal preference then please ignore my comment :)

internal/manifestwork/manifestwork.go
```diff
@@ -175,3 +194,92 @@ func (mwg *manifestWorkGenerator) getOwnedManifestWorks(ctx context.Context, mcm
...
+	}
+}
+
+func (mwg *manifestWorkGenerator) imageDigestForModuleLoaderData(ctx context.Context, mld *api.ModuleLoaderData) (string, error) {
+	ref, err := name.ParseReference(mld.ContainerImage, name.WithDefaultRegistry(""))
+	if err != nil {
+		return "", fmt.Errorf("could not parse the container image name: %v", err)
+	}
+
+	// only digests contain a ':'
```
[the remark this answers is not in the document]
mresvanis: Thank you for pointing this out, as this comment seems to be a bit ambiguous. 

`ref.Identifier()` refers to the part of the image name that would either be a tag `a-tag` or a digest `sha256:<a-digest-hash>` and not to the whole image name, e.g. `image-name:tag` or `image-name@sha256:digest`.

Does this make sense?

internal/manifestwork/manifestwork.go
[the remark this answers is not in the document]
ybettan: > Does this make sense?

It does. Thanks.

internal/cluster/cluster.go
mresvanis: I agree with @qbarrand that the `ClusterClaim` name is an implementation detail, which we should avoid logging. 

The reason for this is that the user has no control over it. KMM Managed (on the spoke) creates this KMM `ClusterClaim` and this log is there to point users to KMM Managed when they see this error. WDYT?

qbarrand: > We can have a bunch of clusterClaims but we return an error only if we didn't find a specific clusterClaim (which is the kernel version).

Exactly. The error message says that we could not find KMM's entry in the cluster claims. I don't think the user needs to know more.

internal/manifestwork/manifestwork.go
qbarrand: Why do we need `RegistryTLS`? I believe it's only used to build or sign.

internal/cache/cache.go
```diff
@@ -0,0 +1,111 @@
...
+		Expiration: expiration,
+	}
+	c.mu.Unlock()
+}
+
+// Get fetches an item from the cache. It returns the item or nil and a bool indicating
+// whether the key was found.
+func (c *cache) Get(k string) (interface{}, bool) {
+	c.mu.RLock()
+	item, found := c.items[k]
```
yevgeny-shnaidman, line 85: you can probably use defer here, instead of calling c.mu.RUnlock, it will also be safer

internal/cache/cache.go
mresvanis: We could definitely do that. The reason I chose an `int64` is to keep memory consumption as low as possible, but I see your point. Does it make sense memory wise or should we just refactor it to `time.Time` as you suggested?

internal/cache/cache.go
mresvanis: The reasoning behind this check is that I consider the expired items as expired, even if they are still in the cache until `DeleteExpired` is called. For example, item "a" has already expired but the `DeleteExpired` has not yet been called and a `Get("a")` arrives, which will not get the expired item back. WDYT?

internal/cache/cache.go
mresvanis: I considered using `defer`, but I thought I'd keep this bit faster (but maybe I should have measured `defer`'s performance before doing such an optimisation). I'll revert this to using a `defer`, since it's safer and more readable.

internal/cache/cache.go
qbarrand: It [appears to make little sense](https://stackoverflow.com/questions/56963226/how-many-bytes-are-in-a-golang-time-object). I would use `time.Time` and switch to `int64` if we ever have a problem with memory consumption.

internal/cache/cache.go
```diff
@@ -0,0 +1,105 @@
...
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	item, found := c.items[k]
+	if !found {
+		return nil, false
+	}
+	if time.Now().After(item.Expiration) {
+		return nil, false
+	}
```
qbarrand, line 52:
```suggestion
	item, found := c.items[k]
	if !found || time.Now().After(item.Expiration) {
		return nil, false
	}
```

internal/cache/cache.go
ybettan: Yea, I got it from the code, just didn't expect it because we have the `DeleteExpired` method so I assume that everything is considered non-expired as long as it is in the cache but this is an implementation detail so if you prefer it that way I don't mind.

internal/cluster/cluster.go
ybettan: Any error message that replaces `ClusterClaim` with `kernel-version ClusterClaim` sounds good to me. I think it will help folks which are not expert in hub&spoke (such as myself when I have written the e2e test) which may have a rough idea of what a clusterClaim is and specifying which kind of clusterClaim helps.

Will you be OK with `kernel-version ClusterClaim not found`?

internal/manifestwork/manifestwork.go
ybettan: I don't have a strong opinion about it but I am just wondering if it won't be confusing why some of the images were manipulated while other where skipped. I can see pros and cons for both approaches (returning an error on the hub or just letting the spoke fail) so please proceed with the option that you see best fits.

internal/manifestwork/manifestwork.go
ybettan: What about `preflight`? Isn't it possible that a customer will run preflight only on the spoke?

internal/manifestwork/manifestwork.go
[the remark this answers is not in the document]
ybettan: Makes a lot of sense now :) 
Can we update the comment to something like:
```
only digests contain a ':' in the identifier
```
?

internal/manifestwork/manifestwork.go
qbarrand: Not as of today. There is no facility to create a `PreflightValidation` resource on the Spoke.

mresvanis: Currently there is no support for `Preflight` on the spokes via the `ManagedClusterModule` CRD. If I'm not mistaken, the user has to create their own `ManifestWork` CR, which includes the desired `Preflight` CR.

internal/manifestwork/manifestwork.go
mresvanis: We are looping over the kernel versions instead of the kernel mappings, because a kernel mapping can be also defined using a regexp. That way one kernel mapping can lead to matching more than one kernel versions. What we would like to have in the resulting `Module` that will be applied on the spoke, is its `KernelMappings` array to include only literal kernel mappings, i.e. one kernel mapping per kernel version. WDYT?

internal/manifestwork/manifestwork.go
mresvanis: I would say that this is just informative, we can legitimately have a kernel version that shouldn't be matched by the `Module`'s kernel mappings. I wouldn't consider it a warning, as it's not a non-fatal error. WDYT?

internal/manifestwork/manifestwork.go
ybettan: That makes sense.

internal/manifestwork/manifestwork.go
ybettan: Ack.

internal/manifestwork/manifestwork.go
ybettan: @mresvanis  I meant, isn't it possible that a user will directly apply a `PreflightValidation` resource directly on the spoke.

@qbarrand Thanks for the answer.

ybettan:
/approve
/hold
`tide` isn't enforcing the `e2e` job anymore because it was renamed from `in-cluster-build`.
Please unhold if there is no need to wait for the linux headers fix before merging this one.

ybettan:
```
kubectl -n kmm-operator-system wait --for=condition=Ready --timeout=1m pod/mod-example-9bad31a8552e1b47-z5gcg
```

I see that the manifestWork was created successfully so my best guess is that something went wrong with the image manipulation and the container couldn't pull the image.

mresvanis:
> ```
> kubectl -n kmm-operator-system wait --for=condition=Ready --timeout=1m pod/mod-example-9bad31a8552e1b47-z5gcg
> ```
> 
> I see that the manifestWork was created successfully so my best guess is that something went wrong with the image manipulation and the container couldn't pull the image.

There was a bug in the image `tag` replacement code, Ι fixed it for good (hopefully) :) Thanks for the support Yoni :+1:

PR #401, opened 2023-05-08
Simplify the e2e-hub flow

qbarrand:
Reuse components from the `e2e` flow.

### DIFF
--- a/ci/managedclustermodule-kmm-ci-build-sign.yaml
+++ b/ci/managedclustermodule-kmm-ci-build-sign.yaml
@@ -60,122 +60,10 @@ subjects:
     kind: Policy
     name: allow-klusterlet-deploy-kmm-modules
 ---
-apiVersion: v1
-kind: Secret
-metadata:
-  name: fake-modprobe-signing-key-pub
-  namespace: kmm-operator-system
-type: Opaque
-stringData:
-  cert: |
-    -----BEGIN CERTIFICATE-----
-    MIIE8DCCAtigAwIBAgIUSRZJsM/2jzb7fm1PdqB/bjUOAtEwDQYJKoZIhvcNAQEN
-    BQAwEjEQMA4GA1UEAwwHTW9kdWxlczAgFw0yMzAxMjQxMDM3MzNaGA8yMTIyMTIz
-    MTEwMzczM1owEjEQMA4GA1UEAwwHTW9kdWxlczCCAiIwDQYJKoZIhvcNAQEBBQAD
-    ggIPADCCAgoCggIBAImPO0SLpA7l3/tOv/+ZwUm+o7W8c2BtDjmoVjLfWP0nwaV1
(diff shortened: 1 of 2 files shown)


ci/managedclustermodule-kmm-ci-build-sign.yaml
```diff
@@ -195,12 +83,12 @@ spec:
             baseImageRegistryTLS:
               insecure: true
             dockerfileConfigMap:
-              name: mod-example
+              name: kmm-kmod-dockerfile
```
ybettan, line 86: I don't think this is a good idea.

I though about it as well and tried it myself.
I don't think we should have any actual build as this part is already tested in the `e2e` job.

I think we should keep the dummy job to isolate as much as we can each workflow and reduce the duplicated parts as much as possible.

.github/workflows/e2e.yaml
```diff
@@ -145,9 +145,6 @@ jobs:
 
     steps:
 
-      - name: Install curl on the machine
```
ybettan, line 148: How are you installing `clusteradm` without `curl`?
https://github.com/kubernetes-sigs/kernel-module-management/pull/401/files#diff-448746927799f21be03f629be184a58511d510c0496399d1907b9e1954be9cc1R6

qbarrand: It's already installed on the node.

ci/managedclustermodule-kmm-ci-build-sign.yaml
qbarrand: But why duplicate keys and Dockerfile when we can reuse those of the e2e job?

ybettan: Regarding the keys, I agree.

Regarding the dockerfile, I have mixed feelings.
Re-using it will duplicate the building test across `e2e` and `e2e-hub`. I think it would be nice if the hub test is only testing the hub functionality in the manner of just checking that all the artifacts are created correctly but not actually have a real build or real.

With that said, I can also see benefits in re-using the current dockerfile. 

WDYT?

.github/workflows/e2e.yaml
ybettan: This is wired, before adding this installation, I was getting `curl: command not found` during the installation of `clusteradm`....

ci/managedclustermodule-kmm-ci-build-sign.yaml
qbarrand: I don't see any downside in re-using the same Dockerfile. Both tests having that part in common may make troubleshooting easier.

.github/workflows/e2e.yaml
qbarrand: Could be. If it was not there I would anyway switch to `wget` to avoid installing anything.

ybettan:
SGTM.

Issue #400, opened 2023-05-08
Provide `KERNEL_FULL_VERSION` as a default build argument


qbarrand:
When creating builds, we provide the `KERNEL_VERSION` build argument, but not `KERNEL_FULL_VERSION` which is substituted in the `containerImage` field. This creates a lot of confusion.
We should provide `KERNEL_FULL_VERSION` as a default build argument. The value should be identical to that of `KERNEL_VERSION`.

PR #402, opened 2023-05-09
Add KERNEL_FULL_VERSION as a default build argument

hershpa:
Fixes (#400)
Signed-off-by: Hersh Pathak <hersh.pathak@intel.com>


yevgeny-shnaidman:
A general question: we now have KERNEL_VERSION and KERNEL_FULL_VERSION as build arguments, and they both have the same value. Why not just keep one of them?

qbarrand:
> Why not just keep one of them?

We don't want to remove `KERNEL_VERSION` as it would break existing Dockerfiles. Maybe in 2.0, although I don't see the urgency of doing that.

yevgeny-shnaidman:
> > Why not just keep one of them?
> 
> We don't want to remove `KERNEL_VERSION` as it would break existing Dockerfiles. Maybe in 2.0, although I don't see the urgency of doing that.

So why not just keep KERNEL_VERSION? The existing Dockerfiles are using it, and the new Dockerfiles can start using it, no?

qbarrand:
> So why not just keep KERNEL_VERSION?

Right now the `containerImage` field replaces `KERNEL_FULL_VERSION` and we provide the `KERNEL_VERSION` build argument. This is very confusing for users. See #400

ybettan:
Can we add a tech debt for removing `KERNEL_VERSION` build arg in [2.0 milestone](https://github.com/kubernetes-sigs/kernel-module-management/milestone/2)?

qbarrand:
@hershpa please rebase this PR on top of `main`.

PR #403, opened 2023-05-09
Use ubuntu-20.04 explicitly in CI e2e tests instead of ubuntu-latest

mresvanis:
This PR replaces [ubuntu-latest](https://github.com/actions/runner-images#available-images) with `ubuntu-20.04` explicitly in the CI e2e Github actions and Dockerfiles. 

When using `ubuntu-latest`, which is currently pointing to `ubuntu-22.04`, we have kernel version `5.15.0-1036-azure` on the VM, but there is no `linux-headers-5.15.0-1036-azure` package in the ubuntu-22.04 repos. So we we get an error when building the CI kmod using the `ubuntu:latest` (also pointing currently to `ubuntu:22.04`) container image, because the respective linux headers cannot be installed. With Ubuntu 20.04 LTS  this is not an issue.  

Example CI error:

```log
E: Unable to locate package linux-headers-5.15.0-1036-azure
E: Couldn't find any package by glob 'linux-headers-5.15.0-1036-azure'
E: Couldn't find any package by regex 'linux-headers-5.15.0-1036-azure'
error building image: error building stage: failed to execute command: waiting for process to exit: exit status 100
```

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -8,7 +8,7 @@ env:
 jobs:
 
   build-operator-image:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     name: Build the KMMO container image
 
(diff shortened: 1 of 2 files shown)


qbarrand:
Bypassing CI for this commit since further unrelated changes are needed for the pipeline to be green.

PR #404, opened 2023-05-09
Update CI e2e-hub test

mresvanis:
This change updates the CI e2e-hub test to wait for the `clusteradm join` command to succeed asynchronously. Waiting for it synchronously sometimes leads to errors like the following:
```log
+ clusteradm join --hub-token *** --hub-apiserver https://192.168.49.2:8443 --wait --cluster-name minikube
  Error: no matches for kind "Klusterlet" in version "operator.open-cluster-management.io/v1"
```
This errors comes from [clusteradm](https://github.com/open-cluster-management-io/clusteradm/blob/95f9d75af1b20d6c1f04298a8e01b5941cd7a867/pkg/cmd/join/exec.go#L261-L265) itself and the only workaround we can do is run the `clusteradm join` command and wait/watch asynchronously for the relevant Kubernetes resources, that indicate a successful join, to be available.

This PR also removes the unneeded `curl` installation in the same CI workflow. It also removes the `apt-get upgrade` command that could lead to more reliability issues on the CI e2e steps.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -145,9 +145,6 @@ jobs:
 
     steps:
 
-      - name: Install curl on the machine
-        run: sudo apt update -y && sudo apt upgrade -y && sudo apt install -y curl
-
       - name: Checkout
         uses: actions/checkout@v3
 
--- a/ci/prow/e2e-hub-spoke-incluster-build
+++ b/ci/prow/e2e-hub-spoke-incluster-build
@@ -24,11 +24,16 @@ hub_api_server=$(kubectl config view -o yaml | yq '.clusters[] | select(.. | .na
 clusteradm join \
   --hub-token ${token} \
   --hub-apiserver ${hub_api_server} \
-  --wait \
   --cluster-name ${MINIKUBE}
-kubectl wait --for=condition=Available deployment/klusterlet-registration-agent -n open-cluster-management-agent
-clusteradm accept --clusters ${MINIKUBE}
 
+# Wait for the join command to succeed
+kubectl wait --for=condition=Established crd managedclusters.cluster.open-cluster-management.io
+timeout 60 sh -c "until kubectl get managedcluster ${MINIKUBE}; do sleep 2; done"
+timeout 30 sh -c "until kubectl get csr -l open-cluster-management.io/cluster-name=${MINIKUBE} | grep Pending; do sleep 2; done"
+timeout 1m bash -c "until kubectl get csr -l open-cluster-management.io/cluster-name=${MINIKUBE} -o json | jq -er \".items[].metadata.name\"; do sleep 2; done"
+kubectl wait --for=condition=Available deployment/klusterlet-registration-agent -n open-cluster-management-agent
+
+clusteradm accept --clusters ${MINIKUBE}
 
 # Installing required addons to the hub/spoke clusters
 clusteradm install hub-addon --names application-manager


ci/prow/e2e-hub-spoke-incluster-build
```diff
@@ -17,6 +17,7 @@ kubectl wait --for=condition=Available deployment/cluster-manager-placement-cont
 kubectl wait --for=condition=Available deployment/cluster-manager-registration-controller -n open-cluster-management-hub
 kubectl wait --for=condition=Available deployment/cluster-manager-registration-webhook -n open-cluster-management-hub
 kubectl wait --for=condition=Available deployment/cluster-manager-work-webhook -n open-cluster-management-hub
+kubectl wait --for condition=established crd managedclusters.cluster.open-cluster-management.io
```
qbarrand, line 20: Not sure if it matters but:
```suggestion
kubectl wait --for condition=Established crd managedclusters.cluster.open-cluster-management.io
```

mresvanis:
I need to thoroughly verify that the current `kubectl wait` fix is what we actually need.

/hold

mresvanis:
In fact the error we bumped into in our CI e2e-hub:

```
+ clusteradm join --hub-token *** --hub-apiserver https://192.168.49.2:8443 --wait --cluster-name minikube
  Error: no matches for kind "Klusterlet" in version "operator.open-cluster-management.io/v1"
```

is most probably a `clusteradm join` issue, as `clusteradm join` registers the `Klusterlet` CRD and should wait for it:
https://github.com/open-cluster-management-io/clusteradm/blob/95f9d75af1b20d6c1f04298a8e01b5941cd7a867/pkg/cmd/join/exec.go#L261-L265

mresvanis:
/unhold

The workaround/fix should be working as expected and I believe we have our root cause.

ci/prow/e2e-hub-spoke-incluster-build
```diff
@@ -24,11 +24,15 @@ hub_api_server=$(kubectl config view -o yaml | yq '.clusters[] | select(.. | .na
 clusteradm join \
   --hub-token ${token} \
   --hub-apiserver ${hub_api_server} \
-  --wait \
   --cluster-name ${MINIKUBE}
+
+# Wait for the join command to succeed
+kubectl wait --for=condition=Established crd managedclusters.cluster.open-cluster-management.io
+timeout 60 sh -c "while ! kubectl get managedcluster ${MINIKUBE} >/dev/null 2>&1; do sleep 2; done"
+timeout 30 sh -c "while ! kubectl get csr -l open-cluster-management.io/cluster-name=${MINIKUBE} 2> /dev/null | grep Pending >/dev/null 2>&1; do sleep 2; done"
```
qbarrand, line 32: I think using `until` would be more readable. I would also not suppress stdout / stderr output, to make troubleshooting easier:
```suggestion
timeout 60 sh -c "until kubectl get managedcluster ${MINIKUBE}; do sleep 2; done"
timeout 30 sh -c "until kubectl get csr -l open-cluster-management.io/cluster-name=${MINIKUBE} | grep Pending; do sleep 2; done"
```
Also, could we target a specific path for `Pending` instead of grepping?

.github/workflows/e2e.yaml
```diff
@@ -145,9 +145,6 @@ jobs:
 
     steps:
 
-      - name: Install curl on the machine
-        run: sudo apt update -y && sudo apt upgrade -y && sudo apt install -y curl
```
ybettan, line 149: Last time I checked, `curl` wasn't on the machine and the installation of `clusteradm` was failing because of that.
What has changed that we don't need it anymore?

mresvanis: I think `curl` is already installed in both [Ubuntu 22.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#installed-apt-packages) and [Ubuntu 20.04](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#installed-apt-packages). Could it be that it was another image that you have been testing with, that doesn't include `curl`?

ci/prow/e2e-hub-spoke-incluster-build
mresvanis: Of course, `until` is quite more readable and I should have conformed to the rest of the script :+1: 

In addition, I replaced `grep`, but since there is no `jsonpath` to indicate that the CR's `condition=Pending` (as it's `Pending` when the `CR.status` is empty), we have no other way to check for this CR's existence other than using `jq` + `select` + `match`. I think that this is still better than just `grep`-ping :)

qbarrand:
Force merging to unlock all PRs depending on this.

PR #420, opened 2023-05-24
Upgrade to controller-runtime v0.15.0

qbarrand:
Remove the logger in the filter types as those functions now get a context.
Upgrade to Kubernetes v1.27.2.
Bump other dependencies.


config/manager-hub/controller_manager_config.yaml
```diff
@@ -1,12 +1,6 @@
-apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
-kind: ControllerManagerConfig
```
ybettan, line 2: Why are we removing `apiVersion` and `kind`? Doesn't it exists in the new version?

config/manager/controller_manager_config.yaml
```diff
@@ -1,12 +1,6 @@
-apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
-kind: ControllerManagerConfig
```
ybettan, line 2: Why are we removing `apiVersion` and `kind`? Doesn't it exists in the new version?

qbarrand: We unmarshal the data ourselves, so no need for these fields anymore.

config/manager-hub/controller_manager_config.yaml
qbarrand: We unmarshal the data ourselves, so no need for these fields anymore.

cmd/manager-hub/main.go
```diff
@@ -87,21 +87,22 @@ func main() {
 
 	setupLogger.Info("Creating manager", "git commit", commit)
 
-	options := ctrl.Options{Scheme: scheme}
-
-	options, err = options.AndFrom(ctrl.ConfigFile().AtPath(configFile))
+	cfg, err := config.ParseFile(configFile)
```
yevgeny-shnaidman, line 90: had the new controller-runtime version removed support for parsing option? If not, then why not use its code?

qbarrand: It has deprecated its config parser: [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/895), [Go doc](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.0/pkg/config#ControllerManagerConfiguration)
Users are asked to move to their own config management system, this is what this PR does.

PR #421, opened 2023-05-24
Add a proposal for a one-DaemonSet KMM

qbarrand:
Add an enhancement proposal that addresses a number of pain points in the way KMM loads kernel modules in v1.

### DIFF
--- a/docs/enhancements/0-single-daemonset.md
+++ b/docs/enhancements/0-single-daemonset.md
@@ -1,0 +1,278 @@
+# Single DaemonSet for KMM
+
+Authors: @qbarrand, @mresvanis, @yevgeny-shnaidman
+
+## Introduction
+
+This enhancement aims at eliminating several pain points related to how KMM loads kernel modules on the nodes.
+
+### Resource utilization
+
+KMM currently creates one DaemonSet per `Module` and compatible kernel version found in the cluster.
+Formally, the number of DaemonSets created $S$ is defined by:
+$$S = \sum_{i=1}^nk_i$$
+
+where:
+- $n$ is the number of `Modules` in the cluster, and;
+- $k_i$ is the number of kernel versions running in the cluster that `Module` $i$ is compatible with.
+
+A big $S$ leads to unnecessary workload for the cluster:
+- on nodes, one ModuleLoader Pod is created per `Module`. Although those just sleep after the `modprobe`
+  `postStart` hook, each running pod incurs an overhead for the container runtime and the kubelet;
+- on the control-plane, a higher number of DaemonSets and Pods creates more events for kube-controller-manager, which
+  trigger more reconciliations and status updates.
+
+### Reliability
+
+Running user-provided ModuleLoader images as DaemonSets makes it difficult for KMM to determine if a module was
+successfully unloaded or not.  
+KMM v1 uses the `preStop` hook to run `modprobe -r` and unload modules.
+`preStop` hook failures cannot prevent the Pod from being terminated.
+If `modprobe` cannot to unload the module for any reason (e.g. userspace applications still using it), then the Pod will
+be terminated anyway and the `kmm.node.kubernetes.io/$namespace.$module.ready` label will be removed from the Node, even
+though the module is still loaded.
+
+## Goals
+
+1. Reduce the overall KMM footprint
+2. Improve the reliability of module loading and unloading
+
+## Non-goals
+
+1. Add further control on the deployment of kmods on nodes
+
+## Design
+
+### KMM: one DaemonSet pulling all kmod images
+
+This proposal reduces the number of DaemonSets run by KMM to one.
+KMM is now composed of two components, communicating over the network:
+- the control-plane is the current operator Deployment;
+- the daemon pods belong to a new DaemonSet that is created and owned by the control-plane Deployment.
+
+After it is installed, the operator creates a DaemonSet targeting on all worker nodes.
+The daemon pods run a new KMM component that pulls kmod images, extracts them locally and loads / unloads kmods.
+
+```
+                                                     ┌─────────────────────────────────────────┐
+                                                     │ Random namespace                        │
+                                                     │                                         │
+                          gets                       │            ┌─────────────┐              │
+                        ┌────────────────────────────┼────────────► Pull secret │              │
+                        │                            │            └───────▲─────┘              │
+                        │         reconciles         │                    │                    │
+                        │   ┌────────────────────────┼────┐               │                    │
+                        │   │                        │    │               │                    │
+                        │   │                        │  ┌─▼──────┐        │                    │
+                        │   │                        │  │ Module ├────────┘                    │
+                        │   │                        │  └────────┘  references                 │
+                        │   │                        │                                         │
+┌───────────────────────┼───┼───────────────────┐    │                                         │
+│ KMM namespace         │   │                   │    │                                         │
+│                       │   │                   │    │                    ┌───────────┐        │
+│                       │   │                   │    │   ┌────────────────► Build Job │        │
+│                       │   │                   │    │   │                └─────┬─────┘        │
+│           ┌───────────┴───┴───┐   creates &   │    │   │                      │              │
+│           │ KMM control-plane │      owns     │    │   │                      │              │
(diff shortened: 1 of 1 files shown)


docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+Once the connection is established, the control-plane maintains a map of the desired state of each node with daemon Pod
+connections, allowing it to notify a specific Pod when that state is modified and some action is needed.  
+To reduce the amount of computation for each reconciliation, we may decide to persist the desired state of each node
+into a new Custom Resource.
+The `Module` and `Node` reconcilers would update that resources, and a new, separate reconciler would watch it and
+notify the corresponding daemon Pod when an update is required.
+
+#### Daemon Pods
+
+The DaemonSet for the daemon Pods is created and owned by the control-plane.
```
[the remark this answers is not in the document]
qbarrand: I don't think this should be an issue if we keep the daemon's footprint low, which you be the case if does nothing.
Maybe we can explore building a custom node affinity for the DaemonSet as you suggest, but I am a bit afraid of the complexity here, considering that we may have to take kernel versions into consideration.
A compromise could be to make the node selector for this DaemonSet configurable in the KMM config - we are already planning to allow custom tolerations there, to allow running the daemon on control-plane nodes.

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+### Daemon Pods running a small controller
+
+In that design, instead of delegating all communication with the API Server to the control-plane, each daemon Pod would
+run a small controller watching `Module` resources and their own Node to determine what kmod they have to download and
+load.
+Instead of communicating over a [gRPC] channel, the daemon Pods would notify the control-plane over the status of a new
+custom resource.  
+However the current `Module` API allows users to reference Secret objects as pull secrets in the `Module`'s namespace.
+The daemon pods are all running in the KMM namespace, and thus we would have to run them with a ServiceAccount that
+has permissions to `get` all Secrets in all namespaces.
```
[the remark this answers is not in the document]
qbarrand: Don't know if you mean one ServiceAccount per daemon Pod - I don't think that's doable with a DaemonSet.
If you mean only one ServiceAccount for all pods, we would need the control-plane to know when those pods start and stop pulling, which sounds rather complex.

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+## Design
+
+### KMM: one DaemonSet pulling all kmod images
+
+This proposal reduces the number of DaemonSets run by KMM to one.
+KMM is now composed of two components, communicating over the network:
+- the control-plane is the current operator Deployment;
+- the daemon pods belong to a new DaemonSet that is created and owned by the control-plane Deployment.
+
+After it is installed, the operator creates a DaemonSet targeting on all worker nodes.
```
ybettan, line 53:
```suggestion
After it is installed, the operator creates a DaemonSet targeting all worker nodes.
```

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+
+1. Add further control on the deployment of kmods on nodes
+
+## Design
+
+### KMM: one DaemonSet pulling all kmod images
+
+This proposal reduces the number of DaemonSets run by KMM to one.
+KMM is now composed of two components, communicating over the network:
+- the control-plane is the current operator Deployment;
```
ybettan, line 50: I assume that the `control-plane` component doesn't have to run on the master nodes right?! In this case, `control-plane` is a bit confusing. One can think we have a similar type of deployment as NFD does where some of the pods runs on master nodes.

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+Locally, daemon Pod should maintain a state with the same lifetime as the kernel modules it loads.
+It is desirable for this state to persist during application restarts, but not node reboots, because by definition all
+KMM-loaded kmods are gone as well.  
+We propose to write that state in a temporary filesystem such as `/run`, which:
+- is not garbage-collected at runtime like `/tmp` is;
+- is emptied during reboots.
+
+#### Connectivity
+
+The control-plane and the daemon Pods communicate over the network, using [gRPC] over TLS.
```
ybettan, line 144:
```suggestion
The control-plane and the daemon Pods communicate over the network, using [gRPC] over HTTP/2.  
```
?

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+KMM-loaded kmods are gone as well.  
+We propose to write that state in a temporary filesystem such as `/run`, which:
+- is not garbage-collected at runtime like `/tmp` is;
+- is emptied during reboots.
+
+#### Connectivity
+
+The control-plane and the daemon Pods communicate over the network, using [gRPC] over TLS.  
+Although connections are initiated by the daemon Pods, communication is bi-directional:
+- daemon Pods send request their config from and send events to the control-plane;
```
ybettan, line 146:
```suggestion
- daemon Pods request their config from and send events to the control-plane;
```

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+            │              │             │                                  │
+            └──────────────┴─────────────┴──────────────────────────────────┘
+                                      pull from
+```
+
+#### Control-plane
+
+The control-plane watches all resources in the cluster that are necessary to determine which module needs to be loaded
+on which node.
+In addition to its current duties such as build, signing and node labeling, it also listens for connections from daemon
```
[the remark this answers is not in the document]
yevgeny-shnaidman: also , the same goes for preflight

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+malicious purposes.
+We intend to run the daemon Pods with a dedicated ServiceAccount with zero API Server privileges.
+The only sensitive data that the daemon Pod will potentially be processing is pull secrets from other namespaces, which
+will only be used to pull the kmod image once and will not be persisted in any way.
+
+### Hub & Spoke: one proxy in each Spoke
+
+In Hub & Spoke setups, we introduce a third KMM component, the proxy, which is both a [gRPC] proxy for the daemon pods
+and a controller providing doing everything the Hub cannot do remotely:
+- label nodes when a module is successfully loaded;
```
ybettan, line 180: Why do we need a new component for that? The non-hub KMM knows how to do it already? Any reason we don't have the `KMM-control-plane` version for hub that will handle build/sign and one for the spoke that will labels nodes as we have today?

docs/enhancements/0-single-daemonset.md
```diff
@@ -0,0 +1,278 @@
...
+Daemon Pods run on all nodes.
+If one of those nodes is compromised, then an attacker could use the Pod's credentials against the API Server for
+malicious purposes.
+We intend to run the daemon Pods with a dedicated ServiceAccount with zero API Server privileges.
+The only sensitive data that the daemon Pod will potentially be processing is pull secrets from other namespaces, which
+will only be used to pull the kmod image once and will not be persisted in any way.
+
+### Hub & Spoke: one proxy in each Spoke
+
+In Hub & Spoke setups, we introduce a third KMM component, the proxy, which is both a [gRPC] proxy for the daemon pods
```
ybettan, line 178: So we stop using `ClusterClaims` and start using gRPC instead for everything needed?

docs/enhancements/0-single-daemonset.md
ybettan: According to the diagram, seems like `Proxy` is just KMM-control-plane for non hub clusters. Am I missing something?

yevgeny-shnaidman:
Daemonset is going to handle all capabilities of KMM V2 ( kernel module, epbf, kpatch etc'). It means that the security context of the Daemonset need to accommodate the  most demanding capability. That probably mean that its security context will need to be privileged, in order to accommodate , for example, epbf

qbarrand:
Superseded by #463

PR #430, opened 2023-06-05
Inject the webhook CA into the validated CRDs

qbarrand:
Cherry-pick of 3fd56ed5b8d1dcc973547b284a5f52b41b5355e0.

/cc @yevgeny-shnaidman @ybettan @mresvanis

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -18,8 +18,8 @@ patches:
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-#- path: patches/cainjection_in_modules.yaml
-#- path: patches/cainjection_in_managedclustermodules.yaml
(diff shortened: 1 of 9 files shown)


config/crd/kustomization.yaml
```diff
@@ -18,8 +18,8 @@ patches:
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
-#- path: patches/cainjection_in_modules.yaml
-#- path: patches/cainjection_in_managedclustermodules.yaml
+- path: patches/cainjection_in_modules.yaml
+#- path: patches/cainjection_in_preflightvalidations.yaml
```
yevgeny-shnaidman, line 22: why do we need it if it is commented out?

qbarrand: In case we ever want to enable it. When creating CRDs, `operator-sdk` adds those commented out lines. I fixed them and added the missing ones for consistency.

PR #434, opened 2023-06-06
additional fixes for ordered upgrade doc

yevgeny-shnaidman:


qbarrand:
/hold
@yevgeny-shnaidman please add an entry for the new file in `docs/mkdocs.yml`.

PR #438, opened 2023-06-12
Add 1.1 docs

qbarrand:
Add docs for:
- soft dependencies;
- in-tree unloading;
- variable substitution.

Rephrase part of the docs for ordered upgrade.

### DIFF
--- a/docs/mkdocs/documentation/deploy_kmod.md
+++ b/docs/mkdocs/documentation/deploy_kmod.md
@@ -50,6 +50,49 @@ The reconciliation loop for `Module` runs the following steps:
     2. successful build jobs;
     3. successful signing jobs.
 
+### Soft dependencies between kernel modules
+
+Some setups may require that several kernel modules be loaded in a specific order to work properly, although the modules
+do not directly depend on each other through symbols.
+`depmod` is usually not aware of those dependencies, and they do not appear in the files it produces.  
+If `mod_a` has a soft dependency on `mod_b`, `modprobe mod_a` will not load `mod_b`.
+
+Soft dependencies can be declared in the `Module` CRD via the
+`.spec.moduleLoader.container.modprobe.modulesLoadingOrder` field:
+
+```yaml
+modulesLoadingOrder:  # optional
+  - mod_a
+  - mod_b
+```
+
+With the configuration above:
+
+- the loading order will be `mod_b`, then `mod_a`;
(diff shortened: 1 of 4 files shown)


docs/mkdocs/documentation/deploy_kmod.md
```diff
@@ -50,6 +50,48 @@ The reconciliation loop for `Module` runs the following steps:
     2. successful build jobs;
     3. successful signing jobs.
 
+### Soft dependencies between kernel modules
+
+Some setups may require that several kernel modules be loaded to work properly, although the modules do not directly
```
yevgeny-shnaidman, line 55:
```suggestion
Some setups may require that several kernel modules be loaded in a specific order to work properly a, although the modules do not directly
```

docs/mkdocs/documentation/deploy_kmod.md
```diff
@@ -50,6 +50,48 @@ The reconciliation loop for `Module` runs the following steps:
...
+```yaml
+modulesLoadingOrder:  # optional
+  - mod_a
+  - mod_b
+```
+
+With the configuration above:
+
+- the loading order will be `mod_b`, then `mod_a`;
+- the unloading order will be `mod_a`, then `mod_b`.
```
yevgeny-shnaidman, line 72: we should probably add that the module to be loaded latest ( the one that depends on all the rest) should be also set in the moduleName field

docs/mkdocs/documentation/deploy_kmod.md
```diff
@@ -50,6 +50,48 @@ The reconciliation loop for `Module` runs the following steps:
...
+        moduleName: mod_a
+        # ...
+
+      # Other fields removed for brevity
+      inTreeModuleToRemove: mod_a
+```
+
+The ModuleLoader pod will first try to unload the in-tree `mod_a` before loading the one contained in the ModuleLoader
+image.  
+When the ModuleLoader pod is terminated and its `mod_a` is unloaded, the in-tree `mod_a` will not be loaded again.
```
yevgeny-shnaidman, line 93: I think we should specify that the module in the InTreeModuleToRemove field can be different from the one in the moduleName

yevgeny-shnaidman:
there is one more thing to mention (came up during Costi's testing): in case of ordered upgrade  namespace + name of the module should not exceed 39 characters

PR #445, opened 2023-06-15
Enforce name and namespace length in the webhook

### LINKED ISSUE #444: When `version` is set, make the webhook enforce the name + namespace length
When `version` is set, labels that include the `Module` name and namespace must be set on nodes.
[Label names must be 63 characters or less](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set), which means that the combined length of `Module` name and namespace may not exceed 39 characters.
The webhook should enforce this limit when the `version` field is set.

qbarrand:
When the `.spec.moduleLoader.container.version` field is set, the combined length of the name and namespace fields should not exceed 40 characters to avoid issues with labels.

Fixes #444 

### DIFF
--- a/api/v1beta1/module_webhook.go
+++ b/api/v1beta1/module_webhook.go
@@ -29,6 +29,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+// maxCombinedLength is the maximum combined length of Module name and namespace when the version field is set.
+const maxCombinedLength = 40
+
 // log is for logging in this package.
 var modulelog = logf.Log.WithName("module-resource")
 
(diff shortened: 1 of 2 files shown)


api/v1beta1/module_webhook.go
```diff
@@ -29,6 +29,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
+// maxCombinedLength is the maximum combined length of Module name and namespace when the version field is set.
+const maxCombinedLength = 40
```
yevgeny-shnaidman, line 33: maybe instead of setting combined length as 40, we will import the constants, parse them and determine the max  combined length that way? In case we will change the label in the future, we will be able to catch the errors in the unit-test.WDYT?

qbarrand: The `maxCombinedLength` unit test uses the 3 functions to determine if that constant is still up-to-date.

PR #454, opened 2023-06-21
Upgrade support between v1.02 and v1.1

yevgeny-shnaidman:
Currently in v1.1 the naming scheme for ModuleLoader and DevicePlugin daemonsets is different the what we had in v1.0.2. We are using hashings and the names are deterministic. This causes issues during upgrade, since the upgraded controller will try to create new daemonsets (new names). The old daemonsets can not be deleted, since it will cause the unloading of kernel modules. Since there is no way to change/delete lifecycle hooks of the running pod, we need to support the naming scheme of v1.0.2
The changes in PR:
1. get the runnings Daemonsets of the module in the earlier stage of the reconciliation loop
2. during ModuleLoader or DevicePlugin handling, check if the appropriate daemonset already exists based on the kernel-version and version label of daemonset
3. if now daemnonset is found, then create a new one using GenerateName option
4. removing the fix for pods being stuck in Terminating state for ~30 seconds, since it changes the Container spec of the Daemonset, which means that during upgrade the DaemonSet will delete the pod and then restart it, which may cause unloading of kernel module
5. unit test fixes

### DIFF
--- a/controllers/module_reconciler.go
+++ b/controllers/module_reconciler.go
@@ -24,6 +24,7 @@ import (
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/api"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/build"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/daemonset"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/metrics"
@@ -31,7 +32,6 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/sign"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/statusupdater"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
-	"github.com/mitchellh/hashstructure/v2"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
@@ -119,6 +119,11 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return res, fmt.Errorf("could get kernel mappings and nodes for modules %s: %w", mod.Name, err)
 	}
 
+	existingModuleDS, err := r.daemonAPI.GetModuleDaemonSets(ctx, mod.Name, mod.Namespace)
+	if err != nil {
+		return res, fmt.Errorf("could not get DaemonSets for module %s, namespace %s: %v", mod.Name, mod.Namespace, err)
+	}
+
 	for kernelVersion, mld := range mldMappings {
 		completedSuccessfully, err := r.reconHelperAPI.handleBuild(ctx, mld)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


config/manager/kustomization.yaml
```diff
@@ -6,8 +6,8 @@ resources:
 
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
+  newTag: upgrade-patch
```
qbarrand, line 10: Please revert those changes.

controllers/module_reconciler_test.go
```diff
@@ -696,10 +693,8 @@ var _ = Describe("ModuleReconciler_handleDevicePlugin", func() {
 			},
 		}
 
-		hashValue, err := hashstructure.Hash(hashData{ModuleVersion: ""}, hashstructure.FormatV2, nil)
-		Expect(err).NotTo(HaveOccurred())
-		name := fmt.Sprintf("%s-device-plugin-%x", mod.Name, hashValue)
-		existingDS := &appsv1.DaemonSet{
+		name := "some name"
```
qbarrand, line 696:
```suggestion
		const name = "some name"
```

controllers/module_reconciler_test.go
```diff
@@ -583,11 +581,13 @@ var _ = Describe("ModuleReconciler_handleDriverContainer", func() {
 			KernelVersion: "kernelVersion1",
 			ModuleVersion: "wr4656",
 		}
-		hashValue, err := hashstructure.Hash(hashData{KernelVersion: mld.KernelVersion, ModuleVersion: mld.ModuleVersion}, hashstructure.FormatV2, nil)
-		Expect(err).NotTo(HaveOccurred())
-		name := fmt.Sprintf("%s-%x", mld.Name, hashValue)
-		existingDS := &appsv1.DaemonSet{
-			ObjectMeta: metav1.ObjectMeta{Namespace: mld.Namespace, Name: name},
+		name := "some name"
```
qbarrand, line 584:
```suggestion
		const name = "some name"
```

config/manager/kustomization.yaml
yevgeny-shnaidman: Fixed

controllers/module_reconciler.go
```diff
@@ -119,6 +119,11 @@ func (r *ModuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return res, fmt.Errorf("could get kernel mappings and nodes for modules %s: %w", mod.Name, err)
 	}
 
+	existingModuleDS, err := r.daemonAPI.GetModuleDaemonSets(ctx, mod.Name, mod.Namespace)
+	if err != nil {
+		return res, fmt.Errorf("could get DaemonSets for module %s, namespace %s: %v", mod.Name, mod.Namespace, err)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/daemonset/daemonset.go
```diff
@@ -214,13 +213,12 @@ func (dc *daemonSetGenerator) SetDriverContainerAsDesired(
...
-				NodeSelector:          nodeSelector,
-				PriorityClassName:     "system-node-critical",
-				ServiceAccountName:    mld.ServiceAccountName,
-				Volumes:               volumes,
+				Containers:         []v1.Container{container},
+				ImagePullSecrets:   GetPodPullSecrets(mld.ImageRepoSecret),
+				NodeSelector:       nodeSelector,
+				PriorityClassName:  "system-node-critical",
+				ServiceAccountName: mld.ServiceAccountName,
+				Volumes:            volumes,
```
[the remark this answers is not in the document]
yevgeny-shnaidman: yes. As mentioned in the commit/PR message, ShareProcessNamespace was introduced in order to solve 30 second "Terminating state" for Module Loader pods. But it means that the update controller(v1.1) will change the existing ModuleLoader's Container spec, which will cause the removal and re-creation of a pod on a node. So, i prefer to skip it, since the 30 seconds issue is not a critical one

controllers/module_reconciler_test.go
yevgeny-shnaidman: fixed

internal/daemonset/daemonset.go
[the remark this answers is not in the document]
qbarrand: Sorry for missing that in the description.

PR #456, opened 2023-06-21
Quote arguments in modprobe commands

qbarrand:
Quote arguments to avoid shell injections.

/cc @yevgeny-shnaidman

### DIFF
--- a/internal/daemonset/daemonset.go
+++ b/internal/daemonset/daemonset.go
@@ -405,46 +405,39 @@ func makeLoadCommand(inTreeModuleToRemove string, spec kmmv1beta1.ModprobeSpec, 
 	var loadCommand strings.Builder
 
 	if inTreeModuleToRemove != "" {
-		loadCommand.WriteString("modprobe -r")
-		if spec.DirName != "" {
-			loadCommand.WriteString(" -d " + spec.DirName)
-		}
-		loadCommand.WriteString(" " + inTreeModuleToRemove + " && ")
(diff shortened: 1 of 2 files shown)


internal/daemonset/daemonset.go
```diff
@@ -405,46 +405,39 @@ func makeLoadCommand(inTreeModuleToRemove string, spec kmmv1beta1.ModprobeSpec,
...
-		if spec.DirName != "" {
-			loadCommand.WriteString(" -d " + spec.DirName)
-		}
-		loadCommand.WriteString(" " + inTreeModuleToRemove + " && ")
+		fmt.Fprintf(&loadCommand, "modprobe -r %q && ", inTreeModuleToRemove)
 	}
 
 	if fw := spec.FirmwarePath; fw != "" {
-		fmt.Fprintf(&loadCommand, "cp -r %s/* %s && ", fw, nodeVarLibFirmwarePath)
+		fmt.Fprintf(&loadCommand, `cp -r "%s/*" %s && `, fw, nodeVarLibFirmwarePath)
```
yevgeny-shnaidman, line 412: what about nodeVarLibFirmwarePath? Don't we want to quote it also?

yevgeny-shnaidman: you are right, then we don't need them in line 458

qbarrand: We do because we concatenate it with file names present in the container image's filesystem, and those are user-controlled strings.

PR #457, opened 2023-06-21
Add further checks on Module version

qbarrand:
Prevent updating to or from an empty ModuleLoader version, as this requires recreating new DaemonSets because of labelSelector updates.

/cc @yevgeny-shnaidman

### DIFF
--- a/api/v1beta1/module_webhook.go
+++ b/api/v1beta1/module_webhook.go
@@ -55,8 +55,20 @@ func (m *Module) ValidateCreate() (admission.Warnings, error) {
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (m *Module) ValidateUpdate(_ runtime.Object) (admission.Warnings, error) {
+func (m *Module) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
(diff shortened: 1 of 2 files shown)


api/v1beta1/module_webhook.go
```diff
@@ -55,9 +55,21 @@ func (m *Module) ValidateCreate() (admission.Warnings, error) {
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (m *Module) ValidateUpdate(_ runtime.Object) (admission.Warnings, error) {
+func (m *Module) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	modulelog.Info("Validating Module update", "name", m.Name, "namespace", m.Namespace)
 
+	if old != nil {
```
yevgeny-shnaidman, line 61: can old be nil in the ValidateUpdate? It will mean that the old object does not exists, and in that case ValidateCreate should have been called, no?

qbarrand: I agree that it should never be nil. However I have not dug into the implementation and I would not fail if `old` is nil but there is a legitimate reason that we have not thought about.

yevgeny-shnaidman: ok, fair enough.

PR #463, opened 2023-07-03
Add a proposal for worker Pods in KMM

qbarrand:
Add an enhancement proposal that addresses a number of pain points in the way KMM loads kernel modules in v1.

/cc @yevgeny-shnaidman @mresvanis @ybettan @hershpa

### DIFF
--- a/docs/enhancements/0001-worker-pods.md
+++ b/docs/enhancements/0001-worker-pods.md
@@ -1,0 +1,233 @@
+# Worker pods for KMM
+
+Authors: @qbarrand, @yevgeny-shnaidman, @mresvanis
+
+## Introduction
+
+This enhancement aims at eliminating several pain points related to how KMM loads kernel modules on nodes.
+
+### Reliability
+
+Running user-provided ModuleLoader images as DaemonSets makes it difficult for KMM to determine if a module was
+successfully unloaded or not.  
+KMM uses the `preStop` hook to run `modprobe -r` and unload modules.
+`preStop` hook failures cannot prevent the Pod from being terminated.  
+If `modprobe` cannot unload the module for any reason (e.g. userspace applications still using it), then the Pod will
+be terminated anyway and the `kmm.node.kubernetes.io/$namespace.$module.ready` label will be removed from the Node, even
+though the module is still loaded.
+
+### Operator upgrades
+
+The use of `postStart` and `preStop` hooks in ModuleLoader Pods means that any Pod restart or termination can unload the
+kernel module.
+This makes updating the DaemonSet spec impossible without interrupting the workload that depends on the kmod, because
+the DaemonSet controller will terminate the existing Pods and create new ones with the updated PodSpec.  
+Another issue is DaemonSet immutable fields; trying to update the `.spec.selector` will fail, which can make it
+challenging to improve the way KMM manages DaemonSets.
+
+### Resource utilization
+
+KMM currently creates one DaemonSet per `Module` and compatible kernel version found in the cluster.
+Formally, the number of DaemonSets created $S$ is defined by:
+$$S = \sum_{i=1}^nk_i$$
+
+where:
+- $n$ is the number of `Modules` in the cluster, and;
+- $k_i$ is the number of kernel versions running in the cluster that `Module` $i$ is compatible with.
+
+A big $S$ leads to unnecessary workload for the cluster:
+- on nodes, one ModuleLoader Pod is created per `Module`. Although those just sleep after the `modprobe`
+  `postStart` hook, each running pod incurs an overhead for the container runtime and the kubelet;
+- on the control-plane, a higher number of DaemonSets and Pods creates more events for kube-controller-manager, which
+  trigger more reconciliations and status updates.
+
+## Goals
+
+1. Improve the reliability of module loading and unloading
+2. Improve operator upgrade experience
+3. Reduce the overall KMM footprint
+
+## Non-goals
+
+1. Add further control on the deployment of kmods on nodes
+
+## Design
+
+### KMM: worker Pods pulling kmod images
+
+This proposal moves from ModuleLoader DaemonSets to standalone, one-shot Pods.
+KMM is now composed of two components:
+- the operator;
+- the worker Pods.
+
+```
+               ┌───────┐
+               │       │
+               │ Nodes ◄──┐
+               │       │  │
+               └───────┘  │ watches
+                          │
+             ┌─────────┐  │
+             │         │  │
+             │ Modules │  │     ┌───────────────────────┐
+             │         │  │     │                       │
+             └─────▲───┘  │     │ ┌───────────────────┐ │
+                   │      │     │ │                   │ │  creates & owns
+        watches &  └──────┴─────┼─┤ Module controller ├─┼──────────────────┐
+        updates status          │ │                   │ │                  │
+                                │ └───────────────────┘ │           ┌──────▼────────────┐
+                                │                       │           │                   │
+                                │                       │           │ NodeModulesConfig │
+┌────────────┐                  │ ┌───────────────────┐ │           │                   │
+│            │                  │ │                   │ │           └──────▲────────────┘
+│ KMM worker ◄──────────────────┼─┤ NodeModulesConfig │ │                  │
+│    Pod     │  creates & owns  │ │ controller        ├─┼──────────────────┘
+│            │                  │ │                   │ │   watches &
+└──────┬─────┘                  │ └───────────────────┘ │   updates status
+       │                        │                       │
+       │                        │          KMM operator │
+       │                        └───────────────────────┘
+       │
+       │
+       │                        ┌──────────────────────┐
+       │                        │                      │
+       └────────────────────────► registry.example.org │
+                pulls from      │                      │
+                                └──────────────────────┘
+```
+
+#### The operator
+
+The operator runs two main controllers: the `Module` controller and the `NodeModulesConfig` controller.
+
+The `Module` controller watches both `Module` and `Node` resources to determine which kmods should be loaded on which
+`Node`.
+It creates one `NodeModulesConfig` resource per `Node`.
+`NodeModulesConfig` resources represent the desired state of all `Modules` targting the node; they contain all the
+required data to pull kmod images and load the modules properly.  
+The `Module` controller watches updates to the `status` of `NodeModulesConfig` objects and aggregates that data into the
+`Module` `status`.
+
+The `NodeModulesConfig` controller watches `NodeModulesConfig` resources and determines when a worker Pod needs to be
+started on a node by looking at the difference between the desired state (spec) and the current one (status).  
+One worker Pod is created per `Node` and per `Module`.  
+When the Pod has completed, the controller reads its outcome and updates the `NodeModulesConfig` status accordingly.
+
+When updating the `NodeModulesConfig` status, the controller includes a timestamp corresponding to the last time a
+worker Pod ran on the node.
+This timestamp can be used to verify if a new worker Pod needs to run when the KMM operator restarts; if it is older
+than the last transition of the `Node`'s `Ready` condition, then the node may have rebooted and we potentially need to
+load the kernel modules.
+
+#### The worker Pods
+
+Worker Pods are created by the `NodeModulesConfig` controller.
+They run KMM code that reads the configuration file, pulls and extracts kmod OCI images and uses `modprobe` to load or
+unload kmods.  
+The configuration file is provided as an annotation to the Pod object and mounted in its filesystem via the [Downward
+API](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/).
+
+##### Security
+
+Worker Pods run a privileged container but are granted zero privilege on the API Server.
+
+##### kmod images
+
+It is worth noting that worker Pods pull the kmod image from registries **into their filesystem**, not the node's.
+That filesystem disappears as soon as the Pod is terminated.  
+This means that node image caching mechanisms cannot be leveraged, and any Pod creation will entail data transfers from
+the kmod image's registry to the Pod.
+
+This concern is mitigated by dropping the requirements on the kmod image.
+The worker image contains all the utilities required to load the kernel modules, and the kmod image is now merely a
+vehicle for `.ko` and `depmod` dependency files.  
+This results in much smaller images that can be based off `scratch` in most cases.
+
+##### Setting the firmware lookup path
+
+Worker Pods will be able to write the firmware lookup path to `/sys/module/firmware_class/parameters/path` each time
+they run.
+This behaviour will be controlled by an operator setting, which will be written as part of the worker configuration.
+
+#### High-level reconciliation loop
+
+```mermaid
+---
+High-level reconciliation loop
+---
+
+flowchart TD
+    RNMS[Reconcile NodeModulesConfig]
+    P[Worker Pod starts\nrunning on Node]
+    DMO{NodeModulesConfig\nspec matches status?}
+    TS{Last Node Ready\ntransition more recent than\nNodeModulesConfig status?}
+    E(((End)))
+
+    M{{Module update}} --> RNMS
+    N{{Node update}} --> RNMS
+
+    CR{{Controller starts}} --> TS
+    NR{{Node becomes Ready}} --> TS
+
+    RNMS --> DMO
+
+    DMO -- Yes --> E
+    DMO -- No --> P
+
+    TS -- Yes --> P
+    TS -- No --> E
+
+    P -- Pod completed or failed --> RO[Controller reads\nPod output] --> OUS[Controller updates\nNodeModulesConfig status] --> DMO
+```
+
+#### Device plugins
+
+KMM will still run the device plugin specified in the `Module` resource as a DaemonSet.
+The controller still adds the `kmm.node.kubernetes.io/$namespace.$module.ready` label on nodes that is relied upon to
+schedule the DaemonSet.
+
+#### Image builds and kmod signing
+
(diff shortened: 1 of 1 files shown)


docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,223 @@
...
+                                     │         │  │
+                                     │ Modules │  │     ┌───────────────────────┐
+                                     │         │  │     │                       │
+                                     └─────▲───┘  │     │ ┌───────────────────┐ │
+                                           │      │     │ │                   │ │    creates & owns
+                                watches &  └──────┴─────┼─┤ Module controller ├─┼─────────────────────┐
+                                updates status          │ │                   │ │                     │
+                                                        │ └───────────────────┘ │            ┌────────▼────────┐
+                                                        │                       │            │                 │
+                                                        │                       │            │ ModuleNodeState │
```
yevgeny-shnaidman, line 80: What about renaming ModuleNodeState --> NodeModulesState? the structure is per Node and it may contain multiple Modules

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,223 @@
...
+    
+    DMO -- Yes --> E
+    DMO -- No --> P
+    
+    CR[Controller starts] --> P
+    NR[Node becomes Ready] --> P
+    
+    P -- Pod completed or failed --> RO[Controller reads\nPod output] --> OUS[Controller updates\nNodeModuleState status] --> DMO
+```
+
```
yevgeny-shnaidman, line 179: We should also mention that Build/Sign flow stays as is

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,223 @@
...
+#### Device plugins
+
+KMM will still run the device plugin specified in the `Module` resource as a DaemonSet; however instead of relying on
```
yevgeny-shnaidman, line 182: why do we need a new label for Device Plugin? If we are removing dependency on module.ready label, then lets just use Module's Selector field and that's it. DevicePlugin should be able to handle the case where device file is not present

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,223 @@
...
+the `kmm.node.kubernetes.io/$namespace.$module.ready` label in its selector, we will introduce a new label
+that is independent of the kmod's state in the node.
+This should make it easier to terminate the device plugin before we try to unload the kmod.
```
yevgeny-shnaidman, line 185: How does it make it easier?

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,223 @@
...
+The new `ModuleNodeState` CRD will be strictly internal to KMM and should not be relied upon for any purpose.  
+Kernel version labeling for nodes will be dropped as it is not required to schedule ModuleLoader Pods anymore.
+
+## Upgrading from DaemonSet-based KMM
+
+There is no practical way of deleting the existing ModuleLoader DaemonSets without running their Pods' `preStop` hook.
+That hook runs `modprobe -r` and unloads the kernel module(s).  
+This means that transitioning from DaemonSets to the new system will entail an interruption of the workload utilizing
+the kernel module.  
+To minimize the downtime, we may add an additional controller that leverages the DaemonSet `OnDelete` strategy to delete
```
yevgeny-shnaidman, line 217: Another way is to ask customer to provide and executable that take a reference to the kernel module. We will deploy it as one-time DS,and it will allow us to remove ModuleLoader DS without worrying that we will unload the kernel module

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,223 @@
...
+The `ModuleNodeState` controller watches `ModuleNodeState` resources and determines when a worker Pod needs to be
+started on a node by looking at the difference between the desired state (spec) and the current one (status).  
+It prepares a worker configuration file in a `ConfigMap` based on the actions that need to be run on the node and
+creates a Pod mounting that `ConfigMap`.  
+When the Pod has completed, the controller reads the worker's output from the same `ConfigMap` and updates the
+`ModuleNodeState` status accordingly.
+
+When updating the `ModuleNodeState` status, the controller includes a timestamp corresponding to the last time a worker
+Pod ran on the node.
+This timestamp can be used to verify if a new worker Pod needs to run when the KMM operator restarts; if it older than
```
yevgeny-shnaidman, line 123: How are we going to update the status of the Module? via Module controller?

docs/enhancements/0001-worker-pods.md
qbarrand: In fact I think we could maybe increment the counter ourselves? There's `try_module_get`, although it seems [deprecated](https://stackoverflow.com/questions/1741415/linux-kernel-modules-when-to-use-try-module-get-module-put).

yevgeny-shnaidman: i looked into yet, but i could not find the common way that will suit all the kernel modules. I think it is OK to ask customer to provide such an executable, since he is familiar with the code and must know how to do it ( if it is needed at all)

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,229 @@
...
+`NodeModulesState` resources represent the desired state of all kmods that a certain node should have loaded; they
+contain all the required data to pull kmod images and load the modules properly.  
+The `Module` controller watches updates to the `status` of `NodeModulesState` objects and aggregates that data into the
+`Module` `status`.
+
+The `NodeModulesState` controller watches `NodeModulesState` resources and determines when a worker Pod needs to be
+started on a node by looking at the difference between the desired state (spec) and the current one (status).  
+It prepares a worker configuration file in a `ConfigMap` based on the actions that need to be run on the node and
+creates a Pod mounting that `ConfigMap`.  
+When the Pod has completed, the controller reads the worker's output from the same `ConfigMap` and updates the
```
yevgeny-shnaidman, line 120: If we are going in the direction of creating Pod per kmod/node, do we still want ConfigMap? Would not it be simpler to just pass all the data as arguments to in the Pod definition itself?

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,229 @@
...
+##### Security
+
+Worker Pods run a privileged container but are granted zero privilege on the API Server.
+
+##### kmod images
+
+It is worth noting that worker Pods pull images from registries **into their filesystem**, not the node's.
+That filesystem disappears as soon as the Pod is terminated.  
+This means that node image caching mechanisms cannot be leveraged, and any Pod creation will entail data transfers from
+the kmod image's registry to the Pod.
```
yevgeny-shnaidman, line 149: We should probably add that the only requirement from  kmod images is that they should include kmod/firmware/config file and that's it, so the base image can be any image, preferably as small as possible (like scratch)

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,229 @@
...
+The `NodeModulesState` controller watches `NodeModulesState` resources and determines when a worker Pod needs to be
+started on a node by looking at the difference between the desired state (spec) and the current one (status).  
+It prepares a worker configuration file in a `ConfigMap` based on the actions that need to be run on the node and
+creates a Pod mounting that `ConfigMap`.  
+When the Pod has completed, the controller reads the worker's output from the same `ConfigMap` and updates the
+`NodeModulesState` status accordingly.
+
+When updating the `NodeModulesState` status, the controller includes a timestamp corresponding to the last time a worker
+Pod ran on the node.
+This timestamp can be used to verify if a new worker Pod needs to run when the KMM operator restarts; if it older than
```
mresvanis, line 125:
```suggestion
This timestamp can be used to verify if a new worker Pod needs to run when the KMM operator restarts; if it is older than
```

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,229 @@
...
+It is worth noting that worker Pods pull images from registries **into their filesystem**, not the node's.
+That filesystem disappears as soon as the Pod is terminated.  
+This means that node image caching mechanisms cannot be leveraged, and any Pod creation will entail data transfers from
+the kmod image's registry to the Pod.
+
+##### Setting the firmware lookup path
+
+Worker Pods will be able to write the firmware lookup path to `/sys/module/firmware_class/parameters/path` each time
+they run.
+This behaviour will be controlled by an operator setting will be written as part of the worker configuration.
```
mresvanis, line 155: I'm not sure what you would like to convey here, but maybe:

```suggestion
This behaviour will be controlled by an operator setting, which will be written as part of the worker configuration.
```

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,229 @@
...
+the last transition of the `Node`'s `Ready` condition, then the node may have rebooted and we potentially need to load
+the kernel modules.
+
+#### The worker Pods
+
+Worker Pods are created by the `NodeModulesState` controller.
+They run KMM code that reads the configuration file, pulls and extracts kmod OCI images, uses `modprobe` to load kmods,
+and write some output to a file.
+Both the configuration and the output file are mounted from a `ConfigMap`.
+The `NodeModulesState` creates the `ConfigMaps` as well as the worker Pods.
```
ybettan, line 135: Do we consider `NodeModuleState` and the `ConfigMap` as user API or internal implementation?

ybettan: I see that you mentioned in the end of the docs that this is a private API and should not be used by users.

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,229 @@
...
+                                   │               ┌──────────────────────┐
+                                   │               │                      │
+                                   └───────────────► registry.example.org │
+                                      pulls from   │                      │
+                                                   └──────────────────────┘
+```
+
+#### The operator
+
+The operator runs two main controllers: the `Module` controller and the `NodeModulesState` controller.
```
ybettan, line 106: Why do we need this new controller? Why don't we go directly from `Module` to worker pods? I can understand the separation of concerns but isn't it just exposing to the cluster another object which is just an intermediate one?

docs/enhancements/0001-worker-pods.md
yevgeny-shnaidman: We are going to create a generic container image that will be the Image of worker pods. In case we are going with the option of Pod per Node per Module, and we are not going to output the result to ConfigMap, what's the point of creating ConfigMap per Pod just for input parameters sake, when we can just as easily code the parameters into the Pod itself?

docs/enhancements/0001-worker-pods.md
yevgeny-shnaidman: customer should only be aware of the Module

docs/enhancements/0001-worker-pods.md
yevgeny-shnaidman: NodeModulesState is basically a current state and config of the Module on the Node. We don''t want to store all that info in the Module status, since it is a lot of info that user does not need, and it will also means that any "bug fix"/development feature" will result in user API change.

docs/enhancements/0001-worker-pods.md
ybettan: I didn't know we are ditching the output file. In this case I also think we can remove the config map, since we have a different config for each pod (unlike app that load the static config once in its lifetime) passing arguments/ENV variables to the container seems more suitable to me.

docs/enhancements/0001-worker-pods.md
qbarrand: Added a few lines about that idea.

qbarrand:
@yevgeny-shnaidman @mresvanis @ybettan @hershpa PTAL

docs/enhancements/0001-worker-pods.md
yevgeny-shnaidman: so, when would we like to remove it? In ordered upgrade we have a dedicated label, and otherwise we have a selector: is it not enough?

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,233 @@
...
+                                └──────────────────────┘
+```
+
+#### The operator
+
+The operator runs two main controllers: the `Module` controller and the `NodeModuleState` controller.
+
+The `Module` controller watches both `Module` and `Node` resources to determine which kmods should be loaded on which
+`Node`.
+It creates one `NodeModuleState` resource per `Node` and `Module` that targets that node.
```
ybettan, line 105: I thought that we wanted to have a single `NodeModuleState` for the states of all modules on a nodes that has a least one module.

What is the benefit of having a `NodeModuleState` per module per node VS having a single `NodeModule*s*State` per node for all modules?

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,233 @@
...
+This timestamp can be used to verify if a new worker Pod needs to run when the KMM operator restarts; if it is older
+than the last transition of the `Node`'s `Ready` condition, then the node may have rebooted and we potentially need to
+load the kernel modules.
+
+#### The worker Pods
+
+Worker Pods are created by the `NodeModuleState` controller.
+They run KMM code that reads the configuration file, pulls and extracts kmod OCI images and uses `modprobe` to load or
+unload kmods.  
+The configuration file is provided as an annotation to the Pod object and mounted in its filesystem via the [Downward
```
ybettan, line 126: Why did we chose this approach instead of passing cmd args or env variables to the worker pod?
Seems like downard API is used in order to expose to the pod information about himself such as its name, namespace etc.

Here we are injecting a configuration file to the pod's metadata to then use the downward API. Isn't it sort of abusing it?

Furthermore, if we insist on a configuration file, then why we don't use the configmap for that? (even if we read the output from the pod instead of the configmap).

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,233 @@
...
+
+##### Security
+
+Worker Pods run a privileged container but are granted zero privilege on the API Server.
+
+##### kmod images
+
+It is worth noting that worker Pods pull the kmod image from registries **into their filesystem**, not the node's.
+That filesystem disappears as soon as the Pod is terminated.  
+This means that node image caching mechanisms cannot be leveraged, and any Pod creation will entail data transfers from
```
ybettan, line 137: Can't we mount the node's cache to the worker pods?

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,233 @@
...
+
+1. **Improve the reliability of module loading and unloading**  
+   KMM now runs its own code in the pod, which gives us a lot more control in error handling.
+
+2. **Improve operator upgrade experience**  
+   Worker pods are one-off, short-lived objects.
+   We never try to update them and are not constrained by spec changes and immutable selectors.
+
+3. **Reduce the overall KMM footprint**  
+   We now run only one Pod per Node, for only a short amount of time.
```
ybettan, line 205: We do create a `NodeModuleState` resource for each module thought. Isn't it increasing the footprint more than what it could be with a single `NodeModulesState`?

ybettan:
Do we want to move to `api/v2alpha1/` instead of `api/v1beta1` as part of the change?

qbarrand:
> Do we want to move to `api/v2alpha1/` instead of `api/v1beta1` as part of the change?

Why? we are not changing the API.

docs/enhancements/0001-worker-pods.md
qbarrand: Fair enough. Let us keep using the `kmm.node.kubernetes.io/$namespace.$module.ready` label for now.

docs/enhancements/0001-worker-pods.md
qbarrand: > Here we are injecting a configuration file to the pod's metadata to then use the downward API. Isn't it sort of abusing it?

I don't see how it constitutes an abuse. Using an annotation makes the config easily readable when getting the Pod with `kubectl`, as opposed to many environment variables or CLI flags.

docs/enhancements/0001-worker-pods.md
qbarrand: The nature and structure of the image cache varies on the container runtime used, and its storage driver.

docs/enhancements/0001-worker-pods.md
qbarrand: Proposal changed to one `NodeModulesState` per `Node`.

docs/enhancements/0001-worker-pods.md
```diff
@@ -0,0 +1,232 @@
...
+`NodeModules` resources represent the desired state of all `Modules` targting the node; they contain all the required
+data to pull kmod images and load the modules properly.  
+The `Module` controller watches updates to the `status` of `NodeModules` objects and aggregates that data into the
+`Module` `status`.
+
+The `NodeModules` controller watches `NodeModules` resources and determines when a worker Pod needs to be started on a
+node by looking at the difference between the desired state (spec) and the current one (status).
+When the Pod has completed, the controller reads its outcome and updates the `NodeModules` status accordingly.
+
+When updating the `NodeModules` status, the controller includes a timestamp corresponding to the last time a worker Pod
```
yevgeny-shnaidman, line 115: I think that we need to add (somewhere) the correlation between NodeModules and worker Pods, meaning: we will be creating one Pod per Node per Module, just to make it clear

yevgeny-shnaidman:
LGTM
@ybettan if you are happy , then please lgtm

ybettan:
@qbarrand Can we unhold?

PR #480, opened 2023-07-19
adding NodeModulesState type

yevgeny-shnaidman:


api/v1beta1/nodemodulestate_types.go
```diff
@@ -0,0 +1,87 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ModuleNodeSpec struct {
+	Name            string                   `json:"name"`
+	Namespace       string                   `json:"namespace"`
+	ModuleLoader    ModuleLoaderSpec         `json:"moduleLoader"`
+	ImageRepoSecret *v1.LocalObjectReference `json:"imageRepoSecret,omitempty"`
+	KmomLoaded      bool                     `json:"kmodLoaded"`
```
qbarrand, line 29:
```suggestion
	KmodLoaded      bool                     `json:"kmodLoaded"`
```

api/v1beta1/nodemodulestate_types.go
```diff
@@ -0,0 +1,87 @@
...
+// NodeModulesStateSpec describes the desired state of modules on the node
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type NodeModulesStateSpec struct {
+	// ModuleStateSpec list the spec of all the modules that need to be executed
+	// on the node
+	Modules []ModuleNodeSpec `json:"modules,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+type ModuleNodeStatus struct {
```
qbarrand, line 41: I think we need a `ModuleLoaderSpec` as well here, to determine if e.g. the desired parameters are different from those the module was loaded with.

api/v1beta1/nodemodulestate_types.go
qbarrand: Do we need a full `ModuleLoaderSpec`? In my PoC I had
```go
type ModuleConfig struct {

	//+optional
	InTreeModuleToRemove string `json:"inTreeModuleToRemove"`

	Modprobe ModprobeSpec `json:"modprobe"`
}
```

yevgeny-shnaidman: ok, i can change it, we will update later if needed

api/v1beta1/nodemodulestate_types.go
yevgeny-shnaidman: I am not sure if we want to support such a flow. Changing the parameters requires module unloading and reloading. It seems more like an upgrade flow

api/v1beta1/nodemodulestate_types.go
yevgeny-shnaidman: fixed

PR #481, opened 2023-07-19
adding ModuleConfig struct to NodeModulesSpec

yevgeny-shnaidman:
this struct will hold the desired configuration of the module on the node (image, to be loaded or not, modprobe configuration etc'), and will be passed as is to the worker Pod as an input


ybettan:
Are we starting to work on the implementation of v2 before the [proposal](https://github.com/kubernetes-sigs/kernel-module-management/pull/463) was merged?

api/v1beta1/nodemodulestate_types.go
```diff
@@ -20,16 +20,20 @@ import (
...
-	Namespace      string `json:"namespace"`
+type ModuleConfig struct {
 	ContainerImage string `json:"containerImage"`
 	//+optional
 	InTreeModuleToRemove string       `json:"inTreeModuleToRemove,omitempty"`
 	Modprobe             ModprobeSpec `json:"modprobe"`
 	KmodLoaded           bool         `json:"kmodLoaded"`
 }
 
+type ModuleNodeSpec struct {
```
ybettan, line 31:
```suggestion
type NodeModuleSpec struct {
```

api/v1beta1/nodemodulestate_types.go
```diff
@@ -20,16 +20,20 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-type ModuleNodeSpec struct {
-	Name           string `json:"name"`
-	Namespace      string `json:"namespace"`
+type ModuleConfig struct {
```
ybettan, line 23: Shouldn't this file be in `api/v2/...`?

yevgeny-shnaidman:
> Are we starting to work on the implementation of v2 before the [proposal](https://github.com/kubernetes-sigs/kernel-module-management/pull/463) was merged?
> 
> I still have some un-aswered questions on the config file passed to the worker pods.

@qbarrand lets answer all the questions officially in the Enhanement

api/v1beta1/nodemodulestate_types.go
yevgeny-shnaidman: i would prefer to leave it for now in v1. The official API to customer is not changing, so i am not sure if we want to move it to v2

api/v1beta1/nodemodulestate_types.go
yevgeny-shnaidman: i will change it

PR #491, opened 2023-07-26
Add the worker application

### LINKED ISSUE #470: Write a new `worker` application under `cmd`

qbarrand:
Implement basic features: pull kmod images, extract them on the filesystem and load / unload them.

Fixes #470


api/v1beta1/nodemodulesconfig_types.go
```diff
@@ -22,6 +22,8 @@ import (
 
 type ModuleConfig struct {
 	ContainerImage string `json:"containerImage"`
+	// When InsecurePull is true, the container image can be pulled without TLS.
```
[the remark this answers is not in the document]
qbarrand: That field is not needed as part of this PR. Let us add it when it is actually needed.

Dockerfile.worker
```diff
@@ -0,0 +1,40 @@
...
+RUN ["go", "mod", "download"]
+
+# Copy the go source
+COPY api api
+COPY cmd/worker cmd/worker
+COPY internal internal
+
+# Copy Makefile
+COPY Makefile Makefile
+COPY docs.mk docs.mk
```
ybettan, line 23: Why do we need to docs in the binary?

cmd/worker/main.go
```diff
@@ -0,0 +1,127 @@
...
+var kmodLoadCmd = &cobra.Command{
+	Use:   "load",
+	Short: "Load a kernel module",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cfgPath := args[0]
+
+		logger.V(1).Info("Reading config", "path", cfgPath)
+
+		cfg, err := configHelper.ReadConfigFile(cfgPath)
```
ybettan, line 44: Didn't we say the config file is a pod annotation?

Dockerfile.worker
qbarrand: `Makefile` includes `docs.mk`, if that file is not present then `make` fails.

cmd/worker/main.go
qbarrand: We did. That annotation is mounted as a file in the pod's filesystem. This is the code running in the worker pod, it reads that config file.

cmd/worker/main.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+var kmodUnloadCmd = &cobra.Command{
+	Use:   "unload",
+	Short: "Unload a kernel module",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cfgPath := args[0]
+
+		logger.V(1).Info("Reading config", "path", cfgPath)
```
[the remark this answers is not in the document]
qbarrand: It is not exactly the same in all 3 commands, because `kmod reload` needs to read for the old and the new config; what we log there is slightly different. @yevgeny-shnaidman WDYT?

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+	if inTree := cfg.InTreeModuleToRemove; inTree != "" {
+		w.logger.Info("Unloading in-tree module", "name", inTree)
+
+		if err = w.mr.Run(ctx, "-rv", inTree); err != nil {
+			return fmt.Errorf("could not remove in-tree module %s: %v", inTree, err)
+		}
+	}
+
+	// TODO copy firmware
+	// TODO handle ModulesLoadingOrder
```
ybettan, line 58: Is this planned for future PRs?

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+
+		args = append(args, moduleName)
+		args = append(args, cfg.Modprobe.Parameters...)
+	}
+
+	return w.mr.Run(ctx, args...)
+}
+
+func (w *worker) ReloadKmod(ctx context.Context, oldCfg, newCfg *kmmv1beta1.ModuleConfig) error {
+	return errors.New("not implemented")
```
ybettan, line 81: Should we add a TODO here?

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+
+func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig) error {
+	imageName := cfg.ContainerImage
+
+	w.logger.Info("Pulling image", "name", imageName)
+
+	pr, err := w.ip.PullToDisk(ctx, imageName, cfg.InsecurePull)
+	if err != nil {
+		return fmt.Errorf("could not pull %q: %v", imageName, err)
+	}
```
ybettan, line 92: Why do we need to pull the image for unloading the kmod? We only need the kmod's name which should be in the config isn't it?

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+func NewImagePuller(baseDir string, logger logr.Logger) ImagePuller {
+	return &imagePuller{
+		baseDir: baseDir,
+		logger:  logger,
+	}
+}
+
+func (i *imagePuller) PullToDisk(ctx context.Context, imageName string, insecurePull bool) (pullResult, error) {
+	logger := i.logger.V(1)
+	logger.Info("Pulling image")
```
ybettan, line 145:
```suggestion
	logger.Info("Pulling image", imageName)
```

internal/worker/worker.go
qbarrand: Yes, see https://github.com/kubernetes-sigs/kernel-module-management/issues/469

internal/worker/worker.go
qbarrand: If we do not pull the image, `modprobe` has no idea about what `kmm-kmod` is. It needs to parse depmod-generated dependency files to understand to what `.ko` files `kmm-kmod` corresponds.

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func NewImagePuller(baseDir string, logger logr.Logger) ImagePuller {
+	return &imagePuller{
+		baseDir: baseDir,
+		logger:  logger,
+	}
+}
+
+func (i *imagePuller) PullToDisk(ctx context.Context, imageName string, insecurePull bool) (pullResult, error) {
```
yevgeny-shnaidman, line 143: General remark: how about we stash the digest implementation for now? We can implement a simple pull onto the container FS. Once we have a working scenario we can come back to it and see if it worth to mount host path and implement the cache with digest. WDYT?

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+
+	if inTree := cfg.InTreeModuleToRemove; inTree != "" {
+		w.logger.Info("Unloading in-tree module", "name", inTree)
+
+		if err = w.mr.Run(ctx, "-rv", inTree); err != nil {
+			return fmt.Errorf("could not remove in-tree module %s: %v", inTree, err)
+		}
+	}
+
+	// TODO copy firmware
```
yevgeny-shnaidman, line 57: This is something that we need to discuss a little bit more, especially with the customers/partners. I think that at least some of them would like to separate the firmware image from the kmod image, so that they can manage the upgrades better

cmd/worker/main.go
```diff
@@ -0,0 +1,127 @@
...
+	Use:   "kmod",
+	Short: "Manage kernel modules",
+}
+
+var kmodLoadCmd = &cobra.Command{
+	Use:   "load",
+	Short: "Load a kernel module",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cfgPath := args[0]
```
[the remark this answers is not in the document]
qbarrand: Yes to all questions.

internal/worker/cmdlog.go
```diff
@@ -0,0 +1,69 @@
...
+		logger: logger,
+		stdErr: stderr,
+		stdOut: stdout,
+		wg:     &sync.WaitGroup{},
+	}
+
+	return &cl, nil
+}
+
+func (cl *CommandLogger) Wait() error {
```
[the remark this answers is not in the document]
qbarrand: In fact it does wait until the underlying program closes its stderr and stdout.

internal/worker/worker.go
qbarrand: I think we can implement that in a subsequent version. Right now I would focus on feature parity with v1.

internal/worker/worker.go
```diff
@@ -0,0 +1,300 @@
...
+		defer rd.Close()
+
+		if err := extractTarToDisk(rd, dstDirFS); err != nil {
+			errs <- err
+		}
+
+		logger.V(1).Info("Done writing tar archive")
+	}()
+
+	go func() {
```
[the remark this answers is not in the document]
qbarrand: The context will be cancelled by any SIGTERM - see `cmd/worker/main.go`. We do not want to wait for image extraction if we receive a SIGTERM.

internal/worker/worker.go
qbarrand: The digest implementation saves an image pull and extraction when we need to reload a kmod from the same kmod image. I agree that it's not so useful now, but since it's already there, would you mind keeping it?

internal/worker/worker.go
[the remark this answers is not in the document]
yevgeny-shnaidman: but then the pod will fail, so what do we care?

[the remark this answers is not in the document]
qbarrand: We want it to fail immediately and not wait for the extraction to complete. The context gives us that.

internal/worker/worker.go
yevgeny-shnaidman: it is only useful in case we start mounting the host directory into the Pod from the beginning?  Are we going to do it from the beginning? It als means implementing garbage collection on the host. I think this is something that we need to implement (if at all) only at later stage, as an optimization.  If you want to keep the code, then i have no objection, lets just add a comment do describe the digest mechanism and add a TODO to remove it later if needed

qbarrand: It is useful without mounting the host directory in the pod, if you use the `kmod reload` feature where you might be pulling the same image twice.

internal/worker/cmdlog.go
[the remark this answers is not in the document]
yevgeny-shnaidman: you are right, i missed the the read from err channel. But why do we need the wg.Wait and close(err) in a goroutine?

cmd/worker/main.go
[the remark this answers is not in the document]
yevgeny-shnaidman: I think we should leave as is - the more explicit the logs, the easier the debugging :)

internal/worker/worker.go
yevgeny-shnaidman: but in that case isn't it easier just to check the imageName in the reloadKmod command? It is safer and quicker than trying to write 2 files instead of one (digest). Also saves a failing the pod when the only thing that failed is figest write, while the image extraction was succesfull

qbarrand: Now removing `dstDir` (containing `fs` and `digest`) if `digest` does not exist or does not contain the expected value.
Keeping `digest` allows simple retries by calling `LoadKmod` or `UnloadKmod` in a loop. Because `digest` is written only if the extraction was successful, we now on the next try that we have an extracted filesystem we can safely use and skip the pull.

internal/worker/worker.go
qbarrand: Added a TODO. But in fact, should we even implement this?
If the pod succeeds in unloading the old kmod, but fails to load the new one, we will not update the `NodeModulesConfig` to reflect that the old kmod is gone.
@yevgeny-shnaidman if we split reloads into two separate tasks and worker pods, we eliminate that problem.

cmd/worker/main.go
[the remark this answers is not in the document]
qbarrand: Let's discuss reload in [that thread](https://github.com/kubernetes-sigs/kernel-module-management/pull/491#discussion_r1277727296).

internal/worker/cmdlog.go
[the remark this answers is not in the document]
qbarrand: We need someone to close the channel, otherwise `return <-errs` will hang forever.

internal/worker/worker.go
yevgeny-shnaidman: agree. we can remove this command and move this decision into the NMC controller

internal/worker/cmdlog.go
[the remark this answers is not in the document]
yevgeny-shnaidman: i mean, why do we need them in goroutine, and not as a usual flow in the function?

internal/worker/worker.go
yevgeny-shnaidman: since reload command is not relevant anymore, do we still want this optimization in the container FS?

qbarrand: I think we do, because it makes retries much easier.

yevgeny-shnaidman: only in case we mount a host directory, but that is not what we are doing right now, no?

PR #495, opened 2023-07-30
Introducing Module-to-NMC reconciliation logic

yevgeny-shnaidman:
NMC package will update/create module's NMC entry
based on the mld configuration. Currently it is a standalone package with unit test. Later on it will be called by the module controller instead of HandleDriverContainer.
GarbageCollecto scenario is not handled by this PR: in case module does not contain mapping for the node's kernel version, but the NMC of the node does contain module version.

Decision flow:
1) based on node's kernel version, taints and labels, the module is
   either flag as should be running on the node or should not be running
   on the node
2) in case it should be running: the module config data of the module
   for that node's NMC is updated/created. In case NMC does not exists -
   it is creates first
3) in case it should not be running: if NMC does not exists, nothing to
   do,otherwise remove modules data from NMC spec


api/v1beta1/nodemodulesconfig_types.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 type ModuleConfig struct {
+	KernelVersion  string `json:"kernelVersion"`
```
ybettan, line 24: Why do we need a `KernelVersion` field? Can't we assume that all modules in the `NodeMoudulesConfigSpec` should run on the node's kernel?

yevgeny-shnaidman: in order to deal with race conditions between updating the MNC data of the module, and node's rebooting with new kernel version. Lets say node is rebooted with new kernel version, but module-nmc controller has not updated the NMC entree yet: then NMC controller will try to boot the inappropriate version of the kernel module. Haveing kernel version will allow NMC controller to verify that the module data in the NMC is relevant for that specific node's  kernel version

internal/filter/filter.go
```diff
@@ -144,28 +142,18 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
-
-			sel = sel.Add(*requirement)
+		res, err := utils.IsObjectSelectedByLabels(node.GetLabels(), mod.Spec.Selector)
+		if err != nil {
+			logger.Error(err, "could not determine if node is selected by module", "node", node.GetName(), "module", mod.Name)
+			return reqs
 		}
 
-		if !sel.Matches(nodeLabelsSet) {
+		if !res {
```
ybettan, line 156:
```suggestion
		if !moduleSelectorMatchNode {
```

internal/filter/filter.go
```diff
@@ -197,28 +185,18 @@ func (f *Filter) FindManagedClusterModulesForCluster(ctx context.Context, cluste
...
-
-			sel = sel.Add(*requirement)
+		res, err := utils.IsObjectSelectedByLabels(cluster.GetLabels(), mod.Spec.Selector)
+		if err != nil {
+			logger.Error(err, "could not determine if cluster is selected by ManagedClusterModule", "node", cluster.GetName(), "module", mod.Name)
+			return reqs
 		}
 
-		if !sel.Matches(clusterLabelsSet) {
+		if !res {
```
ybettan, line 199:
```suggestion
		if !moduleSelectorMatchCluster {
```

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;watch;patch
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=nodemodulesconfigs,verbs=get;list;watch;update;patch;create
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=nodemodulesconfigs/status,verbs=get;update;patch
+
+type ModuleNMCHandler interface {
```
ybettan, line 22: Why is it called `ModuleNMCHandler` and not `NodeModuleConfigHandler` for example?

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+// func (mnr *ModuleNMCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (mnh *moduleNMCHandler) handleModuleNMC(ctx context.Context, mld *api.ModuleLoaderData, nodes []v1.Node) error {
+	var sumErr error
+	for _, node := range nodes {
+		nodeKernelVersion := strings.TrimSuffix(node.Status.NodeInfo.KernelVersion, "+")
+
+		shouldBeOnNode, err := mnh.reconHelper.shouldModuleRunOnNode(node, mld, nodeKernelVersion)
```
ybettan, line 44:
```suggestion
		shouldBeOnNode, err := mnh.reconHelper.shouldModuleRunOnNode(node, mld)
```

We already have this info in `node`

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,139 @@
...
+			err = mnh.reconHelper.disableModuleOnNode(ctx, mld.Namespace, mld.Name, node.Name)
+		}
+		if err != nil {
+			sumErr = fmt.Errorf("%v: %v", sumErr, err)
+		}
+	}
+
+	if sumErr != nil {
+		return fmt.Errorf("failed to handle module %s/%s with nodes: %v", mld.Namespace, mld.Name, sumErr)
+	}
```
ybettan, line 61: Should we use [multierror](https://pkg.go.dev/github.com/hashicorp/go-multierror) instead?

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, mnhh.client, nmc, func() error {
+		return mnhh.nmcHelper.SetNMCAsDesired(ctx, nmc, mld.Namespace, mld.Name, &moduleConfig)
+	})
+
+	if err == nil {
+		logger.Info("enable module in NMC", "name", mld.Name, "namespace", mld.Namespace, "node", nodeName, "result", opRes)
+	}
+	return err
```
ybettan, line 117:
```suggestion
	if err != nil {
	        return fmt.Errorf("failed to create ..., %v, err)
	}
	
	logger.Info("enable module in NMC", "name", mld.Name, "namespace", mld.Namespace, "node", nodeName, "result", opRes)
	return nil
```

How about this instead? I think it is a bit more clear

api/v1beta1/nodemodulesconfig_types.go
ybettan: How about a `LastUpdated` timestamp instead? we can then compare it to `node.metadata.creationTimestamp`.
I think having the kernel version is confusing if we just try to solve a race condition. WDYT?

internal/filter/filter.go
yevgeny-shnaidman: fixed

internal/nmc/module_nmc_controller.go
yevgeny-shnaidman: this controller updates NMC based on the Module. We will have a different controller that try to reconcile node based on NMC, and it will be call NMC controller

internal/nmc/module_nmc_controller.go
yevgeny-shnaidman: we do, but we have already extracted that data above, why not just use it, instead of extracting it again?

internal/nmc/module_nmc_controller.go
yevgeny-shnaidman: fixed

internal/nmc/module_nmc_controller.go
yevgeny-shnaidman: since we are moving to Golang 1.20, changed it to errors package

api/v1beta1/nodemodulesconfig_types.go
yevgeny-shnaidman: why is it confusing? and how would LastUpdated help in that case ( we would add it, but for another issue)? To differentiate between node that just rebooted , to node that rebooted with new kernel, only KernelVersion will help

internal/nmc/module_nmc_controller.go
qbarrand, line 1: Will you add the actual reconciler in a subsequent PR?

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,142 @@
...
+			sumErr = errors.Join(sumErr, err)
+			continue
+		}
+		if shouldBeOnNode {
+			err = mnh.reconHelper.enableModuleOnNode(ctx, mld, node.Name, nodeKernelVersion)
+		} else {
+			err = mnh.reconHelper.disableModuleOnNode(ctx, mld.Namespace, mld.Name, node.Name)
+		}
+
+		sumErr = errors.Join(sumErr, err)
```
qbarrand, line 56: This is not the recommended pattern: https://go-review.googlesource.com/c/go/+/495076
Instead, let us append to a `[]error` and use `errors.Join(errs...)` at the end of the loop.

qbarrand:
/hold
For this not to be merged while I am reviewing 🙂

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get the NodeModulesConfig for node %s: %v", nodeName, err)
+	}
+	if nmc == nil {
+		// NodeModulesConfig does not exists, module was never running on the node, we are good
+		return nil
+	}
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, mnhh.client, nmc, func() error {
+		return mnhh.nmcHelper.SetNMCAsDesired(ctx, nmc, modNamespace, modName, nil)
```
qbarrand, line 133: Reading this code again, I think it would be more straightforward with a dedicated function:
```go
func RemoveModuleConfig(nmsList *[]NodeModuleSpec*, nmsNamespace, nmsName string)
```
This would very much look like the [conditions management code](https://github.com/kubernetes/apimachinery/blob/5cb236977966cfc829e3b94b57fa69b44e0a95bc/pkg/api/meta/conditions.go#L57-L71).
WDYT?

internal/utils/nodeutils.go
```diff
@@ -12,3 +16,18 @@ func IsNodeSchedulable(node *v1.Node) bool {
 	}
 	return true
 }
+
+func IsObjectSelectedByLabels(objectLabels map[string]string, selectorLabels map[string]string) (bool, error) {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: currently there are not other options, unless we create a new file. Since anyway everything is under utils package, maybe we can rename the file to something more general?

internal/nmc/module_nmc_controller.go
qbarrand: I think that would be more straightforward indeed, but unless I am missing something this PR does not do that.

api/v1beta1/nodemodulesconfig_types.go
ybettan: > why is it confusing? and how would LastUpdated help in that case ( we would add it, but for another issue)? To differentiate between node that just rebooted , to node that rebooted with new kernel, only KernelVersion will help

It was confusing to me because I was expecting all modules on a node to have the same kernel version of the node of for dealing with race condition I am used to check timestamps and I didn't understand at first why we need the kernel version.

Since we don't have a way to check timestams anyway then using the kernel version is a good option indeed.

internal/nmc/module_nmc_controller.go
ybettan: Because we are sending the same data twice which is just redundant IMO and may lead to inconsistency in future code changes. What is the cost of extracting it again? This is just a golang struct. It is not like we are re-generating a new API call to k8s.

internal/nmc/module_nmc_controller.go
```diff
@@ -0,0 +1,142 @@
+package nmc
```
ybettan, line 1: Can we rename this file to `internal/nbc/module_nmc.go`? This is just a library and not a controller.

yevgeny-shnaidman: fixed

internal/nmc/module_nmc_controller.go
yevgeny-shnaidman: fixed

internal/nmc/module_nmc_controller.go
yevgeny-shnaidman: Fixed

PR #503, opened 2023-08-03
Introducing Module-NMC controller

yevgeny-shnaidman:
This PR introduces a new controller, that is responsible for reconciliation of NMC based on the data in Module and Node The following are the main points:
1) It watches for changes in Modules, Nodes (labels, becoming
   schedulable) and Jobs belonging owned by Modules (in order to
   reschedule once Build/Sign job are finished)
2) Per reconciled Module, go over all the nodes in the cluster
   and reconcile their NMCs based on the Module/Node state

Changes in the PR:
1) adding functions to filter package that are resposible for filtering
    the events (nodes. modules, jobs) and finding out which modules
    should be reconciled
2)  changing the NMC handler logic a little, adding verification for
    container image presence in case module should be enable.
3)  Module-NMC Reconciliation is being done both in case Kernel Mapping
    exists and in case it does not exists ( solves the garbage
    collection flow)

This PR does not include Module status updates based on NMC updates


cmd/manager/main.go
```diff
@@ -146,11 +149,20 @@ func main() {
...
 	)
-
 	if err = mc.SetupWithManager(mgr, constants.KernelLabel); err != nil {
 		cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.ModuleReconcilerName)
 	}
 
+	mnc := controllers.NewModuleNMCReconciler(
+		client,
+		kernelAPI,
+		nmcHandler,
```
ybettan, line 159: Should this one be renamed to `nmcAPI` to be consistent with the rest of the API passed to it (or to the Module reconciler)?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;watch;patch
```
ybettan, line 25: Why do we need to `patch` nodes?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=nodemodulesconfigs,verbs=get;list;watch;update;patch;create
```
ybettan, line 26: What is the difference between `update` and `patch`?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+		reconHelper: reconHelper,
+		nmcHandler:  nmcHandler,
+	}
+}
+
+func (mnr *ModuleNMCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// get the reconciled module
+	logger := log.FromContext(ctx)
+
+	logger.Info("Starting Module-NMS reconcilation", "module name and namespace", req.NamespacedName)
```
ybettan, line 56: Who says that the reconciled object is a `Module`? Can't it be a `Node`? a `Job`?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+
+	mod, err := mnr.reconHelper.getRequestedModule(ctx, req.NamespacedName)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("Module deleted")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get the requested %s KMMO CR: %w", req.NamespacedName, err)
```
ybettan, line 64:
```suggestion
		return ctrl.Result{}, fmt.Errorf("failed to get the requested %s KMMO CR: %v", req.NamespacedName, err)
```

Are we using the wrapped error anywhere?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+	// get all nodes
+	nodes, err := mnr.reconHelper.getNodesList(ctx)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to get nodes list: %v", err)
+	}
+
+	errs := make([]error, 0, len(nodes))
+	for _, node := range nodes {
+		kernelVersion := strings.TrimSuffix(node.Status.NodeInfo.KernelVersion, "+")
+		mld, err := mnr.kernelAPI.GetModuleLoaderDataForKernel(mod, kernelVersion)
```
ybettan, line 76: Why do we need to do that for every node? Can we extract it outside of the loop?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+		if err != nil && !errors.Is(err, module.ErrNoMatchingKernelMapping) {
+			logger.Info(utils.WarnString(fmt.Sprintf("internal errors while fetching kernel mapping for version %s: %v", kernelVersion, err)))
+			errs = append(errs, err)
+			continue
+		}
+
+		err = mnr.nmcHandler.HandleModuleNMC(ctx, mod, mld, node)
+		errs = append(errs, err)
```
ybettan, line 84: Aren't you appending `nil` errors here as well?

cmd/manager/main.go
yevgeny-shnaidman: This is a controller, we are not creating/returning an Interface ( API), but the struct itself, as the rest of the reconcilers/controllers, that's why there is no API in the name

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to reconcile module %s/%s with nodes: %v", mod.Namespace, mod.Name, err)
+	}
+	return ctrl.Result{}, nil
+}
+
+//go:generate mockgen -source=module_nmc_reconciler.go -package=controllers -destination=mock_module_nmc_reconciler.go moduleNMCReconcilerHelperAPI
+
+type moduleNMCReconcilerHelperAPI interface {
+	getRequestedModule(ctx context.Context, namespacedName types.NamespacedName) (*kmmv1beta1.Module, error)
+	getNodesList(ctx context.Context) ([]v1.Node, error)
```
ybettan, line 98: I am wondering why `getRequestedModule` isn't a method in the `module` package and why we don't have a `node` package that will hold `getNodeList` as well.

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: that depends on the action that you execute on the object. I prefer to mention both of them, so that we can execute them in the code without a need to update the rules

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+	client client.Client
+}
+
+func newModuleNMCReconcilerHelper(client client.Client) moduleNMCReconcilerHelperAPI {
+	return &moduleNMCReconcilerHelper{
+		client: client,
+	}
+}
+
+func (mnrh *moduleNMCReconcilerHelper) getRequestedModule(ctx context.Context, namespacedName types.NamespacedName) (*kmmv1beta1.Module, error) {
```
ybettan, line 111: Isn't this function a dup of `getRequestedModule` in the `module_reconciler.go` file?

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,150 @@
...
+	mod := kmmv1beta1.Module{}
+
+	if err := mnrh.client.Get(ctx, namespacedName, &mod); err != nil {
+		return nil, fmt.Errorf("failed to get the kmmo module %s: %w", namespacedName, err)
+	}
+	return &mod, nil
+}
+
+func (mnrh *moduleNMCReconcilerHelper) getNodesList(ctx context.Context) ([]v1.Node, error) {
```
ybettan, line 120: Same here, this function exist (almost) in `module_reconciler.go`

cmd/manager/main.go
ybettan: `nmcHandler` is a public interface. Why don't we call it `nmcAPI` instead?

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: because this is how this reconciler is defined. It is defined For Module, and for every "watch" it finds the appropriate Module to reconcile

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: Fixed. We do use wrapped errors , when needed. For example in the KernelMapper, to get the appropriate MLD, if exists

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: because the kernel version of the nodes may differ

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: yes, but errors.Join knows how to handle them and ignore

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: As discussed in Slack, i am leaving it here and we can decide on it and the rest of functions in packages in v2.1

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: As discussed in Slack, i am leaving it here and we can decide on it and the rest of functions in packages in v2.1

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: As discussed in Slack, i am leaving it here and we can decide on it and the rest of functions in packages in v2.1

cmd/manager/main.go
yevgeny-shnaidman: moved the NMC-Handler API into the controller, since it implements controller logic

controllers/module_nmc_reconciler.go
ybettan: I meant, "do we use the wrapped error in this case". If you fixed it I assume we were not using it :)

controllers/module_nmc_reconciler.go
ybettan: But I don't see `node` being passed to this function or the implementation depending on the nodes in any way. Am I missing something?

controllers/module_nmc_reconciler.go
ybettan: Won't it be clearer to only append real errors instead of delegating it to `Join`?

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: the node is used to determine the kernelVersion , and to determine kernel mapping (using kernel version). Then both kernel mapping(mld) and node are passed to the function that determines whether the module needs to run on the node or not

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: why to add another "if", if there is already code in the Join that takes care of it? This is the purpose of Join function: decide of there are errors in the array of errors , and create the appropriate error

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: yes, in this case it was a mistake

controllers/module_nmc_reconciler.go
ybettan: > why to add another "if", if there is already code in the Join that takes care of it?

Because it is not trivial for someone who didn't read the `Join` doc to understand those are ignored. Everywhere else in the code we are checking if an error is != `nil` before doing something with it except here.

If you think it is better this way, let's keep it, I just though it is a bit confusing.

internal/filter/filter.go
```diff
@@ -169,6 +211,67 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
+		reqs = append(reqs, reconcile.Request{NamespacedName: nsn})
+		reqSet.Insert(nsn)
+	}
+
+	nms, err := f.nmcHelper.Get(ctx, node.GetName())
+	if err != nil || nms == nil {
+		return reqs
+	}
+
+	// go over modules of NodeModulesConfig and add them to request if they are not there already
```
qbarrand, line 260: Why should we do this? We have enqueued all Modules that we need to reconcile already.
Besides, this function should return NamespacedNames for Modules only, not Nodes or NMCs.

internal/filter/filter.go
```diff
@@ -241,6 +344,16 @@ func (f *Filter) EnqueueAllPreflightValidations(ctx context.Context, mod client.
...
 
+func (f *Filter) FindModuleForJob(ctx context.Context, job client.Object) []reconcile.Request {
+	owner := metav1.GetControllerOfNoCopy(job)
+
+	if owner != nil && owner.Kind == "Module" {
+		req := reconcile.Request{NamespacedName: types.NamespacedName{Name: owner.Name, Namespace: job.GetNamespace()}}
+		return []reconcile.Request{req}
+	}
+	return nil
+}
```
qbarrand, line 355: I do not think we need this function. Using `.Owns(&batchv1.Job{})` in `SetupWithManager` does this already.

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: i am keeping it: if we have a Golang library function in the code, i would expect that developer to read about it and understand

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: I prefer to set such function in a helper interface, makes it easier to write the unit-tests for the main reconciliation loop, and to check all the corner case in a dedicated unit-test for that specific function.

cmd/manager/main.go
```diff
@@ -134,6 +136,7 @@ func main() {
 
 	daemonAPI := daemonset.NewCreator(client, constants.KernelLabel, scheme)
 	kernelAPI := module.NewKernelMapper(buildHelperAPI, sign.NewSignerHelper())
+	//nmcHandler := nmc.NewModuleNMCHandler(client, registryAPI, nmcHelper)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,241 @@
...
+
+	logger.Info("Starting Module-NMS reconcilation", "module name and namespace", req.NamespacedName)
+
+	mod, err := mnr.reconHelper.getRequestedModule(ctx, req.NamespacedName)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			logger.Info("Module deleted")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, fmt.Errorf("failed to get the requested %s KMMO CR: %v", req.NamespacedName, err)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,241 @@
...
+		registryAPI: registryAPI,
+		nmcHelper:   nmcHelper,
+	}
+}
+
+func (mnrh *moduleNMCReconcilerHelper) getRequestedModule(ctx context.Context, namespacedName types.NamespacedName) (*kmmv1beta1.Module, error) {
+	mod := kmmv1beta1.Module{}
+
+	if err := mnrh.client.Get(ctx, namespacedName, &mod); err != nil {
+		return nil, fmt.Errorf("failed to get the kmmo module %s: %w", namespacedName, err)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, mnrh.client, nmc, func() error {
+		return mnrh.nmcHelper.SetModuleConfig(ctx, nmc, mld.Namespace, mld.Name, &moduleConfig)
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to enable module %s/%s in NMC %s: %v", mld.Namespace, mld.Name, nodeName, err)
+	}
+	logger.Info("enable module in NMC", "name", mld.Name, "namespace", mld.Namespace, "node", nodeName, "result", opRes)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,241 @@
...
+
+	opRes, err := controllerutil.CreateOrPatch(ctx, mnrh.client, nmc, func() error {
+		return mnrh.nmcHelper.RemoveModuleConfig(ctx, nmc, modNamespace, modName)
+	})
+
+	if err != nil {
+		return fmt.Errorf("failed to diable module %s/%s in NMC %s: %v", modNamespace, modName, nodeName, err)
+	}
+
+	logger.Info("disable module in NMC", "name", modName, "namespace", modNamespace, "node", nodeName, "result", opRes)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/filter/filter.go
```diff
@@ -100,6 +127,21 @@ func (f *Filter) ModuleReconcilerNodePredicate(kernelLabel string) predicate.Pre
...
+		predicate.Or(nodeBecomesSchedulable, predicate.LabelChangedPredicate{}),
+	)
+}
+
+func ModuleNMCReconcileJobPredicate() predicate.Predicate {
+	return predicate.And(
+		skipDeletions,
+		skipCreations,
+		moduleJobSuccess,
+	)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Why do we need to process failed job? We are only interested in kicking in the reconciliation in case there is a chance that an image was built/created

internal/filter/filter.go
```diff
@@ -169,6 +211,67 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
+func (f *Filter) FindModulesForNMCNodeChange(ctx context.Context, node client.Object) []reconcile.Request {
+	logger := ctrl.LoggerFrom(ctx).WithValues("node", node.GetName())
+
+	reqs := make([]reconcile.Request, 0)
+
+	logger.Info("Listing all modules")
+
+	mods := kmmv1beta1.ModuleList{}
+
+	err := f.client.List(context.Background(), &mods)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/filter/filter.go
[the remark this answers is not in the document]
qbarrand: Don't we want to log failed builds, at least?

internal/filter/filter.go
```diff
@@ -75,6 +83,21 @@ var kmmClusterClaimChanged predicate.Predicate = predicate.Funcs{
...
 }
 
+var moduleJobSuccess predicate.Predicate = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		job, ok := e.ObjectNew.(*batchv1.Job)
+		if !ok {
+			return true
+		}
+		owner := metav1.GetControllerOfNoCopy(job)
+		if owner != nil && owner.Kind == "Module" && job.Status.Succeeded == 1 {
```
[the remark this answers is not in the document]
qbarrand: In fact I don't think we need to filter the owner kind, controller-runtime does that automatically if we use [`Owns`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.15.1/pkg/builder#Builder.Owns).

internal/filter/filter.go
```diff
@@ -169,6 +211,67 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
 	return reqs
 }
 
+func (f *Filter) FindModulesForNMCNodeChange(ctx context.Context, node client.Object) []reconcile.Request {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/filter/filter.go
```diff
@@ -169,6 +211,67 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
+
+	for _, mod := range mods.Items {
+		logger := logger.WithValues("module name", mod.Name)
+
+		logger.V(1).Info("Processing module")
+
+		moduleSelectorMatchNode, err := utils.IsObjectSelectedByLabels(node.GetLabels(), mod.Spec.Selector)
+		if err != nil {
+			logger.Error(err, "could not determine if node is selected by module", "node", node.GetName(), "module", mod.Name)
+			return reqs
```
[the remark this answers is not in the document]
yevgeny-shnaidman: no, changing to continue

internal/filter/filter.go
```diff
@@ -169,6 +211,67 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
+		}
+
+		nsn := types.NamespacedName{Name: mod.Name, Namespace: mod.Namespace}
+
+		reqs = append(reqs, reconcile.Request{NamespacedName: nsn})
+		reqSet.Insert(nsn)
+	}
+
+	nms, err := f.nmcHelper.Get(ctx, node.GetName())
+	if err != nil || nms == nil {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Get function does not return an error in case NMC does not exists.Maybe i should change it to return NotFound instead of nil

internal/filter/filter.go
yevgeny-shnaidman: 1) because we should also find the modules that should be reconciled in the opposite direction ( they should not run on the node anymore. If Node's labels were changed/removed, the first part of the function will find the modules that should run on the node, but we must also check the already running Modules (stored in the NMC) , in case they need to be removed
2) the functions does return NamespacedName of the modules, which are stored in nmc.Spec.Modules

internal/filter/filter.go
```diff
@@ -169,6 +211,71 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
+		if reqSet.Has(nsn) {
+			continue
+		}
+		reqs = append(reqs, reconcile.Request{NamespacedName: nsn})
+	}
+
+	logger.Info("Adding reconciliation requests", "count", len(reqs))
+	logger.V(1).Info("New requests", "requests", reqs)
+
+	return reqs
```
qbarrand, line 276: Could we use [`reqSet.UnsortedList()`](https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#Set.UnsortedList) instead? That way we don't need to maintain two separate collections.

internal/filter/filter.go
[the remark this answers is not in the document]
qbarrand: Misread the `Get` implementation sorry. Let us leave it as is for now.

internal/filter/filter.go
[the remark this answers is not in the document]
yevgeny-shnaidman: That is not related to this controller, this one is responsible for module-nmc reconciliation. Controller responsible for building/signing will log what is needed

internal/filter/filter.go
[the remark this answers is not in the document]
yevgeny-shnaidman: i am keeping the filtering, but removing the mapping function

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: removed the "update"

internal/filter/filter.go
yevgeny-shnaidman: Fixed

internal/filter/filter.go
```diff
@@ -100,6 +127,21 @@ func (f *Filter) ModuleReconcilerNodePredicate(kernelLabel string) predicate.Pre
 	)
 }
 
+func ModuleNMCReconcilerNodePredicate() predicate.Predicate {
+	return predicate.And(
+		skipDeletions,
```
qbarrand, line 132: I think we'd want to delete a NMC if the corresponding node is gone.

internal/filter/filter.go
```diff
@@ -169,6 +211,67 @@ func (f *Filter) FindModulesForNode(ctx context.Context, node client.Object) []r
...
 }
 
+// FindModulesForNMCNodeChange finds the modules that are affected by node changes that result
+// in ModuleNMCReconcilerNodePredicate predicate. First it find all the Module that can run on the node, based
+// on the Modules' Selector field and on node's labels. Then, in case NMC for the node exists, it adds all the
+// Modules already set in NMC ( in case they were not added in a previous step).
+func (f *Filter) FindModulesForNMCNodeChange(ctx context.Context, node client.Object) []reconcile.Request {
+	logger := ctrl.LoggerFrom(ctx).WithValues("node", node.GetName())
+
+	reqs := make([]reconcile.Request, 0)
```
qbarrand, line 221: I think we can remove that variable since it's now unused.

internal/filter/filter.go
yevgeny-shnaidman: Currently i see it as an optimization (it does not affect the correctness of the code right now). So i would schedule it for the next PRs

internal/filter/filter.go
yevgeny-shnaidman: fixed

controllers/module_nmc_reconciler.go
```diff
@@ -0,0 +1,241 @@
...
+				filter.ModuleNMCReconcilerNodePredicate(),
+			),
+		).
+		Watches(
+			&batchv1.Job{},
+			handler.EnqueueRequestForOwner(mgr.GetScheme(), mgr.GetRESTMapper(), &kmmv1beta1.Module{}, handler.OnlyControllerOwner()),
+			builder.WithPredicates(
+				filter.ModuleNMCReconcileJobPredicate(),
+			),
+		).
```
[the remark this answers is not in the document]
yevgeny-shnaidman: No, Owns defines it own filter and handler. Since i would like to keep the filter, i am still using Watches

PR #513, opened 2023-08-08
Add the NodeModulesConfig controller

qbarrand:
Reconcile NodeModulesConfig resources by creating and monitoring worker Pods.
Use the NodeModulesConfig status to maintain the state of modules on nodes.
Use the plural form for the NodeModulesConfig CRD, per convention.

---
Split into several PRs:
- #514


qbarrand:
/hold
Extracting some files from this PR to make it easier to review.

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+	if err := w.client.Get(ctx, types.NamespacedName{Name: nmc.Name}, &node); err != nil {
+		return fmt.Errorf("could not get node %s: %v", nmc.Name, err)
+	}
+
+	readyCondition := FindNodeCondition(node.Status.Conditions, v1.NodeReady)
+	if readyCondition == nil {
+		return fmt.Errorf("node %s has no Ready condition", nmc.Name)
+	}
+
+	if status.LastTransitionTime.Before(&readyCondition.LastTransitionTime) {
```
yevgeny-shnaidman, line 238: maybe we need to add a verification that the readyCondition is true? maybe the node is not in ready state yet. Or, even better - verify that node is schedulable.WDYT

yevgeny-shnaidman, line 238: A comment describing the scenarion (why we need to compare times) would be beneficial here

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+		statusMap[status.Namespace+"/"+status.Name] = &nmc.Status.Modules[i]
+	}
+
+	for _, mod := range nmc.Spec.Modules {
+		moduleNameKey := mod.Namespace + "/" + mod.Name
+
+		logger := logger.WithValues("module", moduleNameKey)
+
+		if err := r.helper.ProcessModuleSpec(ctrl.LoggerInto(ctx, logger), &nmc, &mod, statusMap[moduleNameKey]); err != nil {
+			return ctrl.Result{}, fmt.Errorf("error while processing module %s: %v", moduleNameKey, err)
```
yevgeny-shnaidman, line 101: do we need to return, or can we just log and store the error and continue to next module?

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+	configFileName = "config.yaml"
+	configFullPath = volMountPoingConfig + "/" + configFileName
+
+	volNameConfig       = "config"
+	volMountPoingConfig = "/etc/kmm-worker"
+)
+
+//go:generate mockgen -source=nodemodulesconfig_reconciler.go -package=controllers -destination=mock_nodemodulesconfig_reconciler.go PodManager
+
+type PodManager interface {
```
yevgeny-shnaidman, line 413: WDYT about moving this to a dedicated package?

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+			{NamespacedName: types.NamespacedName{Name: o.GetName()}},
+		}
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		Named(NodeModulesConfigReconcilerName).
+		For(&kmmv1beta1.NodeModulesConfig{}).
+		Owns(
+			&v1.Pod{},
+			builder.WithPredicates(filter.HasLabel(actionLabelKey)),
```
yevgeny-shnaidman, line 140: why do we need to filter on actionLabel? Isn't owner enough?

cmd/manager/main.go
```diff
@@ -163,6 +164,17 @@ func main() {
...
 	}
 
+	workerHelper := controllers.NewWorkerHelper(
+		client,
+		controllers.NewPodManager(client, workerImage, scheme),
+	)
+
+	ctx := ctrl.SetupSignalHandler()
+
+	if err = controllers.NewNodeModulesConfigReconciler(client, workerHelper).SetupWithManager(ctx, mgr); err != nil {
```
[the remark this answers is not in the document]
qbarrand: We now need to pass `ctx` to `SetupWithManager`.

cmd/manager/main.go
```diff
@@ -86,6 +86,7 @@ func main() {
 	}
 
 	operatorNamespace := cmd.GetEnvOrFatalError(constants.OperatorNamespaceEnvVar, setupLogger)
+	workerImage := cmd.GetEnvOrFatalError("RELATED_IMAGES_WORKER", setupLogger)
```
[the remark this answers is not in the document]
qbarrand: The convention is already `RELATED_IMAGES_*`; the OLM tooling lists all these environment variables and configures them in the `relatedImages` field of the CSV.

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+			return &c
+		}
+	}
+
+	return nil
+}
+
+//go:generate mockgen -source=nodemodulesconfig_reconciler.go -package=controllers -destination=mock_nodemodulesconfig_reconciler.go WorkerHelper
+
+type WorkerHelper interface {
```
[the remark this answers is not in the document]
qbarrand: Ideally, unit tests would belong to a different package...

config/manager/manager_worker_image_patch.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+        - name: manager
+          env:
+            - name: RELATED_IMAGES_WORKER
```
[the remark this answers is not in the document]
qbarrand: See https://github.com/kubernetes-sigs/kernel-module-management/pull/513#discussion_r1288315092

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+		// TODO maybe replace this with Owns() if we make nodes the owners of NodeModulesConfigs.
+		Watches(
+			&v1.Node{},
+			handler.EnqueueRequestsFromMapFunc(nodeToNMCMapFunc),
+			builder.WithPredicates(filter.SkipDeletions()),
+		).
+		Complete(r)
+}
+
+func workerPodName(nodeName, moduleName string) string {
```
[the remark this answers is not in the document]
qbarrand: I don't think so, Pods are namespaced.

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+		if err := r.helper.ProcessOrphanModuleStatus(ctrl.LoggerInto(ctx, logger), &nmc, status); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not create the unloader Pod for %s: %v", statusNameKey, err)
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *NodeModulesConfigReconciler) SetupWithManager(ctx context.Context, mgr manager.Manager) error {
+	err := mgr.GetCache().IndexField(ctx, &v1.Pod{}, ".spec.nodeName", func(o client.Object) []string {
```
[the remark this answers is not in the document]
qbarrand: We can then use `.spec.nodeName` as a field selector when listing all Pods. That is useful to get all Pods corresponding to an NMC.

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+			builder.WithPredicates(filter.SkipDeletions()),
+		).
+		Complete(r)
+}
+
+func workerPodName(nodeName, moduleName string) string {
+	return fmt.Sprintf("kmm-worker-%s-%s", nodeName, moduleName)
+}
+
+func GetContainerStatus(statuses []v1.ContainerStatus, name string) v1.ContainerStatus {
```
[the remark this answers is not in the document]
qbarrand: We don't, but if there is ever one, we are covered 🙂

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+		return nil
+	}
+
+	if statusNil || status.Config == nil {
+		logger.Info("Missing status or config; creating loader Pod")
+
+		return w.pm.CreateLoaderPod(ctx, nmc, spec)
+	}
+
+	if !spec.Config.Equal(status.Config) {
```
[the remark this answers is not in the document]
qbarrand: That is a good idea.

controllers/nodemodulesconfig_reconciler.go
qbarrand: > maybe we need to add a verification that the readyCondition is true?

Good point, I think we do.

> Or, even better - verify that node is schedulable.

Let's delay that one a bit - I think we want to offer custom tolerations to users.

controllers/nodemodulesconfig_reconciler.go
qbarrand: We can probably store the error and continue.

controllers/nodemodulesconfig_reconciler.go
qbarrand: What for? It is only used in the reconciler.

controllers/nodemodulesconfig_reconciler.go
qbarrand: This is probably slightly cheaper since it is just a map lookup.

controllers/nodemodulesconfig_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: but it is easier and more convenient to have unit-test in the same package, and all our "helper" interfaces already implemented as private

controllers/nodemodulesconfig_reconciler.go
yevgeny-shnaidman: you mean - kmod on master nodes?

controllers/nodemodulesconfig_reconciler.go
yevgeny-shnaidman: since we are moving build/sign to use pods also, won't it be more convenient to have a dedicated package?

controllers/nodemodulesconfig_reconciler.go
yevgeny-shnaidman: but you already have Owns... it means that it includes checking the owner reference, so it will be done anyway

controllers/nodemodulesconfig_reconciler.go
qbarrand: Yes, for instance.

controllers/nodemodulesconfig_reconciler.go
qbarrand: What commonality is there between build and worker Pods?

controllers/nodemodulesconfig_reconciler.go
qbarrand: Well then indeed we can drop the predicate entirely.

controllers/nodemodulesconfig_reconciler.go
yevgeny-shnaidman: that it is dealing with pods?...  why is pod interface needs to be at this file? It does not deal with reconciliation logic, but with creating Pod objects based on data

controllers/nodemodulesconfig_reconciler.go
qbarrand: Removed.

controllers/nodemodulesconfig_reconciler.go
qbarrand: Now checking that the condition is true.

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+	if err := r.helper.SyncStatus(ctx, &nmc); err != nil {
+		return reconcile.Result{}, fmt.Errorf("could not reconcile status for NodeModulesConfig %s: %v", nmc.Name, err)
+	}
+
+	// Statuses are now up-to-date.
+
+	statusMap := make(map[string]*kmmv1beta1.NodeModuleStatus, len(nmc.Status.Modules))
+
+	for i := 0; i < len(nmc.Status.Modules); i++ {
+		status := nmc.Status.Modules[i]
```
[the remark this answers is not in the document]
qbarrand: I think that's less readable.

controllers/nodemodulesconfig_reconciler.go
```diff
@@ -0,0 +1,681 @@
...
+	statusNil := false
+
+	if status == nil {
+		statusNil = true
+	} else if status.InProgress {
+		logger.Info("Sync status is in progress; skipping")
+		return nil
+	}
+
+	if statusNil || status.Config == nil {
```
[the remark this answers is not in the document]
qbarrand: Correct. Fixed that.

controllers/nodemodulesconfig_reconciler.go
qbarrand: Done.

controllers/nodemodulesconfig_reconciler.go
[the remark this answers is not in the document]
qbarrand: Using `reflect.DeepEqual` instead.

controllers/nodemodulesconfig_reconciler.go
qbarrand: I would only extract common code. This code is not common; it is only used once, in the reconciler.
If we extract everything, we are going to end up with a lot of packages in `internal`.

controllers/nodemodulesconfig_reconciler.go
qbarrand: Added a comment.

controllers/nodemodulesconfig_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: do we actually list the pods like that? Anyway, lets add a small comment here

controllers/nodemodulesconfig_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: right, i missed the usage

controllers/nodemodulesconfig_reconciler.go
yevgeny-shnaidman: if you can add that the reason to compare the times is in order to catch the node reboot scenario, it would be great

controllers/nodemodulesconfig_reconciler.go
yevgeny-shnaidman: i am ok with moving that decision to 2.1

internal/filter/filter.go
```diff
@@ -395,3 +395,7 @@ func NodeLabelModuleVersionUpdatePredicate(logger logr.Logger) predicate.Predica
 		},
 	}
 }
+
+func SkipDeletions() predicate.Predicate {
```
yevgeny-shnaidman, line 399: why not make the original skipDeletions to public

controllers/nodemodulesconfig_reconciler.go
qbarrand: Improved the comment. Does it look OK to you?

controllers/nodemodulesconfig_reconciler.go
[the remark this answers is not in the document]
qbarrand: We do, it allows for efficient reconciliation of the NMC status.

PR #514, opened 2023-08-08
Add `NodeModulesConfig` status management functions

qbarrand:
Use the plural form for the `NodeModulesConfig` CRD, per convention.
Add new CRD fields to be used by the `NodeModulesConfig` reconciler.


api/v1beta1/nodemodulesconfig_types.go
```diff
@@ -47,9 +54,14 @@ type NodeModulesConfigSpec struct {
...
-	Name       string `json:"name"`
-	Namespace  string `json:"namespace"`
-	KmodLoaded bool   `json:"kmodLoaded"`
+	//+optional
+	Config *ModuleConfig `json:"config,omitempty"`
+	//+optional
+	LastTransitionTime *metav1.Time     `json:"lastTransitionTime,omitempty"`
+	Name               string           `json:"name"`
+	Namespace          string           `json:"namespace"`
+	ServiceAccountName string           `json:"serviceAccountName"`
```
yevgeny-shnaidman, line 63: why the ServiceAccount is not part of the Config, if InsecurePull is?

internal/nmc/helper.go
```diff
@@ -74,3 +74,45 @@ func (h *helper) GetModuleEntry(nmc *kmmv1beta1.NodeModulesConfig, modNamespace,
 	}
 	return nil, 0
 }
+
+func findModuleStatus(statuses []kmmv1beta1.NodeModuleStatus, moduleNamespace, moduleName string) *kmmv1beta1.NodeModuleStatus {
```
yevgeny-shnaidman, line 78: it looks a little bit strange that Spec has methods and Status has functions. Can we make them all methods and then decide upon in v2.1?

internal/nmc/helper.go
```diff
@@ -74,3 +74,45 @@ func (h *helper) GetModuleEntry(nmc *kmmv1beta1.NodeModulesConfig, modNamespace,
...
+		if s.Namespace == moduleNamespace && s.Name == moduleName {
+			return &statuses[i]
+		}
+	}
+
+	return nil
+}
+
+func RemoveModuleStatus(statuses *[]kmmv1beta1.NodeModuleStatus, modNamespace, modName string) {
+	if statuses == nil || len(*statuses) == 0 {
```
yevgeny-shnaidman, line 91: what if findModuleStatus also returns an index? Then you can call it from the RemoveModuleStatus and implement removal by playing with index.WDYT?

api/v1beta1/nodemodulesconfig_types.go
```diff
@@ -20,6 +20,13 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type WorkerSyncStatus string
+
+const (
+	WorkerSyncStatusCompleted  WorkerSyncStatus = "Completed"
```
yevgeny-shnaidman, line 26: Do we need "completed" status? Can't we just remove "InProgress"?

api/v1beta1/nodemodulesconfig_types.go
qbarrand: InsecurePull is part of the worker app config because it describes how to pull the image. ServiceAccountName is not part of the config because the worker app does not need it.

internal/nmc/helper.go
qbarrand: Or we make them all functions? 🙂

internal/nmc/helper.go
qbarrand: I was trying to mimick the [apimachinery code](https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/conditions.go) which instead builds a new slice.

PR #515, opened 2023-08-08
Replace methods with functions

qbarrand:
Make methods managing NodeModulesConfig spec static functions.

/cc @yevgeny-shnaidman


yevgeny-shnaidman:
so now we have both functions and methods inside the helper... I still think that defining everything as helper makes the unit-test clearer and more functional, and the helper looks more unified.

qbarrand:
The only remaining method in the helper is `Get`, which brings limited added value (if any).

PR #518, opened 2023-08-09
Module deleted scenario in Module-NMC controller

yevgeny-shnaidman:
When a Module is deleted, all its ModuleConfigs in all the NMCs should be removed

### DIFF
--- a/controllers/module_nmc_reconciler.go
+++ b/controllers/module_nmc_reconciler.go
@@ -8,6 +8,7 @@ import (
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/api"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/nmc"
@@ -52,7 +53,6 @@ func NewModuleNMCReconciler(client client.Client,
 }
 
 func (mnr *ModuleNMCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	// get the reconciled module
 	logger := log.FromContext(ctx)
 
 	logger.Info("Starting Module-NMS reconcilation", "module name and namespace", req.NamespacedName)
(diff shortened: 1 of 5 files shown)


controllers/module_nmc_reconciler.go
```diff
@@ -126,6 +141,22 @@ func newModuleNMCReconcilerHelper(client client.Client, registryAPI registry.Reg
 	}
 }
 
+func (mnrh *moduleNMCReconcilerHelper) setFinalizer(ctx context.Context, mod *kmmv1beta1.Module) error {
+	finalizers := mod.GetFinalizers()
+	for _, finalizer := range finalizers {
+		if finalizer == constants.ModuleFinalizer {
+			return nil
+		}
+	}
```
qbarrand, line 150:
```suggestion
	if controllerutil.ContainsFinalizer(mod, constants.ModuleFinalizer) {
		return nil
	}
```

controllers/module_nmc_reconciler.go
```diff
@@ -135,6 +166,31 @@ func (mnrh *moduleNMCReconcilerHelper) getRequestedModule(ctx context.Context, n
...
+	nmcList := kmmv1beta1.NodeModulesConfigList{}
+	err := mnrh.client.List(ctx, &nmcList)
+	if err != nil {
+		return fmt.Errorf("failed to list NMCs in the cluster: %v", err)
+	}
+
+	errs := make([]error, 0, len(nmcList.Items))
+	for _, nmc := range nmcList.Items {
+		err = mnrh.removeModuleFromNMC(ctx, &nmc, mod.Namespace, mod.Name)
+		errs = append(errs, err)
```
qbarrand, line 179:
```suggestion
		errs = append(
			errs,
			mnrh.removeModuleFromNMC(ctx, &nmc, mod.Namespace, mod.Name),
		)
```
Otherwise I think we keep appending the same `err` variable.

controllers/module_nmc_reconciler.go
```diff
@@ -52,19 +53,31 @@ func NewModuleNMCReconciler(client client.Client,
...
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			logger.Info("Module deleted")
+			logger.Info("Module deleted, nothing to do")
 			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{}, fmt.Errorf("failed to get the requested %s Module: %v", req.NamespacedName, err)
 	}
+	if mod.GetDeletionTimestamp() != nil {
+		//module is being delete
```
qbarrand, line 69:
```suggestion
		// Module is being deleted
```

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: Fixed

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: fixed

controllers/module_nmc_reconciler.go
yevgeny-shnaidman: No, just verified it is working without issues. The pointer are copied into the slice, so it is not an issue to replace the err value with a new pointer

qbarrand: OK then.

PR #519, opened 2023-08-11
Move `controllers` to `internal/controllers`

qbarrand:
Rename `NodeModulesConfigReconciler` to `NMCReconciler`.
Use private interfaces for the `NMCReconciler` for consistency.

/cc @yevgeny-shnaidman


qbarrand:
@yevgeny-shnaidman PTAL

PR #522, opened 2023-08-14
Setting Worker Pod container to run as root

yevgeny-shnaidman:
Worker Pod container application tries to create directory for image pulling on /mnt/img, which fails due to the permission This PR add RunAsUser 0 to the security context


qbarrand:
/hold
Maybe we should instead create that directory in the container image and set the appropriate permissions?

yevgeny-shnaidman:
> /hold Maybe we should instead create that directory in the container image and set the appropriate permissions?

It won't help: the worker code needs to create directory based on image name under /mnt/img during runtime. We can define running as User 0 inside  the Dockerfile.worker, instead of updating the security context

PR #530, opened 2023-08-23
Fixing Unloading Kmod on a previously failed Load Pod

yevgeny-shnaidman:
This PR fixes 2 issues:
1) In case Load Pod is failing, its Config in status is nil.
   if NMC is configured to un load the module, for whatever reason,
   the code must also check if the Config is nil, and then just to
   delete the status for module, in order to avoid panic
2) When unload Pod is failing, its status should be retained as is,
   otherwise the Config is lost, and the following unload pod won't be
   able to run

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -274,7 +274,7 @@ func (w *workerHelperImpl) ProcessModuleSpec(
 
 func (w *workerHelperImpl) ProcessOrphanModuleStatus(
 	ctx context.Context,
-	nmc *kmmv1beta1.NodeModulesConfig,
+	nmcObj *kmmv1beta1.NodeModulesConfig,
 	status *kmmv1beta1.NodeModuleStatus,
 ) error {
 	logger := ctrl.LoggerFrom(ctx)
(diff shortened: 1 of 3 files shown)


internal/controllers/nmc_reconciler.go
```diff
@@ -347,27 +354,32 @@ func (w *workerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1beta1.No
...
-		status := kmmv1beta1.NodeModuleStatus{
-			Namespace: modNamespace,
-			Name:      modName,
+		status := nmc.FindModuleStatus(nmcObj.Status.Modules, modNamespace, modName)
+		if status == nil {
+			status = &kmmv1beta1.NodeModuleStatus{
+				Namespace: modNamespace,
+				Name:      modName,
+			}
 		}
```
qbarrand, line 363: Do we need to find the existing status? We will anyway populate it as desired in this function.

yevgeny-shnaidman: In case the failed pod is the Unload Pod, we need to keep the existing status as is, otherwise we lose the Config, and then we cannot re-run the Unload Pod again

PR #537, opened 2023-08-25
Reconfigure worker pod capabilities

yevgeny-shnaidman:
1) Removing privileged and adding SYS_MODULE capability 2) Worker application in worker pod will run as User 0 in order to
   be able to run modprobe

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -36,7 +36,5 @@ WORKDIR /
 
 COPY --from=builder /workspace/worker /worker
 RUN mkdir -p /mnt/img
-RUN chown 65532:65532 /mnt/img
-USER 65532:65532
 
 ENTRYPOINT ["/worker"]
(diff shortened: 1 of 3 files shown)


Dockerfile.worker
```diff
@@ -36,7 +36,5 @@ WORKDIR /
 
 COPY --from=builder /workspace/worker /worker
 RUN mkdir -p /mnt/img
-RUN chown 65532:65532 /mnt/img
```
qbarrand, line 39: Don't we need this?

yevgeny-shnaidman: No, since worker is not running under User 65532, then we don't need it

qbarrand: What user is it running as then?

yevgeny-shnaidman: the default one - "0"

qbarrand:
/lgtm
Investigate further hardening in https://github.com/kubernetes-sigs/kernel-module-management/issues/538

PR #543, opened 2023-08-29
Adding Garbage Collection for NMC objects for deleted Nodes

yevgeny-shnaidman:
This implemented by setting Node as the owner of the NMC. In case Node is deleted, NMC will also be deleted

### DIFF
--- a/internal/controllers/module_nmc_reconciler.go
+++ b/internal/controllers/module_nmc_reconciler.go
@@ -17,6 +17,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


internal/controllers/module_nmc_reconciler.go
```diff
@@ -236,17 +236,17 @@ func (mnrh *moduleNMCReconcilerHelper) enableModuleOnNode(ctx context.Context, m
 	}
 
 	nmc := &kmmv1beta1.NodeModulesConfig{
-		ObjectMeta: metav1.ObjectMeta{Name: nodeName},
+		ObjectMeta: metav1.ObjectMeta{Name: node.Name},
 	}
 
 	opRes, err := controllerutil.CreateOrPatch(ctx, mnrh.client, nmc, func() error {
-		return mnrh.nmcHelper.SetModuleConfig(nmc, mld, &moduleConfig)
+		return mnrh.nmcHelper.SetModuleConfig(nmc, mld, &moduleConfig, node)
```
qbarrand, line 243: I think I would rather call `controllerutil.SetOwnerReference` here. It does not belong to `SetModuleConfig` in my opinion, which should just manipulate the `spec.modules` list.

yevgeny-shnaidman: Fixed

PR #545, opened 2023-08-30
Fixing worker pod failure in case no pullsecrets are defined

yevgeny-shnaidman:
On initiliazation, worker pod goes over /var/run/kmm/pull-secrets directory in roder to read all pull secret and create authentication keychain. Currently, in case no pullsecrets are define, this function fails, since the directory itself does not exists. This PR defines the directory in the Dockerfile, and during pod creation, pull secrets will be mapping via volumes

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -35,6 +35,6 @@ RUN ["apk", "add", "ca-certificates", "kmod"]
 WORKDIR /
 
 COPY --from=builder /workspace/worker /worker
-RUN mkdir -p /mnt/img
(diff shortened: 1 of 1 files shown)


Dockerfile.worker
```diff
@@ -36,5 +36,6 @@ WORKDIR /
 
 COPY --from=builder /workspace/worker /worker
 RUN mkdir -p /mnt/img
+RUN mkdir -p /var/run/kmm/pull-secrets
```
qbarrand, line 39:
```suggestion
RUN ["mkdir", "-p", "/mnt/img", "/var/run/kmm/pull-secrets"]
```

yevgeny-shnaidman: Fixed

PR #548, opened 2023-09-01
Make all binaries run as regular users

qbarrand:
Create UID & GID 201 for the worker image, and use it by default.
Move the `manager` and `worker` binaries under `/usr/local/bin`.
Remove redundant `USER 65532:65532` from the manager Dockerfile.

/cc @yevgeny-shnaidman


yevgeny-shnaidman:
I have run your code on minikube and still got Error on the pod due to permissions denied. I did managed to install the module by tweaking the sudoers during the build of Dockerfile.worker

PR #549, opened 2023-09-04
Adding support for loading multiple kernel modules(softdep)

yevgeny-shnaidman:


internal/controllers/nmc_reconciler.go
```diff
@@ -749,6 +748,57 @@ func setWorkerContainerArgs(pod *v1.Pod, args []string) error {
...
+		}
+		line := prepareSoftDepLine(modulesNames[i], modulesNames[i+1])
+		softDepData = softDepData + line
+	}
+	return softDepData
+}
+
+func prepareSoftDepLine(dependendModuleName, moduleName string) string {
+	return fmt.Sprintf("softdep %s pre: %s\n", dependendModuleName, moduleName)
+}
```
qbarrand, line 791: Why do we need this function?

internal/controllers/nmc_reconciler.go
```diff
@@ -749,6 +748,57 @@ func setWorkerContainerArgs(pod *v1.Pod, args []string) error {
...
+		Name:      "modules-order",
+		ReadOnly:  true,
+		MountPath: "/etc/modprobe.d",
+	}
+	pod.Spec.Volumes = append(pod.Spec.Volumes, softdepVolume)
+	pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, softDepVolumeMount)
+}
+
+func getModulesOrderAnnotationValue(modulesNames []string) string {
+	softDepData := ""
```
qbarrand, line 778: I think we should use a `strings.Builder` here.

internal/controllers/nmc_reconciler.go
```diff
@@ -749,6 +748,57 @@ func setWorkerContainerArgs(pod *v1.Pod, args []string) error {
...
+			},
+		},
+	}
+	softDepVolumeMount := v1.VolumeMount{
+		Name:      "modules-order",
+		ReadOnly:  true,
+		MountPath: "/etc/modprobe.d",
+	}
+	pod.Spec.Volumes = append(pod.Spec.Volumes, softdepVolume)
+	pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, softDepVolumeMount)
```
qbarrand, line 774: Could we use `podcmd.FindContainerByName` like in other functions in this file to be more future-proof?

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: Fixed

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: Fixed

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: i added it for clearer code, so that we can see what module goes where in the file, but i removed it when moved to builder

PR #551, opened 2023-09-05
Use the `default` ServiceAccount when creating NMC

qbarrand:
Using `default` instead of an empty string allows us to attach all pull secrets configured for that account when setting up the mounts for the worker Pod.

/cc @yevgeny-shnaidman

### DIFF
--- a/config/manager/controller_manager_config.yaml
+++ b/config/manager/controller_manager_config.yaml
@@ -4,3 +4,6 @@ webhookPort: 9443
 leaderElection:
   enabled: true
   resourceID: kmm.sigs.x-k8s.io
+worker:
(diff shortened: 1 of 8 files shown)


config/manager/controller_manager_config.yaml
```diff
@@ -4,3 +4,5 @@ webhookPort: 9443
 leaderElection:
   enabled: true
   resourceID: kmm.sigs.x-k8s.io
+worker:
+  seLinuxType: spc_t
```
yevgeny-shnaidman, line 8: just adding seLinuxType won't help in case of seLinux issue ( as happens on my minikube). You can simply add runAsUser: 0

PR #552, opened 2023-09-05
Explicitly checking PodPending when reporting Build/Sign status

yevgeny-shnaidman:
Currently, PodPending phase of a running pod is not explicitly checked, which may cause an unnecessary errors reported. It does not affect the correctness of the flow, but does causes confusion in logs

### DIFF
--- a/internal/utils/podhelper.go
+++ b/internal/utils/podhelper.go
@@ -119,12 +119,12 @@ func (ph *podHelper) CreatePod(ctx context.Context, pod *v1.Pod) error {
 // GetPodStatus returns the status of a Pod, whether the latter is in progress or not and
 // whether there was an error or not
 func (ph *podHelper) GetPodStatus(pod *v1.Pod) (Status, error) {
-	switch {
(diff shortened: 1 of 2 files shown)


internal/utils/podhelper.go
```diff
@@ -124,6 +124,8 @@ func (ph *podHelper) GetPodStatus(pod *v1.Pod) (Status, error) {
 		return StatusCompleted, nil
 	case pod.Status.Phase == v1.PodRunning:
 		return StatusInProgress, nil
+	case pod.Status.Phase == v1.PodPending:
```
qbarrand, line 127: Can we refactor the whole `switch`?
```go
	switch pod.Status.Phase {
	case v1.PodSucceeded:
		return StatusCompleted, nil
	case v1.PodRunning, v1.PodPending:
		return StatusInProgress, nil
	case v1.PodFailed:
		return StatusFailed, nil
	default:
		return "", fmt.Errorf("unknown status: %v", pod.Status)
}

PR #573, opened 2023-09-20
Record more fields into the NMC status

qbarrand:
Record the service account name as well as the `imageRepoSecret` used in a successful loader Pod into the NMC status so that we can later unload with the same properties.
The service account name is stored as part of the PodSpec.
For the `imageRepoSecret` name, add that volume under a well-known name that can be looked up by `SyncStatus()`.

/cc @yevgeny-shnaidman

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/labels"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/nmc"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/worker"
 	v1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


internal/controllers/nmc_reconciler.go
```diff
@@ -424,14 +424,20 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-					Namespace:          modNamespace,
-					ServiceAccountName: "",
+					Name:      modName,
+					Namespace: modNamespace,
 				},
 			}
 		}
 
+		if irsName, err := getImageRepoSecretName(&p); err != nil {
+			return fmt.Errorf("error while looking for the imageRepoSecret volume: %v", err)
```
yevgeny-shnaidman, line 434: we should record the error and continue instead of returning an error directly

internal/controllers/nmc_reconciler.go
```diff
@@ -508,8 +514,9 @@ const (
 	configFileName = "config.yaml"
 	configFullPath = volMountPoingConfig + "/" + configFileName
 
-	volNameConfig       = "config"
-	volMountPoingConfig = "/etc/kmm-worker"
+	volNameConfig          = "config"
+	volNameImageRepoSecret = "image-repo-secret"
+	volMountPoingConfig    = "/etc/kmm-worker"
```
yevgeny-shnaidman, line 519: volMountPoingConfig -> volMountPointConfig  ?

internal/controllers/nmc_reconciler.go
```diff
@@ -508,8 +514,9 @@ const (
 	configFileName = "config.yaml"
 	configFullPath = volMountPoingConfig + "/" + configFileName
 
-	volNameConfig       = "config"
-	volMountPoingConfig = "/etc/kmm-worker"
+	volNameConfig          = "config"
+	volNameImageRepoSecret = "image-repo-secret"
```
yevgeny-shnaidman, line 518: Where is this name is set into the VolumeMount?

internal/controllers/nmc_reconciler.go
qbarrand: We control the Pod creation so this should never happen. However maybe we can make it a warning log line instead.

yevgeny-shnaidman: warning log and continue works for me

internal/controllers/nmc_reconciler.go
qbarrand: Fixed.

internal/controllers/nmc_reconciler.go
qbarrand: Now correctly used [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/573/files#diff-4abf41a7395eeba9459df6b2fe59c88e93e33e83dd683d40878c61651de912f6R900). Thanks!

internal/controllers/nmc_reconciler.go
qbarrand: Done.

PR #577, opened 2023-09-26
Publish build & sign events

qbarrand:
Add a new reconciler to watch build & signing Pods.


internal/controllers/job_event_reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+			return ctrl.Result{}, fmt.Errorf("error while getting Module %s: %v", moduleNSN, err)
+		}
+	}
+
+	eventAnnotations := map[string]string{
+		"kernel-version": kernelVersion,
+		"pod-name":       pod.Name,
+	}
+
+	if _, ok := pod.GetAnnotations()[createdAnnotationKey]; !ok {
```
yevgeny-shnaidman, line 124: Why do we need this whole annotation handling? Can't we just create and event on PodRunning? Do we even want an event other than Failed or Succeeded for build/sign?

internal/controllers/job_event_reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+
+	hasJobEventFinalizer := predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return controllerutil.ContainsFinalizer(o, constants.JobEventFinalizer)
+	})
+
+	p := predicate.And(
+		isJobPod,
+		predicate.Or(
+			hasJobEventFinalizer,
+			predicate.Not(
```
yevgeny-shnaidman, line 225: Why do we need predicate for not having Annotation? Why not just check job finalizer existence?

cmd/manager-hub/main.go
```diff
@@ -151,6 +152,12 @@ func main() {
 		cmd.FatalError(ctrlLogger, err, "unable to create controller")
 	}
 
+	er := mgr.GetEventRecorderFor("kmm")
```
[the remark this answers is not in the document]
qbarrand: Changed to `eventRecorder`.

internal/controllers/job_event_reconciler.go
```diff
@@ -0,0 +1,232 @@
...
+		if k8serrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+
+		return reconcile.Result{}, fmt.Errorf("could not get pod %s: %v", req.NamespacedName, err)
+	}
+
+	jobType := pod.Labels[constants.PodType]
+
+	if jobType == "" {
```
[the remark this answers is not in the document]
qbarrand: Removed the `if`.

internal/controllers/job_event_reconciler.go
qbarrand: The annotation is there to make sure that we do not send the `created` event twice.

yevgeny-shnaidman: If the created event will be send on PodRunningPhase, won't it be sufficient? Reconciliation will be re-scheduled only in case something has changed in Pod status, which means that the Phase has changed from Running to Success/Failure, no?

qbarrand: Let us consider the following scenario:
- pod P is running, we send the `created` event;
- our controller goes down and back up;
- when it starts, it processes all watched pods, including P

In that case we have no way to determine if we sent the event already or not.

PR #580, opened 2023-09-27
Removing DaemonSet role from the DevicePlugin labels

yevgeny-shnaidman:


internal/controllers/node_label_module_version_reconciler_test.go
```diff
@@ -167,17 +167,44 @@ var _ = Describe("getDevicePluginPods", func() {
...
+		res, err := helper.getDevicePluginPods(ctx, nodeName)
 		Expect(err).ToNot(HaveOccurred())
+		Expect(len(res)).To(Equal(1))
+		Expect(res[0]).To(Equal(pod1))
 	})
 
 	It("error flow", func() {
 		fieldSelector := client.MatchingFields{"spec.nodeName": nodeName}
-		labelSelector := client.HasLabels{constants.DaemonSetRole}
+		labelSelector := client.HasLabels{constants.ModuleNameLabel}
```
qbarrand, line 207: The worker pods also have that label. Aren't they going to interfere?

yevgeny-shnaidman: That's why i have added the verification fro OwnerReference kind to DaemonSet

PR #583, opened 2023-09-28
Record kmod load & unload events

### LINKED ISSUE #473: Expose events

qbarrand:
Record the ModuleLoaded and ModuleUnloaded events.

Fixes #473 

/cc @yevgeny-shnaidman

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/kubectl/pkg/cmd/util/podcmd"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -67,9 +68,10 @@ func NewNMCReconciler(
 	scheme *runtime.Scheme,
 	workerImage string,
 	workerCfg *config.Worker,
+	recorder record.EventRecorder,
 ) *NMCReconciler {
 	pm := newPodManager(client, workerImage, scheme, workerCfg)
-	helper := newWorkerHelper(client, pm)
+	helper := newNMCReconcilerHelper(client, pm, recorder)
 	return &NMCReconciler{
 		client: client,
 		helper: helper,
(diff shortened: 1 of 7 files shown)


internal/controllers/nmc_reconciler.go
```diff
@@ -520,56 +524,101 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+		statusConfig, ok := statusLabels[nsn]
+		if ok && reflect.DeepEqual(specConfig, statusConfig) && !nodeModuleReadyLabels.Has(nsn) {
+			loaded = append(loaded, nsn)
 		}
 	}
 
-	return h.client.Patch(ctx, &node, patchFrom)
+	patchFrom := client.MergeFrom(node.DeepCopy())
+
+	for _, nsn := range unloaded {
```
yevgeny-shnaidman, line 580: I think that this loop and the next one a redundant: you can just call meta.RemoveLabel and meta.SetLabel in the loops where you update loaded and unloaded slices

internal/controllers/nmc_reconciler.go
```diff
@@ -520,56 +524,101 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+			utils.GetKernelModuleReadyNodeLabel(nsn.Namespace, nsn.Name),
+			"",
+		)
+	}
+
+	if err := h.client.Patch(ctx, &node, patchFrom); err != nil {
+		return fmt.Errorf("could not patch node: %v", err)
+	}
+
+	for _, nsn := range unloaded {
```
yevgeny-shnaidman, line 599: a couple of  general questions: 
1) AnnotatedEventF is a void function ( no error is returned). So why not just call the recorder in the SyncStatus, where we know exactly that Load/Unload worker pods have finished successfully?
2)  For build event we create events in a separate controller asynchronously, while for load/unload events we do it synchronously  in the NMC controller.  I think we should decide on a unified way to do it

internal/controllers/nmc_reconciler.go
qbarrand: I thought it would be clearer to use the slices for both the events and the labels. I can change it of you insist.

internal/controllers/nmc_reconciler.go
qbarrand: > AnnotatedEventF is a void function ( no error is returned). So why not just call the recorder in the SyncStatus, where we know exactly that Load/Unload worker pods have finished successfully?

That is another option. If we moved event recording to `SyncStatus`, we would still have to create two slices for loaded / unloaded, since we do not want to send the events before the status is updated (to avoid sending duplicate events when the status update fails). Please let me know what you prefer.

> For build event we create events in a separate controller asynchronously, while for load/unload events we do it synchronously in the NMC controller. I think we should decide on a unified way to do it

We could also send these in a separate controller. I thought it would be simpler to do it here in that function that updates the labels, because it is kind of the same thing. Please let me know what you prefer.

PR #586, opened 2023-09-29
Make the worker set `firmware_class.path`

qbarrand:
Add an operator-wide setting for the firmware path on the node to be set by loading Pods.
Create loading Pods as privileged when they need to set the firmware path; this is required to be able to write to /sys.


cmd/worker/funcs_kmod.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/kubernetes-sigs/kernel-module-management/internal/worker"
+	"github.com/spf13/cobra"
+)
+
+func kmodLoadFunc(cmd *cobra.Command, args []string) error {
```
yevgeny-shnaidman, line 10: what about moving the whole  kmodCmds from main to here? For me it would more sense to create a separate file for all cobra commands, instead just ffor one function used as a RunE function in one specific command

internal/controllers/nmc_reconciler.go
```diff
@@ -867,6 +877,21 @@ func setWorkerContainerArgs(pod *v1.Pod, args []string) error {
 	return nil
 }
 
+func setWorkerPrivileged(pod *v1.Pod, privileged bool) error {
```
yevgeny-shnaidman, line 880: we are already setting security context in the baseWorkerPod function. Can we just add an if clause there and set the priviliged flag?

yevgeny-shnaidman:
I think that this option ( defining the new firmware path in the KMM config) should be well document and marked as a potential bug risk for the users. Unlike the previous option (using MachineConfig), the new option changes the firmware path for all the relevant nodes without any notification to the rest of the cluster components. If there are other hotplug drivers/cluster components that rely on default value, they will never understand when and how the path was changed

internal/controllers/nmc_reconciler.go
qbarrand: We could, but then we would keep adding parameters to that function. If you insist, I can change it.

cmd/worker/funcs_kmod.go
qbarrand: Absolutely, but in a different PR. I kept the changes minimal in this one.

internal/controllers/nmc_reconciler.go
```diff
@@ -660,14 +660,32 @@ func (p *podManagerImpl) CreateLoaderPod(ctx context.Context, nmc client.Object,
 		return fmt.Errorf("could not create the base Pod: %v", err)
 	}
 
-	if err = setWorkerContainerArgs(pod, []string{"kmod", "load", configFullPath}); err != nil {
+	args := []string{"kmod", "load", configFullPath}
```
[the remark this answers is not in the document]
qbarrand: I don't think that would make things clear: now `Create(Un)LoaderPod` functions clearly set the security context to what they need. If we move that logic to another function, I'm afraid we increase complexity for little benefit.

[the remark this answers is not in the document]
yevgeny-shnaidman: We have baseWorkerPod that creates basic Pod based on nmc, we have setWorkerContainerArgs creating arguments based on the parameter and we have setWorkerConfigAnnotation creating annotations based on Config. So why not let  setWorkerSecurityContext set the security context based on WorkerCfg?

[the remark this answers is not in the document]
qbarrand: OK, went for your suggestion.

PR #591, opened 2023-10-04
Fix `prepareSchedulingData` loop variable capture

qbarrand:
Sending the `*v1.Node` loop variable from `prepareSchedulingData` to `prepareNodeSchedulingData` led to an invalid scheduling data map in which all values' `node` field was pointing to the same `v1.Node` object - the latest item in the `targetedNodes` slice.
Send the `v1.Node` to `prepareNodeSchedulingData` instead of a pointer to solve this problem.

/cc @yevgeny-shnaidman


yevgeny-shnaidman:
Nice catch!

PR #592, opened 2023-10-04
Fix device-plugin/worker pod labels validationsduring ordered-upgrade

yevgeny-shnaidman:
This PR inforces the correct ordering of add/removing version labels for worker pods and device-plugin. The correct sequence should always be:
1. remove old version device-plugin label (if exists)
2. remove old version worker pod label (only in case no relevant device-plugin pod is present)
3. add new version worker pod label (only in case the kernel module is not loaded)
4. add new version device-plugin label ( if needed)

We are stalling for worker-pod labels in order to make sure that new device-label pod will not be scheduled prior to new version of kernel module being loaded. Otherwise, old version will never be able to be removed from the kernel

### DIFF
--- a/internal/controllers/node_label_module_version_reconciler.go
+++ b/internal/controllers/node_label_module_version_reconciler.go
@@ -62,7 +62,9 @@ func (nlmvr *NodeLabelModuleVersionReconciler) Reconcile(ctx context.Context, re
 		return ctrl.Result{}, fmt.Errorf("could get device plugin pods for the node %s: %v", node.Name, err)
 	}
 
-	reconLabelsRes := nlmvr.helperAPI.reconcileLabels(modulesVersionLabels, devicePluginPods)
+	loadedKernelModules := nlmvr.helperAPI.getLoadedKernelModules(node.GetLabels())
+
+	reconLabelsRes := nlmvr.helperAPI.reconcileLabels(modulesVersionLabels, devicePluginPods, loadedKernelModules)
 
 	logger := log.FromContext(ctx).WithValues("node name", node.Name)
 	logLabelsUpdateData(logger, reconLabelsRes)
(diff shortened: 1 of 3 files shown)


internal/controllers/node_label_module_version_reconciler.go
```diff
@@ -141,6 +144,18 @@ func (nlmvha *nodeLabelModuleVersionHelper) getDevicePluginPods(ctx context.Cont
 	return devicePluginPods, nil
 }
 
+func (nlmvha *nodeLabelModuleVersionHelper) getLoadedKernelModules(node v1.Node) []types.NamespacedName {
```
qbarrand, line 147: How about replacing the `v1.Node` parameter with a `map[string]string` (the labels) directly?
Also, I don't think this should be a method.

internal/controllers/node_label_module_version_reconciler.go
```diff
@@ -192,6 +211,7 @@ func (nlmvha *nodeLabelModuleVersionHelper) reconcileLabels(modulesLabels map[st
 // validity is checked by verifying that devicePlugin pod defined
 // by name, namespace,role is missing. In case the pod is present, no matter in
 // what state, then the action is invalid
+/*
```
qbarrand, line 214: Can we remove that commented out function?

yevgeny-shnaidman: fixed

internal/controllers/node_label_module_version_reconciler.go
yevgeny-shnaidman: changed the v1.Node to map[string]string
Regarding method -  we know that we don't agree on that :). Lets leave that discussion for the refactoring task that we scheduled for v2.1

PR #596, opened 2023-10-08
Adding support for firmware loading

yevgeny-shnaidman:
1) mapping of the host firmware directory into the load/unload pod only
   in case firmware for the Module is defined
2) during Loading operation, worker pod is copying all firmware files and
   directories from image onto the mapped host directory
3) during Unloading operation, worker pod is deleting all firmware
   files, but not directories ( since those can be shared with other
   modules' firmware files
4) unit tests

### DIFF
--- a/cmd/worker/funcs_kmod_test.go
+++ b/cmd/worker/funcs_kmod_test.go
@@ -46,26 +46,37 @@ var _ = Describe("kmodLoadFunc", func() {
 	})
 
 	DescribeTable(
-		"should try and set firmware_class.path if the flag is defined",
-		func(flagValue *string) {
+		"set firmware class and firmware mount path if defined",
+		func(flagFirmwareClassPathValue, flagFirmwareMountPath *string) {
 			cfg := &kmmv1beta1.ModuleConfig{}
 			ctx := context.TODO()
 
 			cmd := &cobra.Command{}
 			cmd.SetContext(ctx)
 			cmd.Flags().String(worker.FlagFirmwareClassPath, "", "")
+			cmd.Flags().String(worker.FlagFirmwareMountPath, "", "")
 
 			readConfig := ch.EXPECT().ReadConfigFile(configPath).Return(cfg, nil)
-			loadKmod := wo.EXPECT().LoadKmod(ctx, cfg).After(readConfig)
-
-			if flagValue != nil {
+			var loadKmod *gomock.Call
+			if flagFirmwareMountPath != nil {
 				Expect(
-					cmd.Flags().Set(worker.FlagFirmwareClassPath, *flagValue),
+					cmd.Flags().Set(worker.FlagFirmwareMountPath, *flagFirmwareMountPath),
+				).NotTo(
+					HaveOccurred(),
+				)
+				loadKmod = wo.EXPECT().LoadKmod(ctx, cfg, *flagFirmwareMountPath).After(readConfig)
+			} else {
+				loadKmod = wo.EXPECT().LoadKmod(ctx, cfg, "").After(readConfig)
+			}
+
+			if flagFirmwareClassPathValue != nil {
+				Expect(
+					cmd.Flags().Set(worker.FlagFirmwareClassPath, *flagFirmwareClassPathValue),
 				).NotTo(
 					HaveOccurred(),
 				)
 
-				setFirmwarePath := wo.EXPECT().SetFirmwareClassPath(*flagValue)
+				setFirmwarePath := wo.EXPECT().SetFirmwareClassPath(*flagFirmwareClassPathValue)
 				setFirmwarePath.After(readConfig)
 				loadKmod.After(setFirmwarePath)
 			}
@@ -76,8 +87,10 @@ var _ = Describe("kmodLoadFunc", func() {
 				HaveOccurred(),
 			)
 		},
-		Entry("flag not defined", nil),
-		Entry("flag defined and empty", pointer.String("")),
-		Entry("flag defined and not empty", pointer.String("/some/path")),
+		Entry("flags not defined", nil, nil),
+		Entry("class path defined and empty, mount path not defined", pointer.String(""), nil),
+		Entry("class path defined and not empty, mount path not defined", pointer.String("/some/path"), nil),
+		Entry("class path defined and empty, mount path defined", pointer.String(""), pointer.String("some mount path")),
+		Entry("class path defined and not empty, mount path defined", pointer.String("/some/path"), pointer.String("some mount path")),
 	)
 })
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -58,7 +58,9 @@ var kmodUnloadCmd = &cobra.Command{
 			return fmt.Errorf("could not read config file %s: %v", cfgPath, err)
 		}
 
-		return w.UnloadKmod(cmd.Context(), cfg)
+		mountPathFlag := cmd.Flags().Lookup(worker.FlagFirmwareMountPath)
+
+		return w.UnloadKmod(cmd.Context(), cfg, mountPathFlag.Value.String())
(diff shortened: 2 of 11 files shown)


config/manager/kustomization.yaml
```diff
@@ -9,11 +9,11 @@ patches:
...
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
+  newTag: firmware
 - name: worker
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-worker
-  newTag: latest
+  newName: quay.io/yshnaidm/kernel-module-management-worker
+  newTag: firmware
```
qbarrand, line 16: Please discard this.

internal/worker/worker.go
```diff
@@ -50,8 +52,14 @@ func (w *worker) LoadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig) err
 		}
 	}
 
-	// TODO copy firmware
-	// TODO handle ModulesLoadingOrder
+	// prepare firmware
+	if cfg.Modprobe.FirmwarePath != "" {
+		baseImageDir := pr.fsDir + "/" + cfg.Modprobe.FirmwarePath
```
qbarrand, line 57: I'd rather use `filepath.Join`.

internal/worker/worker.go
```diff
@@ -129,7 +137,32 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig) e
 		return fmt.Errorf("could not unload module %s: %v", moduleName, err)
 	}
 
-	// TODO remove firmware
+	//remove firmware files only (no directories)
+	if cfg.Modprobe.FirmwarePath != "" {
+		baseImageDir := pr.fsDir + "/" + cfg.Modprobe.FirmwarePath
```
qbarrand, line 142:
```suggestion
		baseImageDir := filepath.Join(pr.fsDir, cfg.Modprobe.FirmwarePath)
```

qbarrand: Also I think we could find a better name for `baseImageDir`. Maybe `imageFirmwarePath`?

internal/worker/worker.go
```diff
@@ -129,7 +137,32 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig) e
...
+			if err != nil {
+				return err
+			}
+			if !info.IsDir() {
+				relPath, err := filepath.Rel(baseImageDir, path)
+				if err != nil {
+					w.logger.Info(utils.WarnString("failed to get relative path"), "basePath", baseImageDir, "path", path, "error", err)
+					return nil
+				}
+				fileToRemove := firmwareMountPath + "/" + relPath
```
qbarrand, line 153:
```suggestion
				fileToRemove := filepath.Join(firmwareMountPath, relPath)
```

internal/worker/worker.go
```diff
@@ -50,8 +52,14 @@ func (w *worker) LoadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig) err
 		}
 	}
 
-	// TODO copy firmware
-	// TODO handle ModulesLoadingOrder
+	// prepare firmware
+	if cfg.Modprobe.FirmwarePath != "" {
+		baseImageDir := pr.fsDir + "/" + cfg.Modprobe.FirmwarePath
+		w.logger.Info("preparing fimrware for loading", "image directory", baseImageDir, "host mount directory", firmwareMountPath)
+		if err = cp.Copy(baseImageDir, firmwareMountPath); err != nil {
```
qbarrand, line 59: What this function does in case `baseImageDir` exists is not very clear:
```
	// Merge preserves or overwrites existing files under the dir (default behavior).
```
https://pkg.go.dev/github.com/otiai10/copy#DirExistsAction
When does it replace and when does it overwrite?
Also, WDYT about using the `OnError` option to log copy errors that may occur?

cmd/worker/funcs_kmod_test.go
```diff
@@ -76,8 +87,10 @@ var _ = Describe("kmodLoadFunc", func() {
...
 			)
 		},
-		Entry("flag not defined", nil),
-		Entry("flag defined and empty", pointer.String("")),
-		Entry("flag defined and not empty", pointer.String("/some/path")),
+		Entry("flags not defined", nil, nil),
+		Entry("class path defined and empty, mount path not defined", pointer.String(""), nil),
+		Entry("class path  defined and not empty, mount path not defined", pointer.String("/some/path"), nil),
+		Entry("class path  defined and empty, mount path defined", pointer.String(""), pointer.String("some mount path")),
+		Entry("class path  defined and not empty, mount path defined", pointer.String("/some/path"), pointer.String("some mount path")),
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

cmd/worker/main.go
```diff
@@ -40,14 +40,14 @@ var kmodCmd = &cobra.Command{
 var kmodLoadCmd = &cobra.Command{
 	Use:   "load",
 	Short: "Load a kernel module",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.RangeArgs(1, 2),
```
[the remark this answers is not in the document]
yevgeny-shnaidman: You are right, my bad. Fixed

cmd/worker/main.go
```diff
@@ -40,14 +40,14 @@ var kmodCmd = &cobra.Command{
...
 	RunE:  kmodLoadFunc,
 }
 
 var kmodUnloadCmd = &cobra.Command{
 	Use:   "unload",
 	Short: "Unload a kernel module",
-	Args:  cobra.ExactArgs(1),
+	Args:  cobra.RangeArgs(1, 2),
```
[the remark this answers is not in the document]
yevgeny-shnaidman: You are right, my bad. Fixed

config/manager/kustomization.yaml
yevgeny-shnaidman: Fixed

internal/worker/worker.go
yevgeny-shnaidman: Fixed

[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

internal/worker/worker.go
```diff
@@ -50,8 +52,14 @@ func (w *worker) LoadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig) err
 		}
 	}
 
-	// TODO copy firmware
-	// TODO handle ModulesLoadingOrder
+	// prepare firmware
+	if cfg.Modprobe.FirmwarePath != "" {
+		baseImageDir := pr.fsDir + "/" + cfg.Modprobe.FirmwarePath
+		w.logger.Info("preparing fimrware for loading", "image directory", baseImageDir, "host mount directory", firmwareMountPath)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/worker/worker.go
yevgeny-shnaidman: Fixed

internal/worker/worker.go
yevgeny-shnaidman: Fixed

internal/worker/worker.go
yevgeny-shnaidman: You probably mean in case there are already directories/files exist at destination, right? I have looked at the copy package code and run a few tests, and i think we are ok: if directory already exists, then new files will be added,  files with the same name will be overwritten, files present only at dest will be kept as they are. 
Also, added onError function

PR #609, opened 2023-10-18
Set the worker restart policy to `OnFailure`

qbarrand:
Try loading or unloading kernel modules until either:
- the Module is deleted, or;
- the Pod template has changed.

Remove the `inProgress` field of the NMC statuses. To determine if a worker Pod already exists, simply query the cache / API server.
Make `GarbageCollectInUseLabels` list all Pods on the node to determine the full sets of Modules in use.

Flowcharts of the new implementation of `ProcessModuleSpec` and `ProcessUnconfiguredModuleStatus`:
```mermaid
---
title: ProcessModuleSpec
---

flowchart TB
    PE{Pod Exists?}
    ILP{Is Loader Pod?}
    RC{RestartCount > 0?}
    CLP[Create Loader Pod]
    CUP[Create Unloader Pod]
    E[End]
    HD{Hash differs?}
    DP[Delete Pod]
    CD{Spec config\ndiffers with status?}
    OLLT{lastTransitionTime\ntoo old?}
    
    PE -- Yes --> ILP
    PE -- No --> CD
    
    ILP -- Yes --> RC
    ILP -- No --> E
    
    CD -- Yes --> CUP
    CD -- No --> OLLT
    
    OLLT -- Yes --> CLP
    OLLT -- No --> E
    
    RC -- Yes --> HD
    RC -- No --> E
    
    HD -- Yes --> DP
    HD -- No --> E
    
    DP --> E
    CUP --> E
    CLP --> E
```

```mermaid
---
title: ProcessUnconfiguredModuleStatus
---

flowchart TB
    PE{Pod Exists?}
    ILP{Is Loader Pod?}
    RC{RestartCount > 0?}
    CUP[Create Unloader Pod]
    E[End]
    HD{Hash differs?}
    DP[Delete Pod]

    PE -- Yes --> ILP
    PE -- No --> CUP

    ILP -- Yes --> DP
    ILP -- No --> RC

    RC -- Yes --> HD
    RC -- No --> E

    HD -- Yes --> DP
    HD -- No --> E

    DP --> E
    CUP --> E
```

/cc @yevgeny-shnaidman


internal/controllers/nmc_reconciler.go
```diff
@@ -477,29 +519,28 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-				errs,
-				fmt.Errorf("%s: unhandled Pod phase %q", podNSN, phase),
-			)
-		}
+			status.LastTransitionTime = podLTT
 
-		if updateModuleStatus {
 			nmc.SetModuleStatus(&nmcObj.Status.Modules, *status)
+
+			podsToDelete = append(podsToDelete, p)
```
yevgeny-shnaidman, line 543: what about default switch case? just in case, at least we will have some log

internal/controllers/nmc_reconciler.go
```diff
@@ -429,33 +484,15 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-		}
-
-		status.ServiceAccountName = p.Spec.ServiceAccountName
-
-		updateModuleStatus := false
 
 		switch phase {
+		case v1.PodRunning:
+			// Delete Pod if orphan
+			if !specEntries.Has(types.NamespacedName{Namespace: modNamespace, Name: modName}) && status == nil {
```
yevgeny-shnaidman, line 491: not sure i understand the handling of Running Pod: in case spec is missing we delete the running pod? What if it the Unload pod? Why do we want to delete it?

internal/controllers/nmc_reconciler.go
```diff
@@ -651,10 +692,86 @@ func newPodManager(client client.Client, workerImage string, scheme *runtime.Sch
...
+	logger.V(1).Info("Listing worker Pods")
+
+	if err := p.client.List(ctx, &pl, hl, mf); err != nil {
+		return nil, fmt.Errorf("could not list worker pods for node %s: %v", nodeName, err)
+	}
+
+	return pl.Items, nil
+}
+
+func (p *podManagerImpl) MakeLoaderPod(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleSpec) (*v1.Pod, error) {
```
yevgeny-shnaidman, line 771: Maybe rename it to LoaderPodTemplate, or something like that? It is a little bit confusing, the expectation from the name is that the pod is actually created

internal/controllers/nmc_reconciler.go
```diff
@@ -429,33 +484,15 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-		updateModuleStatus := false
 
 		switch phase {
+		case v1.PodRunning:
+			// Delete Pod if orphan
+			if !specEntries.Has(types.NamespacedName{Namespace: modNamespace, Name: modName}) && status == nil {
+				logger.Info("Orphan pod; deleting")
+				podsToDelete = append(podsToDelete, p)
+			}
 		case v1.PodFailed:
```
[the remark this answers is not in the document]
qbarrand: `Failed` is when the Pod is deleted, some of its containers had a return code != 0 and it has a finalizer.

internal/controllers/nmc_reconciler.go
qbarrand: Do we want to log in that case? I'd rather avoid having messages that are not really useful.

internal/controllers/nmc_reconciler.go
qbarrand: WDYT about `LoaderPod`?

internal/controllers/nmc_reconciler.go
qbarrand: It cannot be an unloader Pod because we check `status == nil`. This is to handle cases where we have a restarting loading Pod but the Module was deleted - and so was the spec entry.

internal/controllers/nmc_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: but we are removing finalizer prior to deletion, no?

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: I think that in that case it means that something is really fucked up, and we would like to see notification in log. We are not going to see multiple message in usual flow

internal/controllers/nmc_reconciler.go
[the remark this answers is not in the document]
qbarrand: We are, this function calls `DeletePod` which does that.

internal/controllers/nmc_reconciler.go
qbarrand: One of the other phases is `Pending`, I don't think we want to see anything for it.

yevgeny-shnaidman: ok, then lets just log the Unknown, just in case

internal/controllers/nmc_reconciler.go
qbarrand: Done, PTAL.

PR #619, opened 2023-10-24
adapting Worker Pods status report in Module to v.2

yevgeny-shnaidman:
This commits adds implementation of reporting worker pods status into the Module's status.
The following fields are reported:
1) number of targeted nodes ( based on selector field) 2) number of desired nodes that kernel module should be deployed (based
   on NMCs configured)
3) current number of the nodes that the kernel module is already
   deployed to ( based on NMC spec and status configs)

### DIFF
--- a/internal/controllers/module_nmc_reconciler.go
+++ b/internal/controllers/module_nmc_reconciler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"strings"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
@@ -114,6 +115,9 @@ func (mnr *ModuleNMCReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		errs = append(errs, err)
 	}
 
+	err = mnr.reconHelper.moduleUpdateWorkerPodsStatus(ctx, mod, targetedNodes)
+	errs = append(errs, err)
+
 	err = errors.Join(errs...)
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to reconcile module %s/%s config: %v", mod.Namespace, mod.Name, err)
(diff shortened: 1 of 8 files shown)


internal/controllers/module_nmc_reconciler.go
```diff
@@ -402,6 +408,37 @@ func (mnrh *moduleNMCReconcilerHelper) removeModuleFromNMC(ctx context.Context,
...
+	if err != nil {
+		return fmt.Errorf("failed to get configured NMCs for module %s/%s: %v", mod.Namespace, mod.Name, err)
+	}
+
+	numAvailable := 0
+	for _, nmc := range nmcs {
+		modSpec, _ := mnrh.nmcHelper.GetModuleSpecEntry(&nmc, mod.Namespace, mod.Name)
+		if modSpec == nil {
+			logger.Info(utils.WarnString(
+				fmt.Sprintf("module %s/%s spec is missing in NMC %s although config label is present", mod.Namespace, mod.Name, nmc.Name)))
```
qbarrand, line 424: I think we should use log fields instead of `fmt.Sprintf`.

internal/controllers/module_nmc_reconciler.go
```diff
@@ -402,6 +408,37 @@ func (mnrh *moduleNMCReconcilerHelper) removeModuleFromNMC(ctx context.Context,
 	return nil
 }
 
+func (mnrh *moduleNMCReconcilerHelper) moduleUpdateWorkerPodsStatus(ctx context.Context, mod *kmmv1beta1.Module, targetedNodes []v1.Node) error {
```
qbarrand, line 411: Do you think this function could use the NMC labels `module-configured` and `module-in-use`? Would it make things simpler maybe?

qbarrand:
MGMT-16034

internal/controllers/module_nmc_reconciler.go
yevgeny-shnaidman: This is a part of Warning message, that is how we use it throughout our code for better visibility

internal/controllers/module_nmc_reconciler.go
yevgeny-shnaidman: those labels won't help since they don't signify if the kernel module has been already loaded or not. We could try using ready labels on the nodes, but i thought it is more reliable to be dependent on NMC

PR #646, opened 2023-11-23
Extract symlinks from the module loader images

ivatet-amd:
Extract symlinks in the ImagePuller implementation for modprobe to load the in-tree kernel module dependencies in the worker pod.

Also, add a symlink in the test data archive with "tar czf" and update the corresponding unit test to check if the symlink is extracted and pointing to the correct file from the archive.

---

Unit-tested:
```
$ make unit-test
```

Tested in a cluster with a cherry-pick on top of the downstream "release-2.0":
```
$ make workerimage-build WORKER_IMG=<img>
$ docker push <img>
$ make deploy WORKER_IMG=<img>
```

Then, I created a Module CR with an image with a symlink in it:
```
RUN ln -s /lib/modules/$KERNEL_FULL_VERSION/kernel /opt/lib/modules/$KERNEL_FULL_VERSION/kernel
```

After creating it, I checked the `lsmod` in the worker node and verified that the worker pod had loaded the dependency in-tree module.


ivatet-amd:
Also, I'm planning to add a unit test: add a symlink in "archive.tar" and update the "imagePuller_PullAndExtract" test to verify the symlink is extracted.

ivatet-amd:
Thank you, updated it with some more unit testing.

Not sure how to address and possibly suppress the CodeQL failure since we do want the symlinks to point outside the archive root.

qbarrand:
> since we do want the symlinks to point outside the archive root.

@ivatet-amd I was wondering if we could instead have the worker app create a link from `/path/to/extracted/image/lib/modules` ➡️ `/lib/modules`, which is in turn bind mounted from the host. Doing this should still work and would make sure we do not break link by linking a file from the image to an arbitrary location in the pod's filesystem.

qbarrand:
Implemented in #646.

ivatet-amd:
Thanks @qbarrand, I believe it was https://github.com/kubernetes-sigs/kernel-module-management/pull/651/. 👍

qbarrand:
🤦🏻‍♂️

PR #651, opened 2023-11-30
Extract links from kmod images

qbarrand:
Create a link from `/path/to/extract/image/lib/modules` to `/lib/modules` in the worker Pod after extracting the kmod images to enable dependencies on in-tree kmods.

/cc @yevgeny-shnaidman


qbarrand:
Fixes #646

PR #672, opened 2023-12-19
Making GetModuleReady/GetDevicePluginReadyLabels functions public

yevgeny-shnaidman:
since "ready" labels are public APIs of KMM, operators using KMM may want to schedule they workloads based on the presence of those labels.

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -1,0 +1,13 @@
+package labels
+
+import (
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
+)
+
+func GetKernelModuleReadyNodeLabel(namespace, moduleName string) string {
+	return utils.GetKernelModuleReadyNodeLabel(namespace, moduleName)
+}
+
+func GetDevicePluginNodeLabel(namespace, moduleName string) string {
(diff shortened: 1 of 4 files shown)


pkg/labels/labels.go
```diff
@@ -0,0 +1,13 @@
...
+	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
+)
+
+func GetKernelModuleReadyNodeLabel(namespace, moduleName string) string {
+	return utils.GetKernelModuleReadyNodeLabel(namespace, moduleName)
+}
+
+func GetDevicePluginNodeLabel(namespace, moduleName string) string {
+	return utils.GetDevicePluginNodeLabel(namespace, moduleName)
+}
```
qbarrand, line 13: Could we move the existing functions instead of defining new ones?

yevgeny-shnaidman: I would prefer to keep it like that: currently we have all the code defining the labels in the internal/utils/kmmlabels.go which is convinient. If we move public labels to pkg/label it would work but it will split the code and we will have 2 package handling the labels. For now i would prefer to keep the things as is. We can get back to this discussion in the nig refactoring task that we have scheduled for 2.1

PR #679, opened 2023-12-28
Fixing DevicePlugin upgrade from v1.x to v2.x

yevgeny-shnaidman:
In v1.x we had 2 type of Daemonsets: DevicePlugin and ModuleLoader, which could be distingished by the DaemonsetRole label In v2.x, there only 1 type of daemonset, DevicePlugin, so the DaemonsetRole labels was removed.
During DevicePlugin reconciliation, reconcile get all the device plugin Ds based only on module name label, which means that in case there is a v1.1 ModuleLoader DS running it will also be chosen. Once reconciler tries to reconcile ModuleLoader DS, it will fail
This PR, updates the function that chooses the DevicePlugin DS, by removing old ModuleLoader Ds based on Role label

### DIFF
--- a/internal/controllers/device_plugin_reconciler.go
+++ b/internal/controllers/device_plugin_reconciler.go
@@ -169,7 +169,16 @@ func (dprh *devicePluginReconcilerHelper) getModuleDevicePluginDaemonSets(ctx co
 	if err := dprh.client.List(ctx, &dsList, opts...); err != nil {
 		return nil, fmt.Errorf("could not list DaemonSets: %v", err)
 	}
-	return dsList.Items, nil
+
+	devicePluginsList := make([]appsv1.DaemonSet, 0, len(dsList.Items))
+	// remove the older version module loader daemonsets
+	for _, ds := range dsList.Items {
+		if ds.GetLabels()[constants.DaemonSetRole] != constants.ModuleLoaderRoleLabelValue {
+			devicePluginsList = append(devicePluginsList, ds)
+		}
+	}
+
+	return devicePluginsList, nil
 }
 
 func (dprh *devicePluginReconcilerHelper) handleDevicePlugin(ctx context.Context, mod *kmmv1beta1.Module, existingDevicePluginDS []appsv1.DaemonSet) error {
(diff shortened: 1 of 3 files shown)


internal/controllers/device_plugin_reconciler.go
```diff
@@ -169,7 +169,16 @@ func (dprh *devicePluginReconcilerHelper) getModuleDevicePluginDaemonSets(ctx co
...
 	}
-	return dsList.Items, nil
+
+	devicePluginsList := make([]appsv1.DaemonSet, 0, len(dsList.Items))
+	// remove the older version module loader daemonsets
+	for _, ds := range dsList.Items {
+		if ds.GetLabels()[constants.DaemonSetRole] != constants.ModuleLoaderRoleLabelValue {
+			devicePluginsList = append(devicePluginsList, ds)
+		}
+	}
```
ybettan, line 179: Instead of filtering out all `module-loader` DS. How about we label the `device-plugin` role and use 
```
client.MatchingLabels(map[string]string{
    constants.ModuleNameLabel: name,
    constants.DaemonSetRole: DevicePluginRoleLabelValue,
})
```
in `List.opts` ?

yevgeny-shnaidman: basically, we don't need that label, since starting from KMM V2, the only existing daemonset is device-plugin. I don't want to add a label just for V1 support, which will eventually become obsolete

ybettan: Aren't you adding a label anyway?
https://github.com/kubernetes-sigs/kernel-module-management/pull/679/files#diff-02820ac94e512b02b1879746ff901479469e1dda3c21277d7f7a170cd1c25c71R10

ybettan: I see that we had `DevicePluginRoleLabelValue = "device-plugin"` in 1.1. Can we add this to `opt` for the kubeAPI instead of filtering `ModuleLoaderRoleLabelValue = "module-loader"` after we get the list?

ybettan: After discussing, my suggestion will not fetch the v2 DS. I was first wishing to add a `client.NotMatchingLabels(map[string]string{constants.ModuleNameLabel: name}),` to `opts` but there is no such function - that led me to try to filter only device-plugins with the old label but them we miss the new DS of V2.

PR #680, opened 2024-01-03
Removing `ci/kmm-kmod/Dockerfile` from the CI files.

ybettan:
We are building the driver container using `ci/kmm-kmod-dockerfile.yaml` and `ci/kmm-kmod/Dockerfile` isn't being used and is redundant.

---

/cc @qbarrand


qbarrand:
/hold
@would there be a way to keep `Dockerfile` and use kustomize to [generate the ConfigMap dynamically](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/configGeneration.md)?
For my own local testing, for instance, I do rely on `ci/kmm-kmod` containing `Dockerfile`.

PR #686, opened 2024-01-07
Create a helper interface for mounting OCI images

yevgeny-shnaidman:
worker currently has a framework to support different methods of images to mount: from remote repo, from local tar file etc. All those images must be i the format of the OCI image, so this PR creates the interface that implements the mounting of the OCI image. It will be used by the other image mounter to mount the OCI image that was obtained from their line of image


internal/worker/remoteimagemounter_test.go
```diff
@@ -147,33 +143,7 @@ var _ = Describe("imageMounter_MountImage", func() {
...
-			).To(
-				Equal("a"),
-			)
-
-			Expect(
-				sameFiles(filepath.Join(imgRoot, "link"), filepath.Join(imgRoot, "a")),
-			).To(
-				BeTrue(),
-			)
-
```
qbarrand, line 174: Why remove this?

PR #687, opened 2024-01-09
Fix image name validation

qbarrand:
Fix a bug that would validate a registry hostname with port as a qualified image name with tag.
Make registry hostname and identifier (tag or digest) mandatory in `containerImage` fields of the Module CR.

/cc @mresvanis @ybettan @yevgeny-shnaidman


api/v1beta1/module_webhook_test.go
```diff
@@ -40,6 +40,8 @@ func getLengthAfterSlash(s string) int {
 	return len(before)
 }
 
+const validImageName = "example.com/image:tag"
```
ybettan, line 43:
```suggestion
const validImageName = "example.com/user/image:tag"
```

ybettan: Not crucial if you don't like it.

qbarrand:
Will postpone this change to v2.1.
/hold

PR #716, opened 2024-01-25
Adding an operator-upgrade Github workflow.

ybettan:
This test will ensure the bundle built in the PR can be used to upgrade from the last bundle available of KMM.

### DIFF
--- a/ci/prow/operator-upgrade
+++ b/ci/prow/operator-upgrade
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# operator-sdk run bundle<-upgrade> doesn't support local images and the image
+# has to be pushed to a remote registry.
+# When the command is run, operator-sdk checks the image twice. At first, it
+# validates that the image exists and then it creates a pod in the cluster
+# deploying the image, therefore, the image has to be accessed from both, the
+# running host and the minikube cluster.
+sudo echo "127.0.0.1 host.minikube.internal" | sudo tee -a /etc/hosts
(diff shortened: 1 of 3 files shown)


ybettan:
/cc @qbarrand 
It has passed 5/5 of the tries I did.

I have kept the explicit namespaces collection since although the default behavior [documented](https://troubleshoot.sh/docs/collect/cluster-resources/#namespaces-optional) is to collect data from all namespace I still got partial data and some of the objects were empty.

ci/prow/operator-upgrade
```diff
@@ -0,0 +1,50 @@
...
+# Install OLM
+make operator-sdk
+./bin/operator-sdk olm install
+
+# It takes some time to the operatorhubio-catalog catalogsource to get to a
+# READY connection state, we need this to validate that the catalog is ready
+# before installing a bundle, otherwise we might get a race condition that will
+# resolved in the Subscription not being reconciled after the connecting is
+# ready
+timeout 3m bash -c 'until kubectl -n olm get catalogsource/operatorhubio-catalog -o yaml | yq ".status.connectionState.lastObservedState == \"READY\"" | grep true; do sleep 5; done'
```
qbarrand, line 31:
```suggestion
timeout 3m bash -c 'until [ "$(kubectl -n olm get catalogsource/operatorhubio-catalog -o jsonpath={.status.connectionState.lastObservedState})" = "READY" ]; do sleep 5; done'
```

I think this is slightly cleaner. Then the rest lgtm.

PR #719, opened 2024-01-30
Add a webhook for namespace deletion

### LINKED ISSUE #708: Handle namespace deletion and Module finalization
When a user deletes a Module, KMM tries to unload the kmod on all nodes.
If the user deletes a namespace that contains Modules, KMM is unable to create unloading Pods. This situation creates a deadlock because KMM does not clear the finalizer on Module(s) resources, because it cannot unload it.
We want to make that situation recoverable by preventing namespace deletion when the namespace contains at least one Module.

qbarrand:
This commit makes KMM set the `kmm.node.k8s.io/contains-modules` label on all namespaces that contain at least one `Module`.
It also adds a new webhook to the bundle and the corresponding handler in the manager.
The new webhook rejects namespace deletions if the `kmm.node.k8s.io/contains-modules` label is present on the namespace. This avoids entering situations where the namespace is being deleted and KMM cannot create unloading Pods to honor `Module` deletion.

Fixes #708

/cc @mresvanis @yevgeny-shnaidman @ybettan


internal/webhook/namespace_deletion.go
```diff
@@ -0,0 +1,46 @@
+package webhook
```
yevgeny-shnaidman, line 1: why is this webhook in internal , while the admission webhook is in api package? Seems a little bit strange: eventually they are both webhooks, each one doing their own safeguarding

qbarrand: An upcoming PR will move the existing webhook code under `internal/webhook`. The `Module` webhook is under `api` because that is where `kubebuilder` generated it, but it is undesirable for many reasons - one of them is that people importing the `api` package now also import container-runtime.

internal/controllers/module_nmc_reconciler.go
```diff
@@ -439,6 +450,82 @@ func (mnrh *moduleNMCReconcilerHelper) moduleUpdateWorkerPodsStatus(ctx context.
 	return mnrh.client.Status().Patch(ctx, mod, client.MergeFrom(unmodifiedMod))
 }
 
+type namespaceHelper interface {
```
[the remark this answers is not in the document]
qbarrand: Not sure about this. It feels weird to have a big interface with a lot of unrelated methods. When creating the helper in test cases, we need to provide a lot of `nil` dependencies, unrelated to the features brought by these new methods.
@mresvanis WDYT? should we put those two new methods into the existing interface or put them in a separate one?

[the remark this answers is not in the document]
yevgeny-shnaidman: helper interface is almost always a lot of unrelated methods, and we are using it for the sole purpose of simplifying unit-test. If we start breaking helper interfaces by the related functions will will end up with a LOT of helper interfaces. And we are creating the tested interfaces only once per all the tests of a method, so it should not be a big issue to provide a nil whenever it is needed ( just summarizing some points for @mresvanis )

docs/mkdocs/documentation/deploy_kmod.md
```diff
@@ -236,11 +236,10 @@ KMM will then create worker Pods where required to run `modprobe -r` and unload
...
     allow privileged KMM worker Pods to run.
     It also includes any pull secret referenced in `.spec.imageRepoSecret`.  
-    To avoid situations where KMM is unable to unload the kernel module from nodes:  
-
-      - do not delete those resources while the `Module` resource is still present in the cluster in any state,
-        including `Terminating`;
-      - do not delete any namespace containing at least a `Module` resource.
+    To avoid situations where KMM is unable to unload the kernel module from nodes, make sure those resources are not
+    deleted while the `Module` resource is still present in the cluster in any state, including `Terminating`.  
+    KMM ships with an admission webhook that rejects the deletion of namespaces that contain at least one `Module`
```
ybettan, line 241:
```suggestion
    KMM ships with a validating admission webhook that rejects the deletion of namespaces that contain at least one `Module`
```

internal/controllers/module_nmc_reconciler.go
```diff
@@ -78,13 +81,21 @@ func (mnr *ModuleNMCReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	}
 	if mod.GetDeletionTimestamp() != nil {
 		//Module is being deleted
+		if err = mnr.nsHelper.tryRemovingLabel(ctx, mod.Namespace, mod.Name); err != nil {
```
ybettan, line 84: Shouldn't we remove the label only after finalizing the module?

internal/controllers/module_nmc_reconciler.go
```diff
@@ -439,6 +450,80 @@ func (mnrh *moduleNMCReconcilerHelper) moduleUpdateWorkerPodsStatus(ctx context.
...
+	modList := kmmv1beta1.ModuleList{}
+
+	opt := client.InNamespace(name)
+
+	if err := h.client.List(ctx, &modList, opt); err != nil {
+		return fmt.Errorf("could not list modules in namespace %s: %v", name, err)
+	}
+
+	if count := len(modList.Items); count > 1 {
+		logger.Info("Namespace still contains modules; not removing the label", "count", count)
```
ybettan, line 499: Aren't we going to spam the lobs until the user removes the module?

yevgeny-shnaidman: since this log will be issued only during module deletion, and it is supposed to be short-time occurrence, i think it should be OK. This log will be very helpful in case of some weird behavior that we will need to debug

internal/controllers/module_nmc_reconciler.go
qbarrand: No, otherwise we may remove the finalizer but fail to remove the label for any reason. In that we will never try to remove the label again.

yevgeny-shnaidman: @ybettan we are just trying to prevent namespace deletion, we are not handling the whole SA/RoleBindings etc'. There may a small chance for race condition, but it is better then none

internal/controllers/module_nmc_reconciler.go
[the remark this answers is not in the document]
mresvanis: I will share first my view and then I  suggest to go with whatever makes more sense as a step towards that direction (if we think that direction is what actually makes sense). 

#### Naming `util`, `common`, `helper`, `misc`

IMHO interfaces, pkgs, files etc, named `util`, `common`, `helper`, `misc` can be [avoided](https://go.dev/blog/package-names) easily by keeping those interfaces, pkgs, files focused on what they are purposed for exactly. This way, one might end up with a lot of individual interfaces/pkgs/files, but if they are self-describing and focused, it's a lot easier to use them, edit them, mock them etc. 

#### Helpers in KMM
I understand and support what we did with the helper interfaces and why we did it, it has worked quite well for the codebase and its development so far, which has mostly to do with being able to efficiently unit test the whole codebase. But I think it has started to get out of hand, with some interfaces having too many unrelated methods. While developing and debugging, the cognitive load is quite big and it's getting harder to remember where to search for what. Also, while reading/changing the unit tests, it's getting harder to decide on how many `nil`s I need and for what exact purpose (although not impossible or absolutely huge issue).  

#### Suggestion

I would suggest that we make an effort to gradually refactor the helper interfaces to smaller, but narrowly focused ones and ideally loose the `helper` suffix altogether. In this specific case, although @qbarrand has created an additional interface (which goes towards the direction I would choose), I would rethink its naming and make it as focused as possible, e.g. `namespaceLabeler`. 

> PS. I get that it won't be that easy to split all those helpers and name/refocuse them clearly, but I'm quite positive that it will definitely lead to a more maintainable codebase.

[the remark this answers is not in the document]
qbarrand: Thanks @mresvanis. I'd rather change the name to `namespaceLabeler` than move that code back to the big interface indeed. @yevgeny-shnaidman are you OK with that?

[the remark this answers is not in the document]
yevgeny-shnaidman: No, i am not ok :). But lets push it,  we are going to have a long serious discussion on all the refactoring issues anyway, we can discuss that issue there also

docs/mkdocs/documentation/deploy_kmod.md
qbarrand: Fixed, thanks!

qbarrand:
@mresvanis @yevgeny-shnaidman @ybettan PTAL 🙂

PR #727, opened 2024-02-06
Move webhook code to internal/webhook

qbarrand:
Remove webhook code from the `api{,-hub}` packages to avoid depending on controller-runtime there.
Implement the new `admission.CustomValidator` interface as the old `webhook.Validator` is now deprecated.
Bump controller-gen to v0.14.0.

/cc @mresvanis @yevgeny-shnaidman


internal/webhook/module.go
```diff
@@ -14,72 +14,87 @@ See the License for the specific language governing permissions and
...
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
 // maxCombinedLength is the maximum combined length of Module name and namespace when the version field is set.
 const maxCombinedLength = 40
 
-// log is for logging in this package.
-var modulelog = logf.Log.WithName("module-resource")
+type Module struct {
```
yevgeny-shnaidman, line 36: Maybe we can rename it to ModuleValidation? This way it would be less confusing to read the code, no collision with original Module in our heads :)

qbarrand: That works for me.

internal/webhook/hub/managedclustermodule_webhook.go
```diff
@@ -0,0 +1,87 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+type ManagedClusterModuleValidator struct {
+	logger logr.Logger
+	m      admission.CustomValidator
+}
+
+func NewManagedClusterModule(logger logr.Logger) *ManagedClusterModuleValidator {
```
yevgeny-shnaidman, line 37: Lets change the New... function name also to NewManagedClusterModuleValidator

internal/webhook/module.go
```diff
@@ -14,72 +14,87 @@ See the License for the specific language governing permissions and
...
 // maxCombinedLength is the maximum combined length of Module name and namespace when the version field is set.
 const maxCombinedLength = 40
 
-// log is for logging in this package.
-var modulelog = logf.Log.WithName("module-resource")
+type ModuleValidator struct {
+	logger logr.Logger
+}
+
+func NewModule(logger logr.Logger) *ModuleValidator {
```
yevgeny-shnaidman, line 40: lets change the New function to NewModuleValidator

PR #731, opened 2024-02-07
Move to Go 1.22

qbarrand:
Bump Kubernetes dependencies to v1.30.0.
Bump controller-gen to v0.15.0.

/cc @mresvanis @yevgeny-shnaidman


yevgeny-shnaidman:
why do we not update go.mod?

qbarrand:
> why do we not update go.mod?

If we do that, we require that people importing the API also use 1.22.

qbarrand:
> then it is ok that we require him to update his Go version

Why is that OK?

qbarrand:
> User cannot expect us to always provide the same golang version for all KMM packages.

We are not making that promise, but at the same time I don't think we should be forcing users to upgrade for no benefit for us. Go tries very hard to be [backwards compatible](https://go.dev/doc/go1compat), so Go 1.22 will happily compile code written for 1.19.

yevgeny-shnaidman:
> > User cannot expect us to always provide the same golang version for all KMM packages.
> 
> We are not making that promise, but at the same time I don't think we should be forcing users to upgrade for no benefit for us. Go tries very hard to be [backwards compatible](https://go.dev/doc/go1compat), so Go 1.22 will happily compile code written for 1.19.

My main issue is that we create an image using Go 1.22 , while unit-test and go.mod are 1.21. It is pretty disconcerting, What do we get now by moving go 1.22?  Maybe we can stay on 1.21 and only move to 1.22 when it becomes really beneficiary?

qbarrand:
/hold
Waiting for a new controller-runtime release that supports Kubernetes 1.30 and Go 1.22.

PR #743, opened 2024-02-19
Move webhooks to a separate Deployment

qbarrand:
Adapt deployment files and tests.


Makefile
```diff
@@ -202,12 +208,14 @@ KUSTOMIZE_CONFIG_HUB_DEFAULT ?= config/default-hub
 deploy: deploy-cert-manager manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG) worker=$(WORKER_IMG)
 	cd config/manager-base && $(KUSTOMIZE) edit set image signer=$(SIGNER_IMG)
+	cd config/webhook-server && $(KUSTOMIZE) edit set image webhook-server=$(WEBHOOK_IMG)
```
yevgeny-shnaidman, line 211: I am wondering: we have a different deployment for webhooks, maybe we should also make a deployment target that does not deploy webhooks? We use "deploy" pretty heavily in our testing, but we don't really need webhooks for testing. WDYT?

qbarrand: Well I am happy to add `deploy-no-webhooks` for instance.

yevgeny-shnaidman: maybe deploy-with-webhooks?  Or will it require changing the prow configuration?

qbarrand: I'd rather not break existing behaviour.

yevgeny-shnaidman: ok, then lets add deploy-no-webhook

PR #747, opened 2024-02-25
changing inTreeRemoval to support multiple modules

yevgeny-shnaidman:
This PR contains the following changes
    1) Adding new field InTreeModulesToRemove to the Module that is of type string array
    2) Marking the old field InTreeModuleToRemove as deprecated
    3) adding validation to webhook that InTreeModuleToRemove and InTreeModulesToRemove
       can not be both set in the Module
    4) appropriate changes are done in the MLD structures and in the worker
       configuration
    5) for older version Modules that already use InTreeModuleToRemove
       field, its value is converted into string array in ModuleLoaderData
    6) adding InTreeModulesToRemove to the ModuleConfig of the NMC
    7) in case NMC was already set with InTreeModuleToRemove prior to update
       of the KMM, worker will use InTreeModuleToRemoce in case InTreeModuleToRemove
       is not set
    8) code and unit-tests are updated


qbarrand:
@yevgeny-shnaidman I think we want a [conversion webhook](https://book.kubebuilder.io/multiversion-tutorial/conversion) for this CRD change?
I think we also want to bump the CRD version (e.g. `v1beta2`).
/hold

yevgeny-shnaidman:
> @yevgeny-shnaidman I think we want a [conversion webhook](https://book.kubebuilder.io/multiversion-tutorial/conversion) for this CRD change? I think we also want to bump the CRD version (e.g. `v1beta2`). /hold

I am not sure if we currently have a customer that uses that field. I think that maybe Intel, but they are waiting for this change, so they are probably prepared to just change their operator/yamls. 
Regarding version update - i have not specific opinion

api/v1beta1/nodemodulesconfig_types.go
```diff
@@ -27,8 +27,8 @@ type ModuleConfig struct {
 	// When InsecurePull is true, the container image can be pulled without TLS.
 	InsecurePull bool `json:"insecurePull"`
 	//+optional
-	InTreeModuleToRemove string       `json:"inTreeModuleToRemove,omitempty"`
-	Modprobe             ModprobeSpec `json:"modprobe"`
+	InTreeModulesToRemove []string     `json:"inTreeModulesToRemove,omitempty"`
```
qbarrand, line 30: Wondering how that will work during upgrade. Maybe we need a conversion webhook for NMC actually.

api/v1beta1/module_types.go
```diff
@@ -118,10 +118,16 @@ type KernelMapping struct {
 	// Regexp is a regular expression to be match against node kernels.
 	Regexp string `json:"regexp"`
 
+	// DEPRECATED.
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

api/v1beta1/module_types.go
```diff
@@ -219,10 +225,16 @@ type ModuleLoaderContainerSpec struct {
 	// RegistryTLS set the TLS configs for accessing the registry of the module-loader's image.
 	RegistryTLS TLSOptions `json:"registryTLS"`
 
+	// DEPRECATED.
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

api/v1beta1/nodemodulesconfig_types.go
yevgeny-shnaidman: I think it should not be an issue: InTreeRemoval is optional so when the CRD is updated, NMC is created/updated or worker tried to load NMC it will do it without a problem. what we need to do is maybe add a code to the worker to use InTreeModule instead of InTreeModules in case InTreeModule is not empty, but InTreeModules is.

internal/module/kernelmapper.go
```diff
@@ -115,9 +117,22 @@ func (kh *kernelMapperHelper) prepareModuleLoaderData(mapping *kmmv1beta1.Kernel
...
 
-	mld.InTreeModuleToRemove = mod.Spec.ModuleLoader.Container.InTreeModuleToRemove
-	if mapping.InTreeModuleToRemove != "" {
-		mld.InTreeModuleToRemove = mapping.InTreeModuleToRemove
+	mld.InTreeModulesToRemove = mod.Spec.ModuleLoader.Container.InTreeModulesToRemove
+	if mapping.InTreeModulesToRemove != nil {
+		mld.InTreeModulesToRemove = mapping.InTreeModulesToRemove
+	}
+
+	// [TODO] remove this code after InTreeModuleToRemove deplricated field has been
```
qbarrand, line 125:
```suggestion
	// [TODO] remove this code after InTreeModuleToRemove deprecated field has been
```

internal/module/kernelmapper.go
```diff
@@ -1,3 +1,5 @@
+//nolint:SA1019
```
qbarrand, line 1: Do we need both this and `//nolint`?

internal/module/kernelmapper.go
```diff
@@ -115,9 +117,22 @@ func (kh *kernelMapperHelper) prepareModuleLoaderData(mapping *kmmv1beta1.Kernel
...
+
+	// [TODO] remove this code after InTreeModuleToRemove deplricated field has been
+	// removed from the CRD
+	inTreeModuleToRemove := mod.Spec.ModuleLoader.Container.InTreeModuleToRemove //nolint:staticcheck
+	if mapping.InTreeModuleToRemove != "" {                                      //nolint:staticcheck
+		inTreeModuleToRemove = mapping.InTreeModuleToRemove //nolint:staticcheck
+	}
+
+	// webhook makes sure that InTreeModuleToRemove and InTreeModulesToRemove cannot both
+	// be set, so if inTreeModuleToRemove is nt empty, its value should be set into mld.InTreeModulesToRemove
```
qbarrand, line 133:
```suggestion
	// be set, so if inTreeModuleToRemove is not empty, its value should be set into mld.InTreeModulesToRemove
```

internal/module/kernelmapper.go
yevgeny-shnaidman: fixed

internal/module/kernelmapper.go
yevgeny-shnaidman: removed

internal/module/kernelmapper.go
yevgeny-shnaidman: fixed

PR #751, opened 2024-02-28
Require Go 1.21

qbarrand:
`k8s.io/api` requires 1.21, so bump our version as well.


qbarrand:
Requires https://github.com/kubernetes/test-infra/pull/32093 to be merged

Issue #756, opened 2024-03-07
Install the latest available bundle from the `operatorhubio-catalog` in the cluster instead of building it.


ybettan:
Instead of building the latest bundle from the latest available images u/s we should use the latest bundle image available in the `operatorhubio-catalog` service in the `olm` namesapce.

Following up on https://github.com/kubernetes-sigs/kernel-module-management/issues/478

PR #759, opened 2024-03-11
Getting the last published bundles from OLM instead of building it.

ybettan:
The latest published bundle for our cluster is available in the `operatorhubio-catalog` service in the `olm` namespace, therefore, there is no need to rebuild it.

---

Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/756


ybettan:
/hold
This need some fixes.

PR #760, opened 2024-03-11
Handle `Module` namespace in `PreflightValidation`

qbarrand:
Make `PreflightValidation` objects aware of `Module` namespaces to avoid cross-namespace collisions.
Add `PreflightValidation` `v1beta2` to restructure its `status`.
Add a conversion webhook for `PreflightValidation` `v1beta1` --> `v1beta2`.
Move the code updating the `PreflightValidation` status from `internal/statusupdater` to `internal/preflight` to avoid a dependency cycle.
Update reconciler, libraries and unit tests.

/cc @yevgeny-shnaidman @mresvanis @ybettan


config/manager-base/kustomization.yaml
```diff
@@ -6,8 +6,7 @@ resources:
 
 images:
 - name: signer
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-signimage
-  newTag: latest
+  newName: host.minikube.internal:5000/kmm-signimage
```
yevgeny-shnaidman, line 9: This should probably be reverted?

config/manager/kustomization.yaml
```diff
@@ -9,11 +9,9 @@ patches:
 
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: host.minikube.internal:5000/kmm
```
yevgeny-shnaidman, line 12: and those also?

config/webhook-server/kustomization.yaml
```diff
@@ -9,5 +9,4 @@ commonLabels:
 
 images:
 - name: webhook-server
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-webhook-server
-  newTag: latest
+  newName: host.minikube.internal:5000/kmm-webhook-server
```
yevgeny-shnaidman, line 12: and this one

internal/controllers/preflightvalidation_reconciler.go
```diff
@@ -150,67 +151,78 @@ func (r *PreflightValidationReconciler) getModulesToCheck(ctx context.Context, p
...
-	return r.statusUpdater.PreflightPresetStatuses(ctx, pv, existingModulesName, newModulesNames)
+	return r.statusUpdater.PresetStatuses(ctx, pv, existingModulesName, newModulesNames)
 }
 
-func (r *PreflightValidationReconciler) checkPreflightCompletion(ctx context.Context, name, namespace string) (bool, error) {
-	pv := v1beta12.PreflightValidation{}
-	err := r.client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, &pv)
+func (r *PreflightValidationReconciler) checkPreflightCompletion(ctx context.Context, name string) (bool, error) {
+	pv := v1beta2.PreflightValidation{}
+	err := r.client.Get(ctx, types.NamespacedName{Name: name}, &pv)
```
yevgeny-shnaidman, line 218: why are we not using namespace to search for Preflights?

internal/preflight/preflight.go
```diff
@@ -195,3 +195,36 @@ func (p *preflightHelper) verifySign(ctx context.Context, pv *kmmv1beta1.Preflig
 	}
 	return false, "Waiting for sign verification"
 }
+
+func DeleteModuleStatus(statuses *[]v1beta2.PreflightValidationModuleStatus, nsn types.NamespacedName) {
```
yevgeny-shnaidman, line 199: I see that thisfunction are using only in tests  -what am i missing?

internal/preflight/preflight.go
```diff
@@ -195,3 +195,36 @@ func (p *preflightHelper) verifySign(ctx context.Context, pv *kmmv1beta1.Preflig
...
+	})
+
+	if i != -1 {
+		return &statuses[i], true
+	}
+
+	return nil, false
+}
+
+func UpsertModuleStatus(statuses *[]v1beta2.PreflightValidationModuleStatus, s v1beta2.PreflightValidationModuleStatus) error {
```
yevgeny-shnaidman, line 217: The same for this one: if it is only used in test, it should be defined in the test file

api/v1beta1/preflightvalidation_types.go
```diff
@@ -100,6 +71,58 @@ type PreflightValidation struct {
...
+			}
+
+			dst.Status.Modules = append(dst.Status.Modules, status)
+		}
+	}
+
+	return nil
+}
+
+func (p *PreflightValidation) ConvertFrom(srcRaw conversion.Hub) error {
```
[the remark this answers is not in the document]
qbarrand: The storage version of the CRD (the schema it uses in etcd) is now `v1beta2`, so if a client requests the object in `v1beta1` format then this function will be called.
This method must be implemented to satisfy the `Convertible` interface which is required for the webhook to work.

internal/preflight/preflight.go
qbarrand: It is only used in the tests, but in two different packages, so it cannot be in a `_test.go` file. Do you think it should go somewhere else?

internal/preflight/preflight.go
qbarrand: It is only used in the tests, but in two different packages, so it cannot be in a _test.go file. Do you think it should go somewhere else?

config/manager-base/kustomization.yaml
qbarrand: Thanks! Fixed.

config/manager/kustomization.yaml
qbarrand: Thanks! Fixed.

internal/preflight/preflight.go
yevgeny-shnaidman: yes, i think it is not a good idea to have a test function in a production code.  IF it is used in 2 packages, we can define it in one package, but in the test file, and use it from there, IMHO

internal/preflight/preflight.go
yevgeny-shnaidman: same answer as above

api/v1beta1/preflightvalidation_types.go
```diff
@@ -100,6 +71,58 @@ type PreflightValidation struct {
...
+
+	if count := len(p.Status.CRStatuses); count > 0 {
+		dst.Status.Modules = make([]v1beta2.PreflightValidationModuleStatus, 0, count)
+
+		for k, v := range p.Status.CRStatuses {
+			namespace, name, ok := strings.Cut(k, "/")
+
+			if !ok || v == nil {
+				// Elements whose key is not a namespace name or that are nil are invalid.
+				return nil
```
ybettan, line 90: If they are invalid, shouldn't we return an error?

api/v1beta1/preflightvalidation_types.go
```diff
@@ -100,6 +71,58 @@ type PreflightValidation struct {
...
+	p.ObjectMeta = src.ObjectMeta
+	p.Spec = src.Spec
+
+	p.Status = PreflightValidationStatus{}
+
+	if count := len(src.Status.Modules); count > 0 {
+		p.Status.CRStatuses = make(map[string]*CRStatus, count)
+
+		for _, v := range src.Status.Modules {
+			v := v
```
ybettan, line 118: Why are we shadowing `v` here with a new variable named `v1`?

api/v1beta1/preflightvalidation_types.go
qbarrand: The function signature comes from the `Convertible` interface that we need to implement; we cannot change it.
The situation is that none of the current `v1beta1` `PreflightValidation` resources include a namespace. So all of them are technically invalid.
By skipping all invalid elements we can proceed with the conversion. The new `v1beta2` object can then be reconciled with the updated `PreflightValidation` reconciler that handles namespaces. Now statuses will correctly include the namespace.

api/v1beta1/preflightvalidation_types.go
qbarrand: Not sure I get the question, there is no `v1` here.
We need to do `v := v` because the loop variable is a pointer. If we did not have that line, all elements in the map would point to the same struct. Go [changes this behaviour in 1.22](https://go.dev/blog/loopvar-preview), but in the meantime we still need to do this.

ybettan: I don't think it would. I think that you are referring to [closures](https://gobyexample.com/closures) that are evaluated, apparently, only when called.

In this case, this is a map (not a function/goroutine capturing a closure), therefore, I believe it will work without `v := v` (which is very confusing).

See [this playground](https://go.dev/play/p/S_AGdmKkPpN?v=goprev)

qbarrand: I don't think the playground reproduces what we are doing here. [That one](https://go.dev/play/p/H881vVkZZGX?v=goprev) exhibits the problem.

internal/preflight/preflight.go
qbarrand: Done! PTAL.

api/v1beta1/preflightvalidation_types.go
ybettan: Ohh OK, I see what the issue is, it is different than the issue I remembered. Here `v` in the loop is the same variable through the whole loop so each iteration actually override all prior iterations since it is using the same variable.

I remember that the "go to" fix was indeed to do `v := v` - I am very happy they fixed it in 1.22!

PR #775, opened 2024-03-28
`PreflightValidation`: populate old status keys

qbarrand:
Also populate old-style, un-namespaced status keys to remain compatible with clients designed for pre-2.1 KMM.


api/v1beta1/preflightvalidation_types.go
```diff
@@ -117,6 +117,9 @@ func (p *PreflightValidation) ConvertFrom(srcRaw conversion.Hub) error {
 		for _, v := range src.Status.Modules {
 			v := v
 			p.Status.CRStatuses[v.Namespace+"/"+v.Name] = &v.CRBaseStatus
+
```
yevgeny-shnaidman, line 120: lets just add a TODO to remove it as soon as we stop supporting < 2.1

qbarrand: I think we should keep this code until we deprecate `v1beta1` entirely.

PR #824, opened 2024-05-31
Normalize kernel version in labels and image tags

qbarrand:
Handle uncommon kernel versions with characters such as `+`, by replacing them with `_`.

/cc @ybettan @mresvanis

### DIFF
--- a/internal/api/moduleloaderdata.go
+++ b/internal/api/moduleloaderdata.go
@@ -16,6 +16,11 @@ import (
 type ModuleLoaderData struct {
 	// kernel version
 	KernelVersion string
+
+	// KernelNormalizedVersion is the kernel version with some characters replaced with '_' so that it can be used in
+	// a Kubernetes label or a container image tag.
+	KernelNormalizedVersion string
+
 	// Repo secret for DS images
 	ImageRepoSecret *v1.LocalObjectReference
 
--- a/internal/kernel/kernel.go
+++ b/internal/kernel/kernel.go
@@ -1,0 +1,15 @@
+package kernel
+
+import "strings"
+
+func replaceInvalidChar(r rune) rune {
+	if (r < 'A' || r > 'Z') && (r < 'a' || r > 'z') && (r < '0' || r > '9') && r != '-' && r != '.' && r != '_' {
+		return '_'
+	}
+
+	return r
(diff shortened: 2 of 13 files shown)


internal/kernel/kernel.go
```diff
@@ -0,0 +1,15 @@
+package kernel
+
+import "strings"
+
+func replaceInvalidChar(r rune) rune {
+	if (r < 'A' || r > 'Z') && (r < 'a' || r > 'z') && (r < '0' || r > '9') && r != '-' && r != '.' && r != '_' {
+		return '_'
+	}
```
ybettan, line 8: Shouldn't we just replace `+` with `_`?

internal/api/moduleloaderdata.go
```diff
@@ -16,6 +16,11 @@ import (
 type ModuleLoaderData struct {
 	// kernel version
 	KernelVersion string
+
+	// KernelNormalizedVersion is the kernel version with some characters replaced with '_' so that it can be used in
+	// a Kubernetes label or a container image tag.
+	KernelNormalizedVersion string
```
ybettan, line 22: Do we still need `KernelVersion`?

internal/kernel/kernel.go
qbarrand: Handling any invalid character makes the code more future-proof than just handling `+`.

internal/api/moduleloaderdata.go
qbarrand: We do, in `module_nmc_reconciler.go` for instance.

yevgeny-shnaidman:
do we still need that line:
https://github.com/kubernetes-sigs/kernel-module-management/blob/582546abe7a94a82fbb9062f9c6d713da031df98/internal/controllers/module_nmc_reconciler.go#L316

qbarrand:
@yevgeny-shnaidman I think this new solution could make that line redundant, but if we remove it, we change the behaviour.
Some kernels have a trailing `+`, indicating that the source has been patched before compilation. In most cases, you can still build your kmod against the original headers for that kernel.
If we remove that line, we break the Dockerfile of users who assume that they can build their kmod for a modified kernel using the original headers for that version.

PR #826, opened 2024-06-03
Defaulting to the `latest` tag if no tag was specified.

ybettan:
Defaulting to the `latest` tag if no tag was specified.

In the current code, we handle differently build/sign and deploying a
pre-built kmod.

This is the current behavior:
* If a `build`/`sign` section is set in the module, and the image tag
  isn't specified, then we will return an error.
* If there are no `build`/`sign` section, and the image tag isn't
  specified, then we will default to the `latest` tag.

This change is adjusting the behavior of both workflow to always defaul
to the `latest` image tag if it is not specified in the
`Module`/`ManagedClusterModule`.

This is achieved by adding a new mutating-webhook for the
`Module`&`ManangedClusterModule` objects.

This change is also adding the `nullable` attributes to some API fields
in order for it to work.

---

/cc @qbarrand @yevgeny-shnaidman 

The `+nullable` kubebuilder annotation is required until this [PR](https://github.com/kubernetes/kubernetes/issues/117447#issuecomment-2127733969) is merged.


ybettan:
/hols
The webhook isn't popping the warning for some reason.

internal/webhook/module.go
```diff
@@ -141,6 +149,19 @@ func validateModuleLoaderContainerSpec(container kmmv1beta1.ModuleLoaderContaine
 			(km.InTreeModuleToRemove != "" && container.InTreeModulesToRemove != nil) { //nolint:staticcheck
 			return fmt.Errorf("only one type if field (InTreeModuleToRemove or InTreeModulesToRemove) can be defined in KenrelMapping or Container")
 		}
+
+		var imgToCheck string
+		imgToCheck = container.ContainerImage
+		if km.ContainerImage != "" {
+			imgToCheck = km.ContainerImage
+		}
+		if !strings.ContainsRune(imgToCheck, '@') && !strings.ContainsRune(imgToCheck, ':') {
```
yevgeny-shnaidman, line 158: why can't we just return ImageHashOrTagNotSet from here? Why do we have to wait for the loop to finish?

yevgeny-shnaidman, line 158: what do you think about using NewTag and NewDigest to verify the correctness of the image? Those functions are part of the go-containerregistry/pkg/name package, and they make a little bit more verifications than just the presence of @ or :

yevgeny-shnaidman: Why do we need a warning in the webhook? Container image is either valid or not.

internal/webhook/module.go
```diff
@@ -21,8 +21,10 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/go-logr/logr"
+	multierror "github.com/hashicorp/go-multierror"
```
qbarrand, line 27: I think we should use [`errors.Join`](https://pkg.go.dev/errors#Join) from the standard library instead.

internal/webhook/module.go
qbarrand: I think we should have a new mutating webhook that sets the `latest` tag instead of returning warnings.

ybettan:
/hold
This fix will fail for any field in the CR that is an optional pointer/slice and was not initiated because of [this](https://github.com/kubernetes-sigs/structured-merge-diff/pull/259) issue unless the `+nullable` kubebuilder annotation is added to it.

This is very fragile and will require an API bump for each field that we forgot to check.

@yevgeny-shnaidman @TomerNewman How about we just add a validation-webhook rule to refuse `Module`/`ManagedClusterModule` in which an image with no has or tag was specified?

We can proceed with this mutating webhook without the `+nullable` annotation once the PR fixing the issue is merged in https://github.com/kubernetes-sigs/structured-merge-diff.

ybettan:
/close
Closing in favor of https://github.com/kubernetes-sigs/kernel-module-management/pull/853

PR #850, opened 2024-06-27
Shorten UpdateNodeLabelsAndRecordEvents function in nmc_reconciler.go

TomerNewman:
## Summary
Today, the UpdateNodeLabelsAndRecordEvents function has around 100 lines of code. This commit aims to enhance its readability and maintainability by using helper functions, thereby reducing its total length. Additionally, adding unit tests that covers the new helper functions.

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -153,9 +153,15 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	if err := r.helper.GarbageCollectInUseLabels(ctx, &nmcObj); err != nil {
 		errs = append(errs, fmt.Errorf("failed to GC in-use labels for NMC %s: %v", req.NamespacedName, err))
 	}
+	node := v1.Node{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: nmcObj.Name}, &node); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get node %s: %v", nmcObj.Name, err)
+	}
 
-	if err := r.helper.UpdateNodeLabelsAndRecordEvents(ctx, &nmcObj); err != nil {
+	if loaded, unloaded, err := r.helper.UpdateNodeLabels(ctx, &nmcObj, &node); err != nil {
 		errs = append(errs, fmt.Errorf("could not update node's labels for NMC %s: %v", req.NamespacedName, err))
+	} else {
+		r.helper.RecordEvents(&node, loaded, unloaded)
 	}
 
 	return ctrl.Result{}, errors.Join(errs...)
@@ -213,7 +219,8 @@ type nmcReconcilerHelper interface {
 	ProcessUnconfiguredModuleStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, status *kmmv1beta1.NodeModuleStatus) error
 	RemovePodFinalizers(ctx context.Context, nodeName string) error
 	SyncStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
-	UpdateNodeLabelsAndRecordEvents(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
+	UpdateNodeLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) ([]types.NamespacedName, []types.NamespacedName, error)
+	RecordEvents(node *v1.Node, loadedModules, unloadedModules []types.NamespacedName)
 }
 
 type nmcReconcilerHelperImpl struct {
@@ -221,6 +228,7 @@ type nmcReconcilerHelperImpl struct {
 	pm       podManager
 	recorder record.EventRecorder
 	nodeAPI  node.Node
+	lph      labelPreparationHelper
 }
 
 func newNMCReconcilerHelper(client client.Client, pm podManager, recorder record.EventRecorder, nodeAPI node.Node) nmcReconcilerHelper {
@@ -229,6 +237,7 @@ func newNMCReconcilerHelper(client client.Client, pm podManager, recorder record
 		pm:       pm,
 		recorder: recorder,
 		nodeAPI:  nodeAPI,
+		lph:      newLabelPreparationHelper(),
 	}
 }
 
@@ -555,95 +564,102 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 	return errors.Join(errs...)
 }
 
-func (h *nmcReconcilerHelperImpl) UpdateNodeLabelsAndRecordEvents(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error {
-	node := v1.Node{}
-	if err := h.client.Get(ctx, types.NamespacedName{Name: nmc.Name}, &node); err != nil {
-		return fmt.Errorf("could not get node %s: %v", nmc.Name, err)
-	}
+type labelPreparationHelper interface {
+	getDeprecatedKernelModuleReadyLabels(node v1.Node) sets.Set[string]
+	getNodeKernelModuleReadyLabels(node v1.Node) sets.Set[types.NamespacedName]
+	getSpecLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	getStatusLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+		specLabels, statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName
+	removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+		specLabels, statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName
+}
+type labelPreparationHelperImpl struct{}
 
-	// get all the kernel module ready labels of the node
+func newLabelPreparationHelper() labelPreparationHelper {
+	return &labelPreparationHelperImpl{}
+}
+
+func (lph *labelPreparationHelperImpl) getNodeKernelModuleReadyLabels(node v1.Node) sets.Set[types.NamespacedName] {
 	nodeModuleReadyLabels := sets.New[types.NamespacedName]()
-	deprecatedNodeModuleReadyLabels := sets.New[string]()
 
 	for label := range node.GetLabels() {
 		if ok, namespace, name := utils.IsKernelModuleReadyNodeLabel(label); ok {
 			nodeModuleReadyLabels.Insert(types.NamespacedName{Namespace: namespace, Name: name})
 		}
+	}
+	return nodeModuleReadyLabels
+}
 
+func (lph *labelPreparationHelperImpl) getDeprecatedKernelModuleReadyLabels(node v1.Node) sets.Set[string] {
+	deprecatedNodeModuleReadyLabels := sets.New[string]()
+
+	for label := range node.GetLabels() {
 		if utils.IsDeprecatedKernelModuleReadyNodeLabel(label) {
 			deprecatedNodeModuleReadyLabels.Insert(label)
 		}
 	}
+	return deprecatedNodeModuleReadyLabels
+}
 
-	// get spec labels and their config
+func (lph *labelPreparationHelperImpl) getSpecLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig {
 	specLabels := make(map[types.NamespacedName]kmmv1beta1.ModuleConfig)
+
 	for _, module := range nmc.Spec.Modules {
 		specLabels[types.NamespacedName{Namespace: module.Namespace, Name: module.Name}] = module.Config
 	}
+	return specLabels
+}
 
-	// get status labels and their config
+func (lph *labelPreparationHelperImpl) getStatusLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig {
 	statusLabels := make(map[types.NamespacedName]kmmv1beta1.ModuleConfig)
+
 	for _, module := range nmc.Status.Modules {
 		label := types.NamespacedName{Namespace: module.Namespace, Name: module.Name}
 		statusLabels[label] = module.Config
 	}
+	return statusLabels
+}
 
-	unloaded := make([]types.NamespacedName, 0, len(nodeModuleReadyLabels))
-	loaded := make([]types.NamespacedName, 0, len(specLabels))
+func (h *nmcReconcilerHelperImpl) UpdateNodeLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) ([]types.NamespacedName, []types.NamespacedName, error) {
 
-	patchFrom := client.MergeFrom(node.DeepCopy())
+	// get all the kernel module ready labels of the node
+	nodeModuleReadyLabels := h.lph.getNodeKernelModuleReadyLabels(*node)
+	deprecatedNodeModuleReadyLabels := h.lph.getDeprecatedKernelModuleReadyLabels(*node)
+
+	// get spec labels and their config
+	specLabels := h.lph.getSpecLabelsAndTheirConfigs(nmc)
+
+	// get status labels and their config
+	statusLabels := h.lph.getStatusLabelsAndTheirConfigs(nmc)
 
 	// label in node but not in spec or status - should be removed
-	for nsn := range nodeModuleReadyLabels {
-		_, inSpec := specLabels[nsn]
-		_, inStatus := statusLabels[nsn]
-		if !inSpec && !inStatus {
-			meta.RemoveLabel(
-				&node,
-				utils.GetKernelModuleReadyNodeLabel(nsn.Namespace, nsn.Name),
-			)
-
-			unloaded = append(unloaded, nsn)
-		}
-	}
-
-	// v1 ready labels, deprecated - should be removed
-	for label := range deprecatedNodeModuleReadyLabels {
-		meta.RemoveLabel(&node, label)
-	}
+	nsnLabelsToBeRemoved := h.lph.removeOrphanedLabels(nodeModuleReadyLabels, specLabels, statusLabels)
 
 	// label in spec and status and config equal - should be added
-	for nsn, specConfig := range specLabels {
-		statusConfig, ok := statusLabels[nsn]
-		if ok && reflect.DeepEqual(specConfig, statusConfig) && !nodeModuleReadyLabels.Has(nsn) {
-			meta.SetLabel(
-				&node,
-				utils.GetKernelModuleReadyNodeLabel(nsn.Namespace, nsn.Name),
-				"",
-			)
+	nsnLabelsToBeLoaded := h.lph.addEqualLabels(nodeModuleReadyLabels, specLabels, statusLabels)
 
-			loaded = append(loaded, nsn)
-		}
+	var loadedLabels []string
+	unloadedLabels := deprecatedNodeModuleReadyLabels.UnsortedList()
+
+	for _, label := range nsnLabelsToBeRemoved {
+		unloadedLabels = append(unloadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
+	}
+	for _, label := range nsnLabelsToBeLoaded {
+		loadedLabels = append(loadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
 	}
 
-	if err := h.client.Patch(ctx, &node, patchFrom); err != nil {
-		return fmt.Errorf("could not patch node: %v", err)
+	if err := h.nodeAPI.UpdateLabels(ctx, node, loadedLabels, unloadedLabels); err != nil {
+		return nil, nil, fmt.Errorf("could not update labels on the node: %v", err)
 	}
 
-	for _, nsn := range unloaded {
+	return nsnLabelsToBeLoaded, nsnLabelsToBeRemoved, nil
+}
+
+func (h *nmcReconcilerHelperImpl) RecordEvents(node *v1.Node, loadedModules, unloadedModules []types.NamespacedName) {
+	for _, nsn := range loadedModules {
 		h.recorder.AnnotatedEventf(
-			&node,
-			map[string]string{"module": nsn.String()},
-			v1.EventTypeNormal,
-			"ModuleUnloaded",
-			"Module %s unloaded from the kernel",
-			nsn.String(),
-		)
-	}
-
-	for _, nsn := range loaded {
-		h.recorder.AnnotatedEventf(
-			&node,
+			node,
 			map[string]string{"module": nsn.String()},
 			v1.EventTypeNormal,
 			"ModuleLoaded",
@@ -651,8 +667,44 @@ func (h *nmcReconcilerHelperImpl) UpdateNodeLabelsAndRecordEvents(ctx context.Co
 			nsn.String(),
 		)
 	}
+	for _, nsn := range unloadedModules {
+		h.recorder.AnnotatedEventf(
+			node,
+			map[string]string{"module": nsn.String()},
+			v1.EventTypeNormal,
+			"ModuleUnloaded",
+			"Module %s unloaded from the kernel",
+			nsn.String(),
+		)
+	}
+}
 
-	return nil
+func (lph *labelPreparationHelperImpl) removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+	specLabels, statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName {
+
+	unloaded := make([]types.NamespacedName, 0, len(nodeModuleReadyLabels))
+
+	for nsn := range nodeModuleReadyLabels {
+		_, inSpec := specLabels[nsn]
+		_, inStatus := statusLabels[nsn]
+		if !inSpec && !inStatus {
+			unloaded = append(unloaded, nsn)
+		}
+	}
+	return unloaded
+}
+func (lph *labelPreparationHelperImpl) addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+	specLabels, statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName {
+
+	loaded := make([]types.NamespacedName, 0, len(nodeModuleReadyLabels))
+
+	for nsn, specConfig := range specLabels {
+		statusConfig, ok := statusLabels[nsn]
+		if ok && reflect.DeepEqual(specConfig, statusConfig) && !nodeModuleReadyLabels.Has(nsn) {
+			loaded = append(loaded, nsn)
+		}
+	}
+	return loaded
 }
 
 const (
--- a/internal/controllers/nmc_reconciler_test.go
+++ b/internal/controllers/nmc_reconciler_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -36,7 +37,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const nmcName = "nmc"
+const (
+	nmcName     = "nmc"
+	nsFirst     = "example-ns-1"
+	nsSecond    = "example-ns-2"
+	nameFirst   = "example-name-1"
+	nameSecond  = "example-name-2"
+	imageFirst  = "example-image-1"
+	imageSecond = "example-image-2"
+)
 
 var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 	var (
@@ -101,6 +110,12 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			mod1Name = "mod1"
 			mod2Name = "mod2"
 		)
+		var (
+			loaded   []types.NamespacedName
+			unloaded []types.NamespacedName
+			err      error
+			node     v1.Node
+		)
 		spec0 := kmmv1beta1.NodeModuleSpec{
 			ModuleItem: kmmv1beta1.ModuleItem{
 				Namespace: namespace,
@@ -155,7 +170,9 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			wh.EXPECT().ProcessModuleSpec(contextWithValueMatch, nmc, &spec1, nil),
 			wh.EXPECT().ProcessUnconfiguredModuleStatus(contextWithValueMatch, nmc, &status2),
 			wh.EXPECT().GarbageCollectInUseLabels(ctx, nmc),
-			wh.EXPECT().UpdateNodeLabelsAndRecordEvents(ctx, nmc),
+			kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(nil),
+			wh.EXPECT().UpdateNodeLabels(ctx, nmc, &node).Return(loaded, unloaded, err),
+			wh.EXPECT().RecordEvents(&node, loaded, unloaded),
 		)
 
 		Expect(
@@ -171,6 +188,11 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			mod1Name = "mod1"
 			mod2Name = "mod2"
 		)
+		var (
+			node v1.Node
+			err  error
+		)
+
 		spec0 := kmmv1beta1.NodeModuleSpec{
 			ModuleItem: kmmv1beta1.ModuleItem{
 				Namespace: namespace,
@@ -217,10 +239,11 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			wh.EXPECT().ProcessModuleSpec(contextWithValueMatch, nmc, &spec0, &status0).Return(fmt.Errorf("some error")),
 			wh.EXPECT().ProcessUnconfiguredModuleStatus(contextWithValueMatch, nmc, &status2).Return(fmt.Errorf("some error")),
 			wh.EXPECT().GarbageCollectInUseLabels(ctx, nmc).Return(fmt.Errorf("some error")),
-			wh.EXPECT().UpdateNodeLabelsAndRecordEvents(ctx, nmc).Return(fmt.Errorf("some error")),
+			kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(nil),
+			wh.EXPECT().UpdateNodeLabels(ctx, nmc, &node).Return(nil, nil, fmt.Errorf("some error")),
 		)
 
-		_, err := r.Reconcile(ctx, req)
+		_, err = r.Reconcile(ctx, req)
 		Expect(err).ToNot(BeNil())
 	})
 })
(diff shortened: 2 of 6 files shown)


yevgeny-shnaidman:
It looks good to me , but i have another suggestion:
most of the code of UpdateNodeLabelsAndRecordEvents function deals with preparing the labels for the node. What if we define another helper interface, that will handle that preparation? This way we can unit-test the UpdateNodeLabelsAndRecordEvents properly and in addition test the labels preparation thoroughly also, since it will be in a different interface. @ybettan , @TomerNewman , what do you think? Or is it an overkill?

TomerNewman:
> It looks good to me , but i have another suggestion: most of the code of UpdateNodeLabelsAndRecordEvents function deals with preparing the labels for the node. What if we define another helper interface, that will handle that preparation? This way we can unit-test the UpdateNodeLabelsAndRecordEvents properly and in addition test the labels preparation thoroughly also, since it will be in a different interface. @ybettan , @TomerNewman , what do you think? Or is it an overkill?

so it should be in a different package? maybe in `internal->utils->kmmlabels.go?`

yevgeny-shnaidman:
> > It looks good to me , but i have another suggestion: most of the code of UpdateNodeLabelsAndRecordEvents function deals with preparing the labels for the node. What if we define another helper interface, that will handle that preparation? This way we can unit-test the UpdateNodeLabelsAndRecordEvents properly and in addition test the labels preparation thoroughly also, since it will be in a different interface. @ybettan , @TomerNewman , what do you think? Or is it an overkill?
> 
> so it should be in a different package? maybe in `internal->utils->kmmlabels.go?`

No, it can be in the same package, for example like PodManager interface

internal/controllers/nmc_reconciler.go
```diff
@@ -565,94 +565,69 @@ func (h *nmcReconcilerHelperImpl) UpdateNodeLabelsAndRecordEvents(ctx context.Co
...
-	for label := range node.GetLabels() {
-		if ok, namespace, name := utils.IsKernelModuleReadyNodeLabel(label); ok {
-			nodeModuleReadyLabels.Insert(types.NamespacedName{Namespace: namespace, Name: name})
-		}
-
-		if utils.IsDeprecatedKernelModuleReadyNodeLabel(label) {
-			deprecatedNodeModuleReadyLabels.Insert(label)
-		}
-	}
+	getKernelModuleReadyLabels(&nodeModuleReadyLabels, &deprecatedNodeModuleReadyLabels, node)
```
ybettan, line 568: A `get*` function should return something. In your case, looks like a `set*` function will be a better fit.

BUT, I would keep it a `get*` function that return a new `Set` and then intersect the returned value with `nodeModuleReadyLabels`.

Also, I would split this function into 2, one that deals with labels and second one with deprecated labels.

internal/controllers/nmc_reconciler.go
```diff
@@ -565,94 +565,69 @@ func (h *nmcReconcilerHelperImpl) UpdateNodeLabelsAndRecordEvents(ctx context.Co
...
 
 	// get spec labels and their config
 	specLabels := make(map[types.NamespacedName]kmmv1beta1.ModuleConfig)
-	for _, module := range nmc.Spec.Modules {
-		specLabels[types.NamespacedName{Namespace: module.Namespace, Name: module.Name}] = module.Config
-	}
+	getSpecLabelsAndTheirConfigs(specLabels, nmc)
```
ybettan, line 572: This function should return a map.

internal/controllers/nmc_reconciler.go
```diff
@@ -565,94 +565,69 @@ func (h *nmcReconcilerHelperImpl) UpdateNodeLabelsAndRecordEvents(ctx context.Co
...
-		h.recorder.AnnotatedEventf(
-			&node,
-			map[string]string{"module": nsn.String()},
-			v1.EventTypeNormal,
-			"ModuleUnloaded",
-			"Module %s unloaded from the kernel",
-			nsn.String(),
-		)
+	h.recordModuleEvents(&node, unloaded, "unload")
+	h.recordModuleEvents(&node, loaded, "load")
```
ybettan, line 599: @yevgeny-shnaidman @TomerNewman How about we extract the events stuff to the caller of `UpdateNodeLabelsAndRecordEvents` instead of having 1 function doing "multiple different things?

ybettan:
> It looks good to me , but i have another suggestion:
most of the code of UpdateNodeLabelsAndRecordEvents function deals with preparing the labels for the node. What if we define another helper interface, that will handle that preparation? This way we can unit-test the UpdateNodeLabelsAndRecordEvents properly and in addition test the labels preparation thoroughly also, since it will be in a different interface. @ybettan , @TomerNewman , what do you think? Or is it an overkill?

I think that the parts that add/remove labels from the nodes should be part of the new `node` interface as long as the label is passed as a parameter.

For other cases, I definitely think we should create another interface and I would also somehow take into consideration `internal->utils->kmmlabels.go` as it is serving a similar purpose.

yevgeny-shnaidman:
> > It looks good to me , but i have another suggestion:
> > most of the code of UpdateNodeLabelsAndRecordEvents function deals with preparing the labels for the node. What if we define another helper interface, that will handle that preparation? This way we can unit-test the UpdateNodeLabelsAndRecordEvents properly and in addition test the labels preparation thoroughly also, since it will be in a different interface. @ybettan , @TomerNewman , what do you think? Or is it an overkill?
> 
> I think that the parts that add/remove labels from the nodes should be part of the new `node` interface as long as the label is passed as a parameter.
> 
> For other cases, I definitely think we should create another interface and I would also somehow take into consideration `internal->utils->kmmlabels.go` as it is serving a similar purpose.

I thought about kmmlabels, but its main purpose is to define the labels, to to manipulate what and when should go to wihat node

TomerNewman:
Following our slack conversation, I've also split `UpdateNodeLabelsAndRecordEvents` and moved `UpdateNodeLabels` to Node package.
I've also added and changed the unit test accordingly.
I think the PR is way to big, what do we think about removing some of the features into new PR\s? @ybettan @yevgeny-shnaidman

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-		return fmt.Errorf("could not get node %s: %v", nmc.Name, err)
-	}
+type LabelPreparationHelper interface {
+	GetDeprecatedKernelModuleReadyLabels(deprecatedNodeModuleReadyLabels sets.Set[string], node v1.Node) sets.Set[string]
+	GetKernelModuleReadyLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName], node v1.Node) sets.Set[types.NamespacedName]
+	GetSpecLabelsAndTheirConfigs(specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig, nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	GetStatusLabelsAndTheirConfigs(statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig, nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+}
+type labelPreparationHelperImpl struct {
+}
```
ybettan, line 575:
```suggestion
type labelPreparationHelperImpl struct {}
```

internal/node/node.go
```diff
@@ -72,3 +77,40 @@ func (n *node) FindNodeCondition(cond []v1.NodeCondition, conditionType v1.NodeC
 
 	return nil
 }
+
+func (n *node) AddModulesLabelsToNode(modules []types.NamespacedName, node *v1.Node) {
```
ybettan, line 81: Who is using this method?

internal/node/node.go
```diff
@@ -13,6 +17,7 @@ type Node interface {
 	GetNodesListBySelector(ctx context.Context, selector map[string]string) ([]v1.Node, error)
 	GetNumTargetedNodes(ctx context.Context, selector map[string]string) (int, error)
 	FindNodeCondition(cond []v1.NodeCondition, conditionType v1.NodeConditionType) *v1.NodeCondition
+	UpdateNodeLabels(ctx context.Context, deprecatedNodeModuleReadyLabels sets.Set[string], loaded []types.NamespacedName, unloaded []types.NamespacedName, node *v1.Node) error
```
ybettan, line 20:
```suggestion
	UpdateLabels(ctx context.Context, node *v1.Node, toBeAdded, toBeRemoved []string) error
```

Few comments here:
1. Everything in this package is related to nodes, therefore, there the name `UpdateNodeLabels` can just be `UpdateLabels`.
2. The main object here is the node, therefore, it should be the first param in the method
3. This methods know nothing about `modules`, `readyLabels` etc - it should be simple - get a slice of labels to be added and a list of labels to be removed and just add/remove them from the node
4. Since you are having an API call to the kubeAPI for each label (`meta.RemoveLabel` one by one), I would split this method into 2 `AddLabels` and `RemoveLabels`. If you do a single API call to the kubeAPI for adding/removing all the laels in a single call then it make sense to have a single method for it.

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-	node := v1.Node{}
-	if err := h.client.Get(ctx, types.NamespacedName{Name: nmc.Name}, &node); err != nil {
-		return fmt.Errorf("could not get node %s: %v", nmc.Name, err)
-	}
+type LabelPreparationHelper interface {
+	GetDeprecatedKernelModuleReadyLabels(deprecatedNodeModuleReadyLabels sets.Set[string], node v1.Node) sets.Set[string]
+	GetKernelModuleReadyLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName], node v1.Node) sets.Set[types.NamespacedName]
+	GetSpecLabelsAndTheirConfigs(specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig, nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	GetStatusLabelsAndTheirConfigs(statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig, nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+}
```
ybettan, line 573: Let's make it a non-exported interface (same as `nmcReconcilerHelper`)?
@yevgeny-shnaidman Although all those methods are labels related, won't it be cleaner to define them inside `nmcReconcilerHelper`?

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+type labelPreparationHelperImpl struct {
+}
 
-	// get all the kernel module ready labels of the node
-	nodeModuleReadyLabels := sets.New[types.NamespacedName]()
-	deprecatedNodeModuleReadyLabels := sets.New[string]()
+func NewLabelPreparationHelper() LabelPreparationHelper {
```
ybettan, line 577: If we don't converge all those methods into `nmcReconcilerHelper` then let's make this function private (same as `newNmcReconcilerHelper`

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 	return errors.Join(errs...)
 }
 
-func (h *nmcReconcilerHelperImpl) UpdateNodeLabelsAndRecordEvents(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error {
-	node := v1.Node{}
-	if err := h.client.Get(ctx, types.NamespacedName{Name: nmc.Name}, &node); err != nil {
-		return fmt.Errorf("could not get node %s: %v", nmc.Name, err)
-	}
+type LabelPreparationHelper interface {
```
ybettan, line 568: Are all those methods has to be exported? Even if the interface isn't exported?

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+	return &labelPreparationHelperImpl{}
+}
 
+func (lphi *labelPreparationHelperImpl) GetKernelModuleReadyLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName], node v1.Node) sets.Set[types.NamespacedName] {
 	for label := range node.GetLabels() {
 		if ok, namespace, name := utils.IsKernelModuleReadyNodeLabel(label); ok {
 			nodeModuleReadyLabels.Insert(types.NamespacedName{Namespace: namespace, Name: name})
 		}
+	}
+	return nodeModuleReadyLabels
```
ybettan, line 587: A `set` is a map (pointer) which means that when you `Insert` to it, it get updated, therefore there is not need to also return it.

I would define the Set inside the function and return it.
A function that has no state is easier to debug and preferable IMO

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+}
 
-	// get all the kernel module ready labels of the node
-	nodeModuleReadyLabels := sets.New[types.NamespacedName]()
-	deprecatedNodeModuleReadyLabels := sets.New[string]()
+func NewLabelPreparationHelper() LabelPreparationHelper {
+	return &labelPreparationHelperImpl{}
+}
 
+func (lphi *labelPreparationHelperImpl) GetKernelModuleReadyLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName], node v1.Node) sets.Set[types.NamespacedName] {
```
ybettan, line 581:
```suggestion
func (lphi *labelPreparationHelperImpl) GetNodeKernelModuleReadyLabels(node v1.Node) sets.Set[types.NamespacedName] {
```

ybettan: Same for other methods in this interface.

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-			"Module %s loaded into the kernel",
+			"ModuleUnloaded",
+			"Module %s unloaded from the kernel",
 			nsn.String(),
 		)
 	}
+}
 
-	return nil
+func (h *nmcReconcilerHelperImpl) removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
```
ybettan, line 670: Any reason this method isn't part of the help interface as well?

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
 	}
 
-	if err := h.client.Patch(ctx, &node, patchFrom); err != nil {
-		return fmt.Errorf("could not patch node: %v", err)
-	}
+	return loaded, unloaded, nil
+}
 
-	for _, nsn := range unloaded {
+func (h *nmcReconcilerHelperImpl) RecordEvents(loaded []types.NamespacedName, unloaded []types.NamespacedName, node *v1.Node) {
```
ybettan, line 647: Can we use `loadedModules` and `unloadedModules`?

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
 			nsn.String(),
 		)
 	}
+}
 
-	return nil
+func (h *nmcReconcilerHelperImpl) removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+	specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+	statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+	unloaded []types.NamespacedName) []types.NamespacedName {
```
ybettan, line 673: Can we define this one inside the function?

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,132 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+		if !inSpec && !inStatus {
+			unloaded = append(unloaded, nsn)
+		}
+	}
+	return unloaded
+}
+func (h *nmcReconcilerHelperImpl) addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+	specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+	statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+	loaded []types.NamespacedName) []types.NamespacedName {
```
ybettan, line 686: Can we define this one inside the method?

internal/controllers/nmc_reconciler.go
TomerNewman: we use `NewLabelPreparationHelper` in cmd/manager/main.go line 124.
so I think that if we use this interface, we need this function to be Public

internal/controllers/nmc_reconciler.go
TomerNewman: so all of the methods in labelPreparationHelper should be private?
I was trying to take an example from `podManager` interface

internal/controllers/nmc_reconciler.go
TomerNewman: I thought because its an helper (inner) function, so if someone wants to implement this interface outside of this package, we wont have to force him to implement inner functions

internal/node/node.go
TomerNewman: my bad, should have deleted it sooner, its fixed now

internal/node/node.go
TomerNewman: understood, this is fixed.
now before `UpdateLabels` is being called, we create those two string slices out of the `type.NamespacedName` slices and the set

internal/controllers/nmc_reconciler.go
TomerNewman: done, `LabelPreparationHelper` was changed to `labelPreparationHelper`

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
TomerNewman: yes.
also, do you think `loadedModules` and `unloadedModules` should be []string instead?

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,138 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+type labelPreparationHelper interface {
+	getDeprecatedKernelModuleReadyLabels(node v1.Node) sets.Set[string]
+	getNodeKernelModuleReadyLabels(node v1.Node) sets.Set[types.NamespacedName]
+	getSpecLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	getStatusLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+}
+type labelPreparationHelperImpl struct{}
 
-	// get all the kernel module ready labels of the node
+func NewLabelPreparationHelper() labelPreparationHelper {
```
yevgeny-shnaidman, line 576: We don't need this function to be public. This is an inner interface to the package, it can be initialized by the NewNMCReconcile , just like all other helper interfaces

internal/utils/convert_types.go
```diff
@@ -0,0 +1,26 @@
...
+
+func SetToSlice[T comparable](set sets.Set[T]) []T {
+
+	slice := make([]T, 0, set.Len())
+
+	for item := range set {
+		slice = append(slice, item)
+	}
+	return slice
+}
```
ybettan, line 26: This is already implemented in the `sets` package as `set.UnsortedList`, therefore, we can remove this function.

internal/utils/convert_types.go
```diff
@@ -0,0 +1,26 @@
+package utils
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+func ConvertNamespacedNameToStrings(toBeConverted []types.NamespacedName) []string {
```
ybettan, line 8: This function isn't converting `[]NamespacedName` to `[]string` but `[]NamespacedName` to a very specific KMM label list.

How about we just return lists from `removeOrphanedLabels` and `addEqualLabels` instead of lists of `NamespacedName` and then write another logic to convert it?

We can completely remove this file IMO.

internal/node/node.go
```diff
@@ -72,3 +74,34 @@ func (n *node) FindNodeCondition(cond []v1.NodeCondition, conditionType v1.NodeC
...
+	n.addLabels(node, toBeAdded)
+	n.removeLabels(node, toBeRemoved)
+
+	if err := n.client.Patch(ctx, node, patchFrom); err != nil {
+		return fmt.Errorf("could not patch node: %v", err)
+	}
+	return nil
+}
+
+func (n *node) addLabels(node *v1.Node, toBeAdded []string) {
```
ybettan, line 90:
```suggestion
func (n *node) addLabels(node *v1.Node, labels []string) {
```

internal/node/node.go
```diff
@@ -72,3 +74,34 @@ func (n *node) FindNodeCondition(cond []v1.NodeCondition, conditionType v1.NodeC
...
+	for _, label := range toBeAdded {
+		meta.SetLabel(
+			node,
+			label,
+			"",
+		)
+	}
+}
+
+func (n *node) removeLabels(node *v1.Node, toBeRemoved []string) {
```
ybettan, line 100:
```suggestion
func (n *node) removeLabels(node *v1.Node, labels []string) {
```

internal/controllers/nmc_reconciler.go
```diff
@@ -213,22 +220,25 @@ type nmcReconcilerHelper interface {
 	ProcessUnconfiguredModuleStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, status *kmmv1beta1.NodeModuleStatus) error
 	RemovePodFinalizers(ctx context.Context, nodeName string) error
 	SyncStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
-	UpdateNodeLabelsAndRecordEvents(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
+	PrepareAndUpdateNodeLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) ([]types.NamespacedName, []types.NamespacedName, error)
```
ybettan, line 223:
```suggestion
	UpdateNodeLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) ([]types.NamespacedName, []types.NamespacedName, error)
```

internal/controllers/nmc_reconciler.go
ybettan: You main point is correct, but we can say it about all the methods in the helper function. No one else will implement that interface and we only do it for easier unit-testing so having those methods in the interface is a better option IMO.

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,138 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+			"Module %s unloaded from the kernel",
 			nsn.String(),
 		)
 	}
+}
 
-	return nil
+func (h *nmcReconcilerHelperImpl) removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+	specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+	statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName {
```
ybettan, line 677:
```suggestion
	statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []string {
```
Can we just return a label list instead? No reason to build an object just to parse it later on.

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,138 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+		_, inStatus := statusLabels[nsn]
+		if !inSpec && !inStatus {
+			unloaded = append(unloaded, nsn)
+		}
+	}
+	return unloaded
+}
+func (h *nmcReconcilerHelperImpl) addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+	specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+	statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName {
```
ybettan, line 690:
```suggestion
	statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []string {
```

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +565,138 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+	for nsn := range nodeModuleReadyLabels {
+		_, inSpec := specLabels[nsn]
+		_, inStatus := statusLabels[nsn]
+		if !inSpec && !inStatus {
+			unloaded = append(unloaded, nsn)
+		}
+	}
+	return unloaded
+}
+func (h *nmcReconcilerHelperImpl) addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
```
ybettan, line 688: I would make this function an interface method as well.

internal/controllers/nmc_reconciler.go
ybettan: I agree, but you will need to manually create the object when testing in order to use a mock since there is no dependency injection anymore.

internal/utils/convert_types.go
TomerNewman: @ybettan so in `removeOrphanedLabels` and `addEqualLabels` we will return []string and we should have a function that will convert the []string label list to a []type.NamespaceName list?

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: we do it all the time in the unit-test

internal/utils/convert_types.go
ybettan: No. I mean instead of:
* Returning `[]NamespacedName` and then write code to convert it

Let's just:
* Return `[]string` (then we don't need any logic to convert it) - the only use we do with `[]NamespacedName` is to convert it so it is just redundant.

internal/controllers/nmc_reconciler.go
TomerNewman: following our meeting, I left it as it was

internal/node/node.go
TomerNewman: done

internal/node/node.go
TomerNewman: done

internal/utils/convert_types.go
TomerNewman: The file was deleted

internal/utils/convert_types.go
TomerNewman: File was deleted

internal/controllers/nmc_reconciler.go
TomerNewman: done.
now it is initialized in `newNMCReconcilerHelper` instead of `NewNMCReconcile` because we need it in  `nmcReconcilerHelperImpl` struct

internal/node/node.go
```diff
@@ -72,3 +76,45 @@ func (n *node) FindNodeCondition(cond []v1.NodeCondition, conditionType v1.NodeC
...
+	}
+	return nil
+}
+
+type nodeHelperImpl struct{}
+
+type nodeHelper interface {
+	addLabels(node *v1.Node, labels []string)
+	removeLabels(node *v1.Node, labels []string)
+}
```
ybettan, line 97: I am not sure a helper interface is the best option here. I would make them non-exported functions (not methods) in the package.

Here are some points to consider when deciding if something should be an interface/function.
1. As a general rule of thumb, if a method isn't using its "self" object then it should be a function.
2. If an API is exported, using its "self" object and wish to be mocked in unit-tests then we use an exported interface.
3. In some cases, we are having exceptions in the form of "helper, non-exported interfaces" - in cases where a packages has a lot of non-exported functions/methods we sometimes choose to put them inside an helper interface (even if the "self" object isn't being used just for the ease of unit-testing the package because it becomes difficult to test.

In this case, `addLabels` and `removeLabels`:
1. Does not use the "self" (node) object - leaning towards being functions rather than methods.
2. Are pretty short/simple and there are only 2 "helper" functions so it is not too difficult to unit-test `UpdateLabels`

Therefore I would make those non-exported functions and remove the helper interface in this package.

internal/controllers/nmc_reconciler.go
```diff
@@ -213,14 +219,16 @@ type nmcReconcilerHelper interface {
 	ProcessUnconfiguredModuleStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, status *kmmv1beta1.NodeModuleStatus) error
 	RemovePodFinalizers(ctx context.Context, nodeName string) error
 	SyncStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
-	UpdateNodeLabelsAndRecordEvents(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
+	UpdateNodeLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) ([]types.NamespacedName, []types.NamespacedName, error)
+	RecordEvents(loadedModules []types.NamespacedName, unloadedModules []types.NamespacedName, node *v1.Node)
```
ybettan, line 223:
```suggestion
	RecordEvents(node *v1.Node, loadedModules, unloadedModules []types.NamespacedName, )
```

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +564,154 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-		return fmt.Errorf("could not get node %s: %v", nmc.Name, err)
-	}
+type labelPreparationHelper interface {
+	getDeprecatedKernelModuleReadyLabels(node v1.Node) sets.Set[string]
+	getNodeKernelModuleReadyLabels(node v1.Node) sets.Set[types.NamespacedName]
+	getSpecLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	getStatusLabelsAndTheirConfigs(nmc *kmmv1beta1.NodeModulesConfig) map[types.NamespacedName]kmmv1beta1.ModuleConfig
+	addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+		specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+		statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName
```
ybettan, line 574:
```suggestion
	addEqualLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
		specLabels, statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName
```

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +564,154 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+	removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
+		specLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig,
+		statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName
```
ybettan, line 577:
```suggestion
	removeOrphanedLabels(nodeModuleReadyLabels sets.Set[types.NamespacedName],
		specLabels, statusLabels map[types.NamespacedName]kmmv1beta1.ModuleConfig) []types.NamespacedName
```

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +564,154 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-				&node,
-				utils.GetKernelModuleReadyNodeLabel(nsn.Namespace, nsn.Name),
-				"",
-			)
+	nsnLabelsToBeLoaded := h.lph.addEqualLabels(nodeModuleReadyLabels, specLabels, statusLabels)
 
-			loaded = append(loaded, nsn)
-		}
+	unloadedLabels := []string{}
+	loadedLabels := []string{}
```
ybettan, line 645:
```suggestion
        var loadedLabels unloadedLabels []string
```

See the difference:
https://go.dev/play/p/JBrR0homl_7

Using `nil` will result in an error in case we do something illegal with the slices (which is better than having a bug that isn't erroring).

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +564,154 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
+		unloadedLabels = append(unloadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
+	}
+	for _, label := range nsnLabelsToBeLoaded {
+		loadedLabels = append(loadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
 	}
 
-	if err := h.client.Patch(ctx, &node, patchFrom); err != nil {
-		return fmt.Errorf("could not patch node: %v", err)
+	if err := h.nodeAPI.UpdateLabels(ctx, node, loadedLabels, unloadedLabels); err != nil {
+		return nil, nil, fmt.Errorf("could not patch node: %v", err)
```
ybettan, line 658:
```suggestion
		return nil, nil, fmt.Errorf("could not update labels on the node: %v", err)
```

internal/node/node.go
TomerNewman: @ybettan  so in that case, you think that `addLabels` and `removeLabels` should not be mocked? 
they are pretty simple, right, but mocking them helps for example in line 173 in node_test.go

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/controllers/nmc_reconciler.go
TomerNewman: done

internal/node/node.go
TomerNewman: deleted the `nodeHelper` interface

ybettan: > so in that case, you think that addLabels and removeLabels should not be mocked?

Indeed. In this case I don't think the mocking benefit worth writing another interface.

internal/controllers/nmc_reconciler.go
```diff
@@ -555,104 +564,149 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
...
-			)
+	nsnLabelsToBeLoaded := h.lph.addEqualLabels(nodeModuleReadyLabels, specLabels, statusLabels)
 
-			loaded = append(loaded, nsn)
-		}
+	var loadedLabels, unloadedLabels []string
+
+	for label := range deprecatedNodeModuleReadyLabels {
+		unloadedLabels = append(unloadedLabels, label)
+	}
```
ybettan, line 646:
```suggestion
var loadedLabels []string
unloadedLabels := deprecatedNodeModuleReadyLabels.UnsortedList()

for _, label := range nsnLabelsToBeRemoved {
    unloadedLabels = append(unloadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
}
```

internal/node/node.go
```diff
@@ -13,6 +14,9 @@ type Node interface {
 	GetNodesListBySelector(ctx context.Context, selector map[string]string) ([]v1.Node, error)
 	GetNumTargetedNodes(ctx context.Context, selector map[string]string) (int, error)
 	FindNodeCondition(cond []v1.NodeCondition, conditionType v1.NodeConditionType) *v1.NodeCondition
+	UpdateLabels(ctx context.Context, node *v1.Node, toBeAdded, toBeRemoved []string) error
+	addLabels(node *v1.Node, labels []string)
+	removeLabels(node *v1.Node, labels []string)
```
ybettan, line 19: Let's remove them from the interface and make then non-exported *functions*.

https://github.com/kubernetes-sigs/kernel-module-management/pull/850#discussion_r1673821877

internal/node/node_test.go
```diff
@@ -220,3 +269,64 @@ var _ = Describe("FindNodeCondition", func() {
...
+
+	BeforeEach(func() {
+		n = NewNode(nil)
+		node = v1.Node{}
+	})
+
+	It("Should add labels", func() {
+		labels := []string{kernelModuleReadyNodeLabel}
+
+		Expect(node.Labels).ToNot(HaveKey(kernelModuleReadyNodeLabel))
```
ybettan, line 288:
```suggestion
```

internal/node/node_test.go
```diff
@@ -220,3 +269,64 @@ var _ = Describe("FindNodeCondition", func() {
...
+		Expect(node.Labels).To(HaveKey(kernelModuleReadyNodeLabel))
+	})
+	It("Should not add labels", func() {
+		var labels []string
+
+		Expect(node.Labels).ToNot(HaveKey(kernelModuleReadyNodeLabel))
+		n.addLabels(&node, labels)
+		Expect(node.Labels).ToNot(HaveKey(kernelModuleReadyNodeLabel))
+	})
+})
```
ybettan, line 299: We can remove this test - it is not really testing anything different than the previous spec.

internal/node/node_test.go
```diff
@@ -220,3 +269,64 @@ var _ = Describe("FindNodeCondition", func() {
...
+					kernelModuleReadyNodeLabel: "",
+				},
+			},
+		}
+	})
+
+	It("Should remove labels", func() {
+		labels := []string{kernelModuleReadyNodeLabel}
+
+		Expect(node.Labels).To(HaveKey(kernelModuleReadyNodeLabel))
```
ybettan, line 321:
```suggestion
```

internal/node/node_test.go
```diff
@@ -220,3 +269,64 @@ var _ = Describe("FindNodeCondition", func() {
...
+		n.removeLabels(&node, labels)
+		Expect(node.Labels).ToNot(HaveKey(kernelModuleReadyNodeLabel))
+	})
+	It("Should not remove labels", func() {
+		var labels []string
+
+		Expect(node.Labels).To(HaveKey(kernelModuleReadyNodeLabel))
+		n.removeLabels(&node, labels)
+		Expect(node.Labels).To(HaveKey(kernelModuleReadyNodeLabel))
+	})
```
ybettan, line 331: We can remove this test - it is not really testing anything different than the previous spec.

internal/node/node_test.go
TomerNewman: @ybettan  why not? in theory `kernelModuleReadyNodeLabel` can be added no matter what in `addLabels`, so dont we need to test in both directions? (one with kernelModuleReadyNodeLabel and one without kernelModuleReadyNodeLabel)

ybettan: In the previous test, you start with an empty node, add a label and check if the label was added.
In this test, you start with an empty node, call `addLabels` with an empty list and check that nothing happened.
I am not sure what added value this test have over the previous one.

internal/node/node_test.go
ybettan: Same as https://github.com/kubernetes-sigs/kernel-module-management/pull/850#discussion_r1677109887.

It is not crucial at all to have it. If you think there is benefit in it then feel free to keep it. I don't see a lot of value in it but it is your decision :)

internal/controllers/nmc_reconciler_test.go
```diff
@@ -143,6 +158,9 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			reflect.TypeOf((*context.Context)(nil)).Elem(),
 		)
 
+		kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(err)
```
ybettan, line 161:
```suggestion
		kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(nil)
```

If we don't return any error, then let's return `nil` it is clearer then searching what `err` got as its value.

Also, any reason we don't put this line inside the `InOrder` block?

internal/controllers/nmc_reconciler_test.go
```diff
@@ -143,6 +158,9 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
...
+		Expect(err).NotTo(HaveOccurred())
```
ybettan, line 162: What are we testing here? If `err` was not initialized, then it will be `nil` by default.

internal/controllers/nmc_reconciler_test.go
```diff
@@ -205,6 +229,8 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 		contextWithValueMatch := gomock.AssignableToTypeOf(
 			reflect.TypeOf((*context.Context)(nil)).Elem(),
 		)
+		kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(err)
+		Expect(err).NotTo(HaveOccurred())
```
ybettan, line 233: Same as https://github.com/kubernetes-sigs/kernel-module-management/pull/850/files#r1677111464

internal/controllers/nmc_reconciler_test.go
```diff
@@ -217,10 +243,10 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			wh.EXPECT().ProcessModuleSpec(contextWithValueMatch, nmc, &spec0, &status0).Return(fmt.Errorf("some error")),
 			wh.EXPECT().ProcessUnconfiguredModuleStatus(contextWithValueMatch, nmc, &status2).Return(fmt.Errorf("some error")),
 			wh.EXPECT().GarbageCollectInUseLabels(ctx, nmc).Return(fmt.Errorf("some error")),
-			wh.EXPECT().UpdateNodeLabelsAndRecordEvents(ctx, nmc).Return(fmt.Errorf("some error")),
+			wh.EXPECT().UpdateNodeLabels(ctx, nmc, &node).Return(nil, nil, fmt.Errorf("some error")),
 		)
 
-		_, err := r.Reconcile(ctx, req)
+		_, err = r.Reconcile(ctx, req)
 		Expect(err).ToNot(BeNil())
```
ybettan, line 250: This comment is for purpose learning only as it is not directly related to your PR.

This is not a great validation. The description of the spec is `should complete all the reconcile functions and return combined error`, therefore, the validation should be to (first return different error messages and then) validate that the combined error contain each one of the error messages of the individual errors.

internal/controllers/nmc_reconciler_test.go
```diff
@@ -1158,39 +1184,143 @@ var _ = Describe("nmcReconcilerHelperImpl_RemovePodFinalizers", func() {
...
+			n.EXPECT().
+				UpdateLabels(
+					ctx,
+					&node,
+					[]string{fmt.Sprintf("kmm.node.kubernetes.io/%s.%s.ready", nsFirst, nameFirst)},
+					[]string{fmt.Sprintf("kmm.node.kubernetes.io/%s.%s.ready", nsSecond, nameSecond)},
+				).Return(nil),
+		)
+		_, _, err := wh.UpdateNodeLabels(ctx, &nmc, &node)
+		Expect(err).ToNot(HaveOccurred())
```
ybettan, line 1305: This is only testing that no error was returned. What about the part that validates that the node was updated correctly?

internal/controllers/nmc_reconciler_test.go
TomerNewman: fixed, moved `kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(err)` to the InOrder section (`Expect(err).NotTo(HaveOccurred())` should not be there indeed)

internal/controllers/nmc_reconciler_test.go
TomerNewman: fixed, moved `kubeClient.EXPECT().Get(ctx, types.NamespacedName{Name: nmc.Name}, &node).Return(err)` to the InOrder section (`Expect(err).NotTo(HaveOccurred())` should not be there indeed)

internal/controllers/nmc_reconciler_test.go
TomerNewman: fixed, moved it also to the InOrder section

internal/node/node_test.go
TomerNewman: test is removed

internal/node/node_test.go
TomerNewman: test is removed

internal/controllers/nmc_reconciler_test.go
TomerNewman: I added another Except for the new node label (next line)

internal/controllers/nmc_reconciler_test.go
ybettan: You don't have to add it to the PR so we can merge it faster. You can fix it in a future PR if you like. But to answer your question, the answer is YES.

One way to check it is like this example:
https://github.com/kubernetes-sigs/kernel-module-management/blob/c44a5e8d908a3bdd039f1f5cb8ee18663499eb6e/internal/registry/registry_test.go#L57

internal/node/node_test.go
```diff
@@ -118,6 +120,59 @@ var _ = Describe("GetNodesListBySelector", func() {
...
+		unloaded := []string{unloadedKernelModuleReadyNodeLabel}
+
+		clnt.EXPECT().Patch(ctx, gomock.Any(), gomock.Any(), gomock.Any()).Return(nil)
+
+		err := n.UpdateLabels(ctx, &node, loaded, unloaded)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(node.Labels).To(HaveKey(loadedKernelModuleReadyNodeLabel))
+	})
+
+	It("Should fail to patch node", func() {
```
ybettan, line 159:
```suggestion
	It("Should fail if we fail to patch node", func() {
```

PR #853, opened 2024-07-02
Refusing `Modules` with container images without a sha or a tag.

ybettan:
In the current code, we handle differently build/sign and deploying a pre-built kmod.

This is the current behavior:
* If a `build`/`sign` section is set in the module, and the image tag isn't specified, then we will return an error.
* If there are no `build`/`sign` section, and the image tag isn't specified, then we will default to the `latest` tag.

This change is adjusting the behavior of both workflow to enforce the customer to explicitely set a sha or a tag in his container image.

---

/cc @yevgeny-shnaidman @TomerNewman 
This PR is replacing https://github.com/kubernetes-sigs/kernel-module-management/pull/826 due to its difficulties.

### DIFF
--- a/internal/webhook/module.go
+++ b/internal/webhook/module.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/go-logr/logr"
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
@@ -111,9 +112,25 @@ func validateModule(mod *kmmv1beta1.Module) (admission.Warnings, error) {
 	return nil, validateModprobe(mod.Spec.ModuleLoader.Container.Modprobe)
 }
 
+func validateImageFormat(img string) error {
+
+	if !strings.Contains(img, ":") && !strings.Contains(img, "@") {
+		return fmt.Errorf("container image must explicitely set a tag or digest; got: %s", img)
+	}
+
+	return nil
+}
+
 func validateModuleLoaderContainerSpec(container kmmv1beta1.ModuleLoaderContainerSpec) error {
+
 	if container.InTreeModulesToRemove != nil && container.InTreeModuleToRemove != "" { //nolint:staticcheck
 		return fmt.Errorf("only one of the Container's fields: InTreeModulesToRemove or InTreeModuleToRemove can be defined")
+	}
+
+	if contImg := container.ContainerImage; contImg != "" {
+		if err := validateImageFormat(contImg); err != nil {
+			return fmt.Errorf("failed to validate image format: %v", err)
+		}
 	}
 
 	for idx, km := range container.KernelMappings {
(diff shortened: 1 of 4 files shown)


internal/webhook/module.go
```diff
@@ -112,10 +113,15 @@ func validateModule(mod *kmmv1beta1.Module) (admission.Warnings, error) {
 }
 
 func validateModuleLoaderContainerSpec(container kmmv1beta1.ModuleLoaderContainerSpec) error {
+
 	if container.InTreeModulesToRemove != nil && container.InTreeModuleToRemove != "" { //nolint:staticcheck
 		return fmt.Errorf("only one of the Container's fields: InTreeModulesToRemove or InTreeModuleToRemove can be defined")
 	}
 
+	if contImg := container.ContainerImage; contImg != "" && !strings.Contains(contImg, ":") && !strings.Contains(contImg, "@") {
```
yevgeny-shnaidman, line 121: WDYT about using NewTag and NewDigest from ithub.com/google/go-containerregistry? It provides a little bit more verifications on the input

internal/webhook/module.go
```diff
@@ -129,8 +135,14 @@ func validateModuleLoaderContainerSpec(container kmmv1beta1.ModuleLoaderContaine
...
 		}
 
-		if container.ContainerImage == "" && km.ContainerImage == "" {
-			return fmt.Errorf("missing spec.moduleLoader.container.kernelMappings[%d].containerImage", idx)
+		if kmImg := km.ContainerImage; kmImg == "" {
+			if container.ContainerImage == "" {
+				return fmt.Errorf("missing spec.moduleLoader.container.kernelMappings[%d].containerImage", idx)
+			}
+		} else {
+			if !strings.Contains(kmImg, ":") && !strings.Contains(kmImg, "@") {
```
yevgeny-shnaidman, line 143: WDYT about using NewTag and NewDigest from ithub.com/google/go-containerregistry? It provides a little bit more verifications on the input

internal/webhook/module.go
ybettan: Fixed.

internal/webhook/module.go
ybettan: Fixed.

internal/webhook/module.go
```diff
@@ -111,11 +112,31 @@ func validateModule(mod *kmmv1beta1.Module) (admission.Warnings, error) {
...
+
 	if container.InTreeModulesToRemove != nil && container.InTreeModuleToRemove != "" { //nolint:staticcheck
 		return fmt.Errorf("only one of the Container's fields: InTreeModulesToRemove or InTreeModuleToRemove can be defined")
 	}
 
+	contImg := container.ContainerImage
+	err := validateImageFormat(contImg)
+	if contImg != "" && err != nil {
+		return fmt.Errorf("failed to validate image format: %v", err)
+	}
```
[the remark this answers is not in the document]
ybettan: Fixed with a small adjustment.

ybettan:
We cannot use `name.NewTag` is it doesn't like the variable substitution that we have in our API:
```
tag can only contain the characters `abcdefghijklmnopqrstuvwxyz0123456789_-.ABCDEFGHIJKLMNOPQRSTUVWXYZ`: $KERNEL_FULL_VERSION
```

Rolling back to the manual validation of the presence of `@` or `:` in the image name.

PR #871, opened 2024-08-07
Set a default value for `setFirmwareClassPath` in the worker config.

ybettan:
When firmware is added to the driver-cotnainer there are 2 actions that were taken:

1. Copy the firmware files onto the node at `/var/lib/firmware`
2. Add `/var/lib/firmware` to `/sys/module/firmware_class/parameters/path` in order to update the kernel lookup directories for firmware files.

In order to make this feature configurable, we will allow the user to choose the path on the nodes to write the firmware files to.

In order to make it work with no additional changes, when firmware files are added to the driver-container, we added a default directory to add the firmware files to on the nodes.

To keep it simple, we will also always add this path to the kernel lookup directories.

---

/assign @yevgeny-shnaidman 
unit-tests are still failing but the code should stay the same.

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -4,10 +4,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 	"path/filepath"
 	"reflect"
 	"strings"
+
+	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/config"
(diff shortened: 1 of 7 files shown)


internal/controllers/nmc_reconciler.go
```diff
@@ -852,8 +853,12 @@ func (p *podManagerImpl) LoaderPodTemplate(ctx context.Context, nmc client.Objec
 	}
 
 	if nms.Config.Modprobe.FirmwarePath != "" {
-		args = append(args, "--"+worker.FlagFirmwareMountPath, worker.FirmwareMountPath)
-		if err = setFirmwareVolume(pod, p.workerCfg.SetFirmwareClassPath); err != nil {
+		firmwareClasspath := p.workerCfg.SetFirmwareClassPath
```
yevgeny-shnaidman, line 856: one suggestion:
lets move line 837
args = append(args, "--"+worker.FlagFirmwareClassPath, *p.workerCfg.SetFirmwareClassPath)
also into this if clause and completely remove the if clause on line 836

ybettan: Good point. Will do.

PR #885, opened 2024-08-26
Refactoring the `worker` package.

ybettan:
Moving the helper interface file containing the interface implementation rather than in the file containing the interface definition.

---

/cc @yevgeny-shnaidman


ybettan:
/close
All this code is going to be removed in a follow-up PR.

PR #886, opened 2024-08-26
Adding an init container to the worker pod.

ybettan:
This init container will be pulled by the container-runtime of the cluster and will copy all `.ko` and firmware files to a shared volume accessible by the worker container to load.

---

/assign @yevgeny-shnaidman 
/hold
Didn't test it e2e yet.

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -54,6 +54,9 @@ const (
 	nodeModulesConfigFinalizer = "kmm.node.kubernetes.io/nodemodulesconfig-reconciler"
 	volumeNameConfig           = "config"
 	workerContainerName        = "worker"
+	initContainerName          = "image-extractor"
+	sharedFilesDir             = "/tmp"
+	volNameTmp                 = "tmp"
 )
 
 //+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=nodemodulesconfigs,verbs=get;list;watch
@@ -824,7 +827,7 @@ func (p *podManagerImpl) ListWorkerPodsOnNode(ctx context.Context, nodeName stri
 }
 
 func (p *podManagerImpl) LoaderPodTemplate(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleSpec) (*v1.Pod, error) {
-	pod, err := p.baseWorkerPod(ctx, nmc.GetName(), &nms.ModuleItem, nmc)
+	pod, err := p.baseWorkerPod(ctx, nmc, &nms.ModuleItem, &nms.Config)
 	if err != nil {
 		return nil, fmt.Errorf("could not create the base Pod: %v", err)
 	}
@@ -846,6 +849,13 @@ func (p *podManagerImpl) LoaderPodTemplate(ctx context.Context, nmc client.Objec
 		}
 
 		args = append(args, "--"+worker.FlagFirmwarePath, *firmwareHostPath)
+
+		firmwarePathContainerImg := filepath.Join(nms.Config.Modprobe.FirmwarePath, "*")
+		firmwarePathWorkerImg := filepath.Join(sharedFilesDir, nms.Config.Modprobe.FirmwarePath)
+		if err = addCopyCommand(pod, firmwarePathContainerImg, firmwarePathWorkerImg); err != nil {
+			return nil, fmt.Errorf("could not add the copy command to the init container: %v", err)
+		}
+
 		if err = setFirmwareVolume(pod, firmwareHostPath); err != nil {
 			return nil, fmt.Errorf("could not map host volume needed for firmware loading: %v", err)
 		}
@@ -871,7 +881,7 @@ func (p *podManagerImpl) LoaderPodTemplate(ctx context.Context, nmc client.Objec
 }
 
 func (p *podManagerImpl) UnloaderPodTemplate(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleStatus) (*v1.Pod, error) {
-	pod, err := p.baseWorkerPod(ctx, nmc.GetName(), &nms.ModuleItem, nmc)
+	pod, err := p.baseWorkerPod(ctx, nmc, &nms.ModuleItem, &nms.Config)
 	if err != nil {
 		return nil, fmt.Errorf("could not create the base Pod: %v", err)
 	}
@@ -898,6 +908,13 @@ func (p *podManagerImpl) UnloaderPodTemplate(ctx context.Context, nmc client.Obj
 			return nil, fmt.Errorf("firmwareHostPath was not set while the Module requires firmware unloading")
 		}
 		args = append(args, "--"+worker.FlagFirmwarePath, *firmwareHostPath)
+
+		firmwarePathContainerImg := filepath.Join(nms.Config.Modprobe.FirmwarePath, "*")
+		firmwarePathWorkerImg := filepath.Join(sharedFilesDir, nms.Config.Modprobe.FirmwarePath)
+		if err = addCopyCommand(pod, firmwarePathContainerImg, firmwarePathWorkerImg); err != nil {
+			return nil, fmt.Errorf("could not add the copy command to the init container: %v", err)
+		}
+
 		if err = setFirmwareVolume(pod, firmwareHostPath); err != nil {
 			return nil, fmt.Errorf("could not map host volume needed for firmware unloading: %v", err)
 		}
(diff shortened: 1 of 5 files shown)


ybettan:
/unhold

It was tested locally a kmod requiring firmware and everything work as expected.

internal/controllers/nmc_reconciler.go
```diff
@@ -54,6 +56,9 @@ const (
 	nodeModulesConfigFinalizer = "kmm.node.kubernetes.io/nodemodulesconfig-reconciler"
 	volumeNameConfig           = "config"
 	workerContainerName        = "worker"
+	initContainerName          = "files-getter"
```
yevgeny-shnaidman, line 59: maybe "image-extractor"?

internal/controllers/nmc_reconciler.go
```diff
@@ -923,12 +934,16 @@ var (
...
-	owner client.Object,
-) (*v1.Pod, error) {
+type ContainerArg struct {
+	KmodsPath        string
+	KmodsHostPath    string
+	FirmwarePath     string
+	FirmwareHostPath string
+}
+
+func (p *podManagerImpl) baseWorkerPod(ctx context.Context, nmc client.Object, item *kmmv1beta1.ModuleItem,
```
yevgeny-shnaidman, line 944: This is an internal interface used only by NMC reconciler. Both item and moduleConfig are part of NMS object, why not just path NMS and let baseWorkerPod find whaever info it needs?

internal/controllers/nmc_reconciler.go
```diff
@@ -1138,6 +1236,7 @@ func setFirmwareVolume(pod *v1.Pod, firmwareHostPath *string) error {
 
 	pod.Spec.Volumes = append(pod.Spec.Volumes, firmwareVolume)
 	container.VolumeMounts = append(container.VolumeMounts, firmwareVolumeMount)
+
```
yevgeny-shnaidman, line 1239: nit:we can remove this

internal/controllers/nmc_reconciler.go
```diff
@@ -993,8 +1014,32 @@ func (p *podManagerImpl) baseWorkerPod(
...
 		},
+		{
+			Name:      volNameTmp,
+			MountPath: sharedFilesDir,
+			ReadOnly:  true,
+		},
 	}
 
+	const initContainerArgTmpl = `
+mkdir -p {{.KmodsHostPath}};
```
yevgeny-shnaidman, line 1025: IMHO KmodHostPath is not a very good name. We are not coping to host filesystem, but to the worker container filesystem

internal/controllers/nmc_reconciler.go
```diff
@@ -993,8 +1014,32 @@ func (p *podManagerImpl) baseWorkerPod(
...
+cp -R {{.KmodsPath}} {{.KmodsHostPath}};
```
yevgeny-shnaidman, line 1026: KmodsPath -> containerImageDirName?

internal/controllers/nmc_reconciler.go
```diff
@@ -993,8 +1014,32 @@ func (p *podManagerImpl) baseWorkerPod(
...
 			ReadOnly:  true,
 		},
+		{
+			Name:      volNameTmp,
+			MountPath: sharedFilesDir,
+			ReadOnly:  true,
+		},
 	}
 
+	const initContainerArgTmpl = `
```
yevgeny-shnaidman, line 1024: i think that suing templates in this case is an overkill: we have a very short string, which can easily be produced by Sprintf oranother simple option

internal/controllers/nmc_reconciler.go
```diff
@@ -1109,8 +1172,43 @@ func setWorkerSofdepConfig(pod *v1.Pod, modulesLoadingOrder []string) error {
...
+	container, _ := podcmd.FindContainerByName(pod, initContainerName)
+	if container == nil {
+		return errors.New("could not find the worker container")
+	}
+
+	if firmwareHostPath == nil {
+		return errors.New("hostFirmwarePath must be set")
+	}
+
+	const argTmplFirmwareAddition = `
```
yevgeny-shnaidman, line 1186: same as for other init container arguments: it is a short string that can be produced by a simple Sprintf, instead of the whole template flow. Maybe we can create a function that defines all the arguments for init container, instead of splitting between this function and baseWorkerPod?

yevgeny-shnaidman:
I think we need to verify one important issue: using kmod already defined on the host. We must verify that when we are using "cp -r" we don't break the symbolic links

internal/controllers/nmc_reconciler.go
ybettan: I agree in is not the best name. I will find something else.

internal/controllers/nmc_reconciler.go
ybettan: `containerDirName` was a good name when we were mounting the whole `containerImage` into a `DirName` which is not the case anymore. Here we are referring to the kmods path.

internal/controllers/nmc_reconciler.go
ybettan: I think that is the only way I found in order to have a clean arg with different commands in the `pod.initContainers[].arg` which isn't super long with additional `\n` in it.

internal/controllers/nmc_reconciler.go
ybettan: At first I was doing it together but I think it is cleaner when the firmware commands are added inside the firmware section rather than having another `if` block in the base pod to handle firmware addition.

internal/controllers/nmc_reconciler.go
ybettan: Sure.

internal/controllers/nmc_reconciler.go
ybettan: Because the 2 consumers are using this method with different input. One of them use a `spec` and another one the `status` so I had to go with the next larger object to contain everything I need.

internal/controllers/nmc_reconciler.go
ybettan: I have renamed all the struct fields to:
```
KmodsPathSrc   
KmodsPathDst   
FirmwarePathSrc
FirmwarePathDst
```

internal/controllers/nmc_reconciler.go
ybettan: Changed to `fmt.Sprintf`.

internal/controllers/nmc_reconciler.go
ybettan: Changed to fmt.Sprintf.

ybettan:
> I think we need to verify one important issue: using kmod already defined on the host. We must verify that when we are using "cp -r" we don't break the symbolic links

I am not sure I understand this part.

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: How about containerImagePath and WorkerImagePath? This way it becomes clear what we are talking about: copying from containerImage into the Worker Container

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: please see my comment below

internal/controllers/nmc_reconciler.go
ybettan: Fixed.

ybettan:
/hold

I still need to validate that we are not breaking Dockerfile with simlinks for `/lib/modules`

internal/controllers/nmc_reconciler.go
ybettan: Fixed.

ybettan:
> I think we need to verify one important issue: using kmod already defined on the host. We must verify that when we are using "cp -r" we don't break the symbolic links

I have tested it as the following.

1. I have created a new kmod based on:
```
FROM fedora:39 as builder

ARG KERNEL_VERSION


RUN dnf update -y && \
    dnf install -y \
    git \
    make \
    gcc

RUN dnf install -y \
    https://rpmfind.net/linux/fedora/linux/updates/39/Everything/x86_64/Packages/k/kernel-modules-${KERNEL_VERSION}.rpm \
    https://rpmfind.net/linux/fedora/linux/updates/39/Everything/x86_64/Packages/k/kernel-devel-${KERNEL_VERSION}.rpm

RUN mkdir -p /opt/lib/modules/${KERNEL_VERSION}/kernel && \
    ln -s /lib/modules/${KERNEL_VERSION}/kernel /opt/lib/modules/${KERNEL_VERSION}/kernel
```

2. I have added a `sleep` command as the last command of the init-container (and build a custom KMM image)
3. I have connected to the sleeping init-container and validated that the link remains:
```
ybettan kernel-module-management (init-container) $ oc exec -it pod/kmm-worker-minikube-kmm-ci -c image-extractor -- ls -l /opt/lib/modules/6.10.4-100.fc39.x86_64/kernel/
total 4
lrwxrwxrwx. 1 root root 42 Sep  1 18:57 kernel -> /lib/modules/6.10.4-100.fc39.x86_64/kernel
```

/unhold

PR #894, opened 2024-09-01
[WIP] Adding firmware to the `e2e` CI job.

ybettan:
In order to do so, a few steps were required:

* Add firmware to the driver .c file.
* Add a firmware binary to the container image.
* Add a `modprobe.firmwarePath` section in the `Module`
* Run `minikube` with `--container-runtime=crio` because crio gives the pod permissions write to `/sys/module/firmware_class/parameters/path`.
* Mounting the firmware lookup path from the host to minikube because the code looking for the firmware file is in the driver which runs from the kernel which is not part of the `minikube` container.

--

/assign @yevgeny-shnaidman

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -17,10 +17,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Build the image
-        run: make docker-build IMG=kmm:local
+        run: make docker-build IMG=localhost/k8s-sigs/kmm:local
 
       - name: Export the image
-        run: docker save -o kmm_local.tar kmm:local
+        run: docker save -o kmm_local.tar localhost/k8s-sigs/kmm:local
 
       - name: Upload the image
         uses: actions/upload-artifact@v4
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e.yaml
```diff
@@ -151,8 +151,16 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      #- name: Extract firmware path
```
yevgeny-shnaidman, line 154: can we remove this part, if it is commented out?

.github/workflows/e2e.yaml
```diff
@@ -151,8 +151,16 @@ jobs:
...
+      #  run: |
+      #    firmware_path=$(cat config/manager/controller_config.yaml | grep firmwareHostPath | cut -d" " -f4-)
+      #    echo "::set-output name=firmware_path::$firmware_path"
+
       - name: Create the minikube cluster
         uses: ./.github/actions/create-minikube-cluster
+        with:
+          start-args: --container-runtime=crio --mount --mount-string="/lib/firmware:/lib/firmware"
+          #start-args: --container-runtime=crio --mount --mount-string="${{ secrets.firmware_path }}:${{ secrets.firmware_path }}"
```
yevgeny-shnaidman, line 163: and this one also?

.github/workflows/e2e.yaml
ybettan: I have uncomment it. We need it in order to extract the `firmwareHostPath` from the configmap instead of hard coding it.

.github/workflows/e2e.yaml
ybettan: Updated.

ybettan:
/close

It is not easy to implement, one of the issues is that `--insecure-registry` doesn't work properly with `--container-runtime=crio`.
See https://github.com/kubernetes/minikube/issues/13932

In addition to that, having a kmod that required firmware will not let us remove `modeprobe.firmwarePath` from the module and test simpler flow as well.

PR #899, opened 2024-09-05
Using the Module's `spec.moduleLoader.container.imagePullPolicy`.

ybettan:
When we moved to the worker-pod implementation in v2 we started to pull kmod images via HTTP from teh worker-pod directly and we ignored this field.

Now that we moved to pulling using the cluster's container-runtime, the `imagePullPolicy` can be passed to the init-container definition.

---

/assign @yevgeny-shnaidman


yevgeny-shnaidman:
i think you need to update the bundle

ybettan:
> i think you need to update the bundle

There is no bundle in this repo. I have updated the manifests.

PR #901, opened 2024-09-08
Introducing FileSystemHelper interface

yevgeny-shnaidman:
This interface handles filesystem functionality used mainly by worker In the latter PRs it will be updated to extend the support of various in-tree removal scenarious. In addition it will allow for better testing of the functionality related to deleting/verifying/copying files


cmd/worker/funcs_kmod.go
```diff
@@ -17,7 +18,8 @@ func rootFuncPreRunE(cmd *cobra.Command, args []string) error {
 	logger.Info("Starting worker", "version", Version, "git commit", commit)
 
 	mr := worker.NewModprobeRunner(logger)
-	w = worker.NewWorker(mr, logger)
+	fh := utils.NewFSHelper(logger)
```
ybettan, line 21:
```suggestion
	fsh := utils.NewFSHelper(logger)
```

internal/utils/fshelper.go
```diff
@@ -0,0 +1,47 @@
...
+			fh.logger.Info("Removing dst file", "file", fileToRemove)
+			err = os.Remove(fileToRemove)
+			if err != nil {
+				fh.logger.Info(WarnString("failed to delete file"), "file", fileToRemove, "error", err)
+			}
+		}
+		return nil
+	})
+	return err
+}
```
ybettan, line 47:
```suggestion
func (fh *fsHelper) RemoveSrcFilesFromDst(srcDir, dstDir string) error {
	err := filepath.Walk(srcDir, func(path string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if !info.IsDir() {
			relPath, err := filepath.Rel(srcDir, path)
			if err != nil {
				fh.logger.Info(WarnString("failed to get relative path"), "srcDir", srcDir, "path", path, "error", err)
				return nil
			}
			fileToRemove := filepath.Join(dstDir, relPath)
			fh.logger.Info("Removing dst file", "file", fileToRemove)
			err = os.Remove(fileToRemove)
			if err != nil {
				fh.logger.Info(WarnString("failed to delete file"), "file", fileToRemove, "error", err)
			}
		}
		return nil
	})
	if err != nil {
	    return fmt.Errorf("failed to remove file %s/* from %s\n", srcDir, dstDir)
	}
	return nil
}
```

internal/utils/fshelper.go
```diff
@@ -0,0 +1,47 @@
...
+			fileToRemove := filepath.Join(dstDir, relPath)
+			fh.logger.Info("Removing dst file", "file", fileToRemove)
+			err = os.Remove(fileToRemove)
+			if err != nil {
+				fh.logger.Info(WarnString("failed to delete file"), "file", fileToRemove, "error", err)
+			}
+		}
+		return nil
+	})
+	return err
```
ybettan, line 46:
```suggestion
	if err != nil {
	    return fmt.Errorf("failed to remove file %s/* from %s\n", srcDir, dstDir)
	}
	return nil
```

internal/utils/fshelper.go
```diff
@@ -0,0 +1,47 @@
+package utils
```
ybettan, line 1: Can we have a full name for this file? `filesystemHelper.go`?

cmd/worker/funcs_kmod.go
yevgeny-shnaidman: Done

internal/utils/fshelper.go
yevgeny-shnaidman: done

internal/utils/fshelper.go
yevgeny-shnaidman: changed the name. Regarding its own package: we currently have a lot of minor interfaces/functionalities in utils. We can discuss if we want to dismantle this package in a later PRs

ybettan: OK then.

PR #903, opened 2024-09-10
support inTree removal scenario for nodes without .ko files

yevgeny-shnaidman:
Modprobe will fail with error in case it cannot find an appropriate .ko file when asked to remove a kernel module. With in-tree removal scenario, modporbe search path is /lib/modules on the host, and not the worker image.
This PR contains the following:
1) Adding FileExists function that check the presence of a file under
   a search path based on a regexp
2) in worker flow, in case in-tree removal scenario is requisted, worker
   will verify if the requested .ko file(s) are present on the host,
   remove from the request missing files and will proceed with the
   removal scenario for the rest
3) unit-test


config/manager/kustomization.yaml
```diff
@@ -9,11 +9,11 @@ patches:
...
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
+  newTag: check-intree-removal
 - name: worker
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-worker
-  newTag: latest
+  newName: quay.io/yshnaidm/kernel-module-management-worker
+  newTag: check-intree-removal
```
ybettan, line 16: Let's not push this part :)

internal/worker/worker.go
```diff
@@ -47,10 +47,26 @@ func (w *worker) LoadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, fir
...
 		w.logger.Info("Unloading in-tree modules", "names", inTreeModulesToRemove)
+		modulesToUnload := make([]string, 0, len(inTreeModulesToRemove))
+		for _, module := range inTreeModulesToRemove {
+			exists, err := w.fh.FileExists("/lib/modules", fmt.Sprintf("^%s.ko", module))
+			if err != nil {
+				w.logger.Info(utils.WarnString(fmt.Sprintf("failed to check if module file %s present on the host:", module)), "error", err)
+				continue
+			}
+			if !exists {
+				w.logger.Info("not trying to unload in-tree module, since its file is not present on the host", "file", module)
```
ybettan, line 58:
```suggestion
				w.logger.Info("not trying to unload in-tree module, since its file is not present on the host", "file", module+".ko")
```

internal/worker/worker.go
```diff
@@ -47,10 +47,26 @@ func (w *worker) LoadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, fir
 
 	if inTreeModulesToRemove != nil {
 		w.logger.Info("Unloading in-tree modules", "names", inTreeModulesToRemove)
+		modulesToUnload := make([]string, 0, len(inTreeModulesToRemove))
+		for _, module := range inTreeModulesToRemove {
+			exists, err := w.fh.FileExists("/lib/modules", fmt.Sprintf("^%s.ko", module))
+			if err != nil {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: faulty regexp, some internal filesystem access error...

config/manager/kustomization.yaml
yevgeny-shnaidman: fixed

internal/worker/worker.go
yevgeny-shnaidman: fixed, printing the module name

PR #911, opened 2024-09-22
Fixing module loading during kernel upgrade

yevgeny-shnaidman:
When the node upgrade includes kernel upgrade, NMC Spec will change to include new kernel version. In that case NMC controller will try first to create unloader worker pod. Since the unloader worker pod uses the old configuration (from status), which uses the old image, the modprobe in the worker pod will fail, since it won't find kernel module under the /opt/lib/modules/<new kernel> path
This PR fixes the issue by creating unloader pod in case of difference in spec and status of NMC only in case kernels are equal. Otherwise, it creates loader pod, since it means that node was rebooted, and the kernel module is not loaded yet, since the status contains old kernel

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -321,14 +321,23 @@ func (h *nmcReconcilerHelperImpl) ProcessModuleSpec(
 	}
 
 	if pod == nil {
+		// new module is introduced, need to load it
 		if status == nil {
 			logger.Info("Missing status; creating loader Pod")
 			return h.pm.CreateLoaderPod(ctx, nmcObj, spec)
 		}
 
+		/* configuration changed for module: if spec status contain the same kernel,
+		unload the kernel module, otherwise - load kernel modules, since the pod
+		is not running, the module cannot be loaded using the old kernel configuration
+		*/
 		if !reflect.DeepEqual(spec.Config, status.Config) {
-			logger.Info("Outdated config in status; creating unloader Pod")
-			return h.pm.CreateUnloaderPod(ctx, nmcObj, status)
+			if spec.Config.KernelVersion == status.Config.KernelVersion {
+				logger.Info("Outdated config in status; creating unloader Pod")
+				return h.pm.CreateUnloaderPod(ctx, nmcObj, status)
+			}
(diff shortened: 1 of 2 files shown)


internal/controllers/nmc_reconciler.go
```diff
@@ -321,14 +321,23 @@ func (h *nmcReconcilerHelperImpl) ProcessModuleSpec(
...
+		*/
 		if !reflect.DeepEqual(spec.Config, status.Config) {
-			logger.Info("Outdated config in status; creating unloader Pod")
-			return h.pm.CreateUnloaderPod(ctx, nmcObj, status)
+			if spec.Config.KernelVersion == status.Config.KernelVersion {
+				logger.Info("Outdated config in status; creating unloader Pod")
+				return h.pm.CreateUnloaderPod(ctx, nmcObj, status)
+			}
+			logger.Info("Outdated config in status and kernels differ, probably due to upgrade; creating loader Pod")
+			return h.pm.CreateLoaderPod(ctx, nmcObj, spec)
```
ybettan, line 340: Why do we need to create the loader pod here?

In this case, the node's reboot ready state is newer than the config status so the loader pod will be created in line 348 anyway isn't it?

yevgeny-shnaidman: I thought about that, but eventually went with this solution for a number of reasons:
1) this makes the code more readable, since the conditions are stated clearly
2) makes unit tests more readable and easier to check

ybettan: I would agree if this was a separate case. But IMO it falling exactly in the "node rebooting" option.

After all:
1. the node has rebooted
2. It has rebooted after the last update on the NMC status
3. If an unloaderPod was required, then it has already been done.

Seems like a pretty good fit to me.

yevgeny-shnaidman: That condition was added just for node reboot scenario. I agree it will work, but it will be much harder to understand and especially debug such cases.

ybettan: We discussed offline, we will refactor this function in a later PR to be in a switch-case flow which will make it clearer - in this case, it is indeed clearer to have it separate by conditions and each condition to contain all it actions.

LGTM

PR #913, opened 2024-09-25
Fixing race condition during cluster upgrade with kernel upgrade

yevgeny-shnaidman:
The following scenario causes worker pods to be stuck during cluster upgrade:
1) kernel module is loaded into the node. NMC contains both spec and
   status using the current kernel version
2) cluster upgrade starts. As part of the upgrade node becomes
   Unschedulable
3) Module-NMC removes Spec from NMC, since the node is Uschedulable 4) Node becomes schedulable
5) NMC controller tries to unload kernel module using the NMC Status
   confiration, which contains old kernel.
6) Worker unload pods get stuck in Error, since Node is running the new
   kernel
7) Module-NMC updates Spec of NMC, but since worker pod exists, nothing
   is done

Solution:
   When processing orphaned NMC statuses (status exists but spec does
not), NMC controller should ignore modules that have statuses created prior to Node's Ready timestamp

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -121,6 +121,11 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 		statusMap[status.Namespace+"/"+status.Name] = &nmcObj.Status.Modules[i]
 	}
 
+	node := v1.Node{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: nmcObj.Name}, &node); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get node %s: %v", nmcObj.Name, err)
+	}
+
 	errs := make([]error, 0, len(nmcObj.Spec.Modules)+len(nmcObj.Status.Modules))
 
 	for _, mod := range nmcObj.Spec.Modules {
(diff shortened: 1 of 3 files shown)


internal/controllers/nmc_reconciler_test.go
```diff
@@ -490,50 +510,51 @@ var _ = Describe("nmcReconcilerHelperImpl_ProcessModuleSpec", func() {
...
 		)
 	})
 
-	It("should return an error if we could not get the node", func() {
-		nmc := &kmmv1beta1.NodeModulesConfig{
-			ObjectMeta: metav1.ObjectMeta{Name: nmcName},
-		}
-
-		cfg := kmmv1beta1.ModuleConfig{ContainerImage: "some-image"}
+	/*
```
mresvanis, line 519: Do we want to keep this commented-out test?

internal/controllers/nmc_reconciler.go
```diff
@@ -157,10 +162,12 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	if err := r.helper.GarbageCollectInUseLabels(ctx, &nmcObj); err != nil {
 		errs = append(errs, fmt.Errorf("failed to GC in-use labels for NMC %s: %v", req.NamespacedName, err))
 	}
-	node := v1.Node{}
-	if err := r.client.Get(ctx, types.NamespacedName{Name: nmcObj.Name}, &node); err != nil {
-		return ctrl.Result{}, fmt.Errorf("could not get node %s: %v", nmcObj.Name, err)
-	}
+	/*
```
mresvanis, line 165: Should we remove this commented-out code?

internal/controllers/nmc_reconciler_test.go
```diff
@@ -722,14 +745,27 @@ var _ = Describe("nmcReconcilerHelperImpl_ProcessUnconfiguredModuleStatus", func
 		},
 	}
 
+	node := v1.Node{}
+
+	It("should not nothing , if the node has been rebooted/ready lately", func() {
```
mresvanis, line 750:
```suggestion
	It("should do nothing, if the node has been rebooted/ready lately", func() {
```

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: Fixed

internal/controllers/nmc_reconciler_test.go
yevgeny-shnaidman: Fixed

PR #915, opened 2024-10-06
Adding an enhancement for checking image existence using the container-runtime.

ybettan:
---

/cc @yevgeny-shnaidman @mresvanis @TomerNewman

### DIFF
--- a/docs/enhancements/0002-container-runtime.md
+++ b/docs/enhancements/0002-container-runtime.md
@@ -1,0 +1,147 @@
+# Using the container-runtime to check images existance
+
+Authors: @ybettan, @yevgeny-shnaidman
+
+## Introduction
+
+This enhancement aims at eliminating several pain points related to how KMM checks if container images exists.
+
+The current implementation of KMM is checking whether images exists by generating an HTTP[s] request to the
+container registry. It requires KMM to manually access a bunch of configurations files on the nodes such as:
+* Container images mirroring configs
+* TLS configs
+* Installed CAs
+
+In addition to that, in order to reduce the number of registy accesses, we also need to maintain a cache as
+we do for the Hub&Spoke topology.
+
+[We recently started using the cluster's container-runtime for pulling kmod images from the worker pods](https://github.com/kubernetes-sigs/kernel-module-management/pull/886) which was a great step forward to maximum automation for pulling images
+in the cluster. Once we start checking the images existance using the container-runtime as well, we will be
+at ~100% automation for handlering images - everything the cluster does when pulling images, KMM will do as well.
+
+## Goals
+
+1. Remove the need for KMM to pull the images by itself using HTTP[s] and specific
+   configuration, and instead rely on the built-in kubernetes functionality and cache.
+
+## Non-goals
+
+1. Add further control for building/signing kmod images.
+
+## Design
+
+### The new `ModuleImagesConfig` CRD
+
+The new CRD will hold information regarding the existance of images mentioned in `Module` objects.
+The cluster will contain a single `ModuleImagesConfig` per `Module` in the cluster.
+The `generation` and `observedGeneration` are used to trigger the `ModuleImagesConfig` controller in some cases.
+
+```yaml
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: ModuleImagesConfig
+metadata:
+  ...
+  ...
+spec:
+  items:
+    - image: registry.example.com/org/kmod1:kver
+      generation: 1 # default value
+    - image: registry.example.com/org/kmod2:kver
+      generation: 2
+      build:
+        ....
+      sign:
+        ...
+    - image: registry.example.com/org/kmod3:kver
+      generation: 1 # default value
+status
+  obserevedGeneration: 1
+  items:
+    - image: registry.example.com/org/kmod1:kver
+      status: Exists
+      observedGeneration: 1
+    - image: registry.example.com/org/kmod2:kver
+      status: DoesNotExist
+      observedGeneration: 1
+```
+
+### Module-NMC controller's changes
+
+* The only controller creating or modifying `ModuleImagesConfig` objects - one per `Module`
+* Update `ModuleImagesConfig`'s spec by adding/modifying/removing entries.
+* Checking `ModuleImagesConfig`'s status to determine whether the defined image exists in the image
+  repository or not.
+* Continues to create `NodeModulesConfig` objects as before.
+
+```mermaid
+flowchart TD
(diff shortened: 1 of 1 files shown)


docs/enhancements/0002-container-runtime.md
```diff
@@ -0,0 +1,122 @@
...
+In addition to that, in order to reduce the number of registy accesses, we also need to maintain a cache as
+we do for the Hub&Spoke topology.
+
+[We recently started using the cluster's container-runtime for pulling kmod images from the worker pods](https://github.com/kubernetes-sigs/kernel-module-management/pull/886) which was a great step forward to maximum automation for pulling images
+in the cluster. Once we start checking the images existance using the container-runtime as well, we will be
+at ~100% automation for handlering images - everything the cluster does when pulling images, KMM will do as well.
+
+## Goals
+
+1. Reduce the number of manuall actions KMM has to take to pull images correctly.
```
yevgeny-shnaidman, line 24:
```suggestion
1. Remove the need for KMM to pull the images by itself using http/s and specific configuration, and instead rely on the built-in kubernetes functionality and cache
```

docs/enhancements/0002-container-runtime.md
```diff
@@ -0,0 +1,122 @@
...
+    - image: registry.example.com/org/kmod1:kver
+      status: Exists
+    - image: registry.example.com/org/kmod2:kver
+      status: DoesNotExist
+```
+
+### Module-NMC controller's changes
+
+* The only controller creating `ModuleImagesConfig` objects - one per `Module`
+* Checking `ModuleImagesConfig`'s status to determine whether an `NMC` object should be created
```
yevgeny-shnaidman, line 62:
```suggestion
* Checking `ModuleImagesConfig`'s status to determine whether the defined image present/exists in the image repository or not
```

docs/enhancements/0002-container-runtime.md
```diff
@@ -0,0 +1,122 @@
...
+### ModuleImagesConfig controller (MIC) (previously called the BuildSign controller) changes
+
+* Continue to build/sign kmods as before
+* Creating pods to check if images exists and updates the `ModuleImagesConfig`'s status accordingly
+
+**The following chart only describe the new functionality added to the controller.**
+```mermaid
+flowchart TD
+    MIC((ModuleImagesConfig)) -->|Reconcile| MICC(MIC controller)
+        MICC --> |Image has not status| POD((Pod))
```
yevgeny-shnaidman, line 83: In case image has no status - NMC creates the pod and exits, it should not wait for pod completion. Once the pod is failed/completed, the reconcile loop will be kicked in again

yevgeny-shnaidman:
There is one  more scenario that we need to handle:
1) user creates a KMM module
2) Module NMC creates a MIC
3) MIC controller tries to pull the image, image is missing and the MIC status is updated accordingly
4) user creates an images in the repo
5) user triggers module's reconcilaition
6) MIC reconciliation is not triggered since MIC spec is not changed

We might consider adding some counter to the MIC spec , so that Module NMC might update it and thus trigger reconciliation.

docs/enhancements/0002-container-runtime.md
ybettan: Done.

docs/enhancements/0002-container-runtime.md
ybettan: Merged a slight variation of your suggestion.

ybettan:
> There is one more scenario that we need to handle:
> 
> user creates a KMM module
> Module NMC creates a MIC
> MIC controller tries to pull the image, image is missing and the MIC status is updated accordingly
> user creates an images in the repo
> user triggers module's reconcilaition
> MIC reconciliation is not triggered since MIC spec is not changed
> We might consider adding some counter to the MIC spec , so that Module NMC might update it and thus trigger reconciliation.

What do you think about just adding the `ImagePullPolicy` from the module to the `ModuleImagesConfig` correct entry?
We are using this `ImagePullPolicy` in the init-container so we can use the same policy for checking if the image exists

1. Checking the image existence and pulling it will be consistent
2. We don't need to handle any user race condition - if he override the image, he should use `ImagePullPolicy=always` to always pull or delete the image on the node manually before re-triggering the reconciliation.

WDYT?

docs/enhancements/0002-container-runtime.md
ybettan: I will change it in the chart.

docs/enhancements/0002-container-runtime.md
```diff
@@ -0,0 +1,147 @@
...
+    MIC((ModuleImagesConfig)) -->|Reconcile| MICC(MIC controller)
+    MICC --> |Image has no status| J1((.))
+        J1 --> |Pull pod exists| J2((.))
+            J2 --> |Succeed| USE{{Update status to 'Exists'}}
+                USE --> GENBUMP{{observedGeneration++}}
+                    GENBUMP --> DELETE{{Delete the pod}}
+                        DELETE --> DONE(((Done)))
+            J2 --> |Failed| USDE{{Update status to 'DoesNotExist'}}
+                USDE --> GENBUMP
+        J1 --> |No pull pod| POD((Pull Pod)) --> DONE
```
yevgeny-shnaidman, line 105: Pull Pod -> Create Pull Pod

ybettan: I have used different forms on purpose (circles are object) so when there is a circle it means it is an object - in this case a new object.

Actions are using a different form.

docs/enhancements/0002-container-runtime.md
```diff
@@ -0,0 +1,147 @@
...
+* The only controller creating or modifying `ModuleImagesConfig` objects - one per `Module`
+* Update `ModuleImagesConfig`'s spec by adding/modifying/removing entries.
+* Checking `ModuleImagesConfig`'s status to determine whether the defined image exists in the image
+  repository or not.
+* Continues to create `NodeModulesConfig` objects as before.
+
+```mermaid
+flowchart TD
+    MOD((Module)) -->|Reconcile| MNMCC(Module-NMC controller)
+MNMCC --> |No MIC exists for the Module| MIC((MIC)) --> DONE
```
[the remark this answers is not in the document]
ybettan: I like the simplicity of a creation (circle) of MIC as we create (circle) an NMC.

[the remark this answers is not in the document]
ybettan: Fixed.

docs/enhancements/0002-container-runtime.md
ybettan: Fixed.

PR #925, opened 2024-11-10
Fixing handling ImageRepoSecret in worker pod

yevgeny-shnaidman:
Since moving to using crio and init-container for pull kernel module container image, we need to set the ImageRepoSecret(if defined) in the worker Pod and also there is no need for mapping ImageRepoSecret and secrets of the SA into worker pod volumes (was need for crane) This PR does the following:
1) set ImageRepoSecret into the worker Pod, if defined in the KMM Module 2) remove creating volumes for SA's secret and ImaRepo secret in the
   worker pod
3) remove pullSecretHelper interface implementation as not needed 4) uni-test updates


internal/controllers/nmc_reconciler.go
```diff
@@ -1098,17 +1090,21 @@ func (p *podManagerImpl) baseWorkerPod(ctx context.Context, nmc client.Object, i
...
 			ServiceAccountName: item.ServiceAccountName,
-			Volumes:            append(volumes, psv...),
+			Volumes:            volumes,
 		},
 	}
 
-	if err = ctrl.SetControllerReference(nmc, &pod, p.scheme); err != nil {
+	if item.ImageRepoSecret != nil {
+		pod.Spec.ImagePullSecrets = []v1.LocalObjectReference{*item.ImageRepoSecret}
+	}
```
ybettan, line 1099: How about we do this part before the pod template?

Something like:
```
var imagePullSecrets []v1.LocalObjectReference
if ... {
   imagePullSecrets = append(...)
}
```
 
and then add `ImagePullSecrets: imagePullSecrets` in the pod definition?

1. I think it is clearer because you see more stuff in the pod's definition rather than additional ifs after it
2. the field is set to `omitEmtpty` so an empty value won't show up

ybettan:
I will also replace the work `crio` with `container-runtime` in the commit message as crio has nothing to do with the u/s version.

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: Fixed

yevgeny-shnaidman:
> I will also replace the work `crio` with `container-runtime` in the commit message as crio has nothing to do with the u/s version.

Fixed

PR #940, opened 2024-11-23
Add support for user defined tolerances to modules

tsprasannaa:
There is a requirement to run selected pods on a cordoned node. One such example is device driver upgrade. The procedure involves the node to be cordoned. At the same time, there should be a way to run house keeping pods that carry out the driver upgrades. These pods should be schedulable.
This PR adds support to carry the tolerances for the pods that carry out the house keeping operations. These user defined tolerances will match the taint that gets added to the nodes.

ModuleSpec will be used to carry the tolerance to the pods which will be used during pod creation


ybettan:
@tsprasannaa

Thank you for this great PR. Looks like you have thought this one through. I don't have any special comments about it.
(You do have some CI errors though)
/approve

@yevgeny-shnaidman can LGTM when he is happy with it as well.

internal/node/node.go
```diff
@@ -30,7 +30,14 @@ func NewNode(client client.Client) Node {
 	}
 }
 
-func (n *node) IsNodeSchedulable(node *v1.Node) bool {
+func (n *node) IsNodeSchedulable(node *v1.Node, tolerations []v1.Toleration) bool {
+	for _, toleration := range tolerations {
```
yevgeny-shnaidman, line 34: If the node has 2 taints, but we only added one toleration to the KMM Module, then the node will be deemed schedulable, although it should be so, no?

internal/build/pod/maker.go
```diff
@@ -134,6 +134,7 @@ func (m *maker) podSpec(mld *api.ModuleLoaderData, containerImage string, pushIm
 		RestartPolicy: v1.RestartPolicyNever,
 		Volumes:       volumes(mld.ImageRepoSecret, buildConfig),
 		NodeSelector:  selector,
+		Tolerations:   mld.Tolerations,
```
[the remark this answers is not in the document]
ybettan: I was thinking about it, but I think we do need it.

For example:
* If it needs to be built on a specific arch and we set the selector of the build pod to that node (which is currently tainted)
* In the extreme case in which all nodes are tainted simultaneously (there is no reason for us to block this option).

internal/node/node.go
ybettan: I agree, we are only satisfied (in the current PR) with one fitting taint<-->toleration while what we need is to make sure that **all** the taints are tolerated.

internal/build/pod/maker.go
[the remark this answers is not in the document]
yevgeny-shnaidman: tainting all nodes simultaneously does not make sense, it will cause a lot of other issues within the cluster.  I agree regarding the different arch scenario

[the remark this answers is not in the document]
tsprasannaa: Thanks @yevgeny-shnaidman @ybettan . As discussed in this thread, as the tolerance addition is controlled by other controllers/users, the ideal scenario would be adding taints to nodes in an orderly fashion to get the upgrades done.

internal/controllers/nmc_reconciler.go
```diff
@@ -155,6 +151,14 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 		delete(statusMap, moduleNameKey)
 	}
 
+	// removing label of loaded kmods
+	if !r.nodeAPI.IsNodeSchedulable(&node, nil) {
```
yevgeny-shnaidman, line 155: This is a problematic conditions: it means that we ignore the tolerations and the cordoned node will be always unschedulable, and that means that the code never reaches line 165, which will handle unloading of the kernel module as needed during upgrade

internal/node/node.go
tsprasannaa: @yevgeny-shnaidman I have updated the matching criteria. Please review. Thanks

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: That depends on how upgrade is managed. We support 2 flows:
1. the label on the node is changed to a new version, and in that case the unloading will be done in the ProcessModuleSpec
2. the label on the node is removed, and will be later changed (this will allows user to execute whatever maintenance on the node that is needed). In that case the spec will be removed from NMC, and this scenario is handled in the ProcessUnconfiguredModuleStatus

tsprasannaa: Thanks @yevgeny-shnaidman. In that case, let me move the API just before GarbageCollectInUseLabels(). That should help both the scenarios. Let me know if works

internal/node/node.go
```diff
@@ -32,16 +32,22 @@ func NewNode(client client.Client) Node {
...
+		for _, toleration := range tolerations {
+			if taint.Key == toleration.Key && taint.Value == toleration.Value && taint.Effect == toleration.Effect {
+				continue TAINTLOOP
+			}
+		}
 		if taint.Effect == v1.TaintEffectNoSchedule {
 			return false
 		}
 	}
 	return true
```
ybettan, line 47:
```suggestion
var toleratedTaints int
for _, taint := range node.Spec.Taints {
    for _, toleration := range tolerations {
        if toleration.ToleratesTaint(taint) {
            toleratedTaints++
            break
        }
    }
}
return toleratedTaints == len(node.Spec.Taints)
```

WDYT?


@yevgeny-shnaidman this is changing the original behavior but I believe this is a more "complete" solution.

yevgeny-shnaidman: @ybettan your code looks good, but you are missing some corner cases:
if there is no toleration exists for the taint, but taints' effect is missing or it's effect is PreferNoSchedule, then the function should return true. 
@tsprasannaa a couple of minor comments for your code:  i would prefer not to use labels (TAINTLOOP) if not really needed. In this case it can be solved by a variable in the loop. Also, you should check if taint effect NoExecute, alongside with noSchedule

tsprasannaa: @ybettan @yevgeny-shnaidman Comments taken care. Please take a look. Thanks

internal/controllers/nmc_reconciler.go
tsprasannaa: @yevgeny-shnaidman Pls review this change as well

internal/controllers/nmc_reconciler.go
```diff
@@ -169,6 +165,14 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 		}
 	}
 
+	// removing label of loaded kmods
+	if !r.nodeAPI.IsNodeSchedulable(&node, nil) {
```
yevgeny-shnaidman, line 169: We actually need to re-work the removal of the ready labels, since we cannot remove them for all the Modules, only for the ones without tolerations. But it is better if we do it in a different PR. I suggest we push this one , and then we will update the flow of the removing the ready labels. @TomerNewman will you have time to take care of this?

TomerNewman: Absolutely, after merging it I will open a new PR

PR #944, opened 2024-12-03
Removing kmod labels without appropriate tolerations

TomerNewman:
This change addresses #940.
Kmod labels are now deleted only if the kernel module does not have the appropriate tolerations for the taints on the node.

---
/cc @yevgeny-shnaidman  @ybettan

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -128,7 +128,7 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 
 	errs := make([]error, 0, len(nmcObj.Spec.Modules)+len(nmcObj.Status.Modules))
-
+	var readyLabelsToRemove []string
 	for _, mod := range nmcObj.Spec.Modules {
 		moduleNameKey := mod.Namespace + "/" + mod.Name
 
@@ -136,6 +136,7 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 
 		// skipping handling NMC spec module until node is ready
 		if !r.nodeAPI.IsNodeSchedulable(&node, mod.Config.Tolerations) {
+			readyLabelsToRemove = append(readyLabelsToRemove, utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name))
 			delete(statusMap, moduleNameKey)
 			continue
 		}
@@ -166,8 +167,8 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 
 	// removing label of loaded kmods
-	if !r.nodeAPI.IsNodeSchedulable(&node, nil) {
-		if err := r.nodeAPI.RemoveNodeReadyLabels(ctx, &node); err != nil {
+	if readyLabelsToRemove != nil {
+		if err := r.nodeAPI.UpdateLabels(ctx, &node, nil, readyLabelsToRemove); err != nil {
 			return ctrl.Result{}, fmt.Errorf("could remove node %s labels: %v", node.Name, err)
 		}
 		return ctrl.Result{}, nil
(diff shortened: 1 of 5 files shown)


internal/node/node.go
```diff
@@ -130,3 +125,28 @@ func removeLabels(node *v1.Node, labels []string) {
 		)
 	}
 }
+
+func getLabelsToRemove(node *v1.Node, nmcObj *v1beta1.NodeModulesConfig) []string {
+	var labelsToRemove []string
+	for _, module := range nmcObj.Spec.Modules {
+		toDeleteLabel := len(node.Spec.Taints) == 0
+		for _, taint := range node.Spec.Taints {
+			hasToleration := false
```
ybettan, line 134:
```suggestion
			isTolerated := false
```

internal/node/node.go
```diff
@@ -130,3 +125,28 @@ func removeLabels(node *v1.Node, labels []string) {
 		)
 	}
 }
+
+func getLabelsToRemove(node *v1.Node, nmcObj *v1beta1.NodeModulesConfig) []string {
+	var labelsToRemove []string
+	for _, module := range nmcObj.Spec.Modules {
+		toDeleteLabel := len(node.Spec.Taints) == 0
```
ybettan, line 132: Doesn't it mean that we will delete the label if the node has no taints? If the node has no taints then the module will be loaded and we don't need to remove the label.

internal/node/node.go
```diff
@@ -130,3 +125,28 @@ func removeLabels(node *v1.Node, labels []string) {
 		)
 	}
 }
+
+func getLabelsToRemove(node *v1.Node, nmcObj *v1beta1.NodeModulesConfig) []string {
+	labelsToRemove := make([]string, 0)
```
ybettan, line 130:
```suggestion
	var labelsToRemove []string
```

Can we make this instead? The difference is that with the current code and empty slice will be treated as a pointer to an empty array while the suggestion will return `nil` in this case which I think is preferable.

WDYT?

internal/node/node.go
```diff
@@ -130,3 +125,28 @@ func removeLabels(node *v1.Node, labels []string) {
...
+				if toleration.ToleratesTaint(&taint) {
+					isTolerated = true
+					break
+				}
+			}
+			if !isTolerated {
+				shouldRemoveLabel = true
+				break
+			}
+			shouldRemoveLabel = false
```
ybettan, line 145: Why do we need this line? The default value set outside the loop is already `false` isn't it?

TomerNewman: you are right, I missed this line after my previous fix

internal/node/node.go
TomerNewman: Thought about it also, if we return `nil` its like we return nothing, but if we return an empty slice, its like we are saying we did not find any labels to remove.
IDM to switch it back though but that was my thought when I changed that.

internal/controllers/nmc_reconciler.go
```diff
@@ -167,7 +167,7 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 
 	// removing label of loaded kmods
 	if !r.nodeAPI.IsNodeSchedulable(&node, nil) {
```
yevgeny-shnaidman, line 169: using IsNodeSchedulable with nil will always return true, so we can simply delete this "if" clause. But i suggest that we actually collect the modules that are not schedulable on the node at line 139, and then we just delete their labels at this line (169).  This way it also means that we don't need the whole node-taints-tolerations loop again in the RemoveNodeReadyLabels function

TomerNewman: good idea

internal/controllers/nmc_reconciler.go
```diff
@@ -128,14 +128,15 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 
 	errs := make([]error, 0, len(nmcObj.Spec.Modules)+len(nmcObj.Status.Modules))
-
+	var labelsToRemove []string
```
yevgeny-shnaidman, line 131: WDYT about changing the variable's name to readyLabelsToRemove

yevgeny-shnaidman:
LGTM by me
@ybettan please take a look , and if no further comments we can push

PR #945, opened 2024-12-04
Fixing Module CR deletion during node reboot

yevgeny-shnaidman:
In case node it rebooted(for any reason) and the Module CR is being deleted at the same period of time, the following sequence happens: 1) node becomes NotReady
2) module-nmc controller removes Spec from NMC
3) nmc controller does not schedule any unloader pod, since the node is
   not ready
4) node becomes ready
5) nmc controller does not schedule unloader pod, since after node
   reboot it ignores scheduling unloader pod, since the kernel module is
not loaded anyway
6) no unloader pod running, meaning Status in the NMC is not deleted,
   meaning Module CR cannot be finalized, since Status in NMC is present

Solution:
When nmc controller processes Modules without Spec, and the node has been rebooted, the nmc controller will delete the Status from NMC.


TomerNewman:
looks good to me. Perhaps @ybettan could provide additional insights

Issue #967, opened 2025-01-07
Modules with a  '.' in their name can not be deleted


GerrySeidman:
Modules with a  '.' in their name can not be deleted

FAILING Example with module named "test.1"
- Created Module "test.1" with the below YAML
- Deleted Module "test.1"
- KMM properly deletes the kmods on the nodes
- KMM does NOT remove the finalizer from the Module (oc delete module hangs)
    -  The finalizer is never being removed from the module object.
- NOTE: Manually (ie via 'oc edit module' ) removing the finalizer was the only way to delete the module

```
apiVersion: kmm.sigs.x-k8s.io/v1beta1
kind: Module
metadata:
  name: test.1
spec:
  moduleLoader:
    container:
      modprobe:
        moduleName: yfs
      kernelMappings:
        - regexp: '^.+\_4\.x86_64$'
          containerImage: example.com/test-module
```

Working Example with module named "test-1"
- Created Module "test-1" with the below YAML
- Deleted Module "test-1"
- KMM properly deletes the kmods on the nodes
- KMM properly deletes the module

```
apiVersion: kmm.sigs.x-k8s.io/v1beta1
kind: Module
metadata:
  name: test-1
spec:
  moduleLoader:
    container:
      modprobe:
        moduleName: yfs
      kernelMappings:
        - regexp: '^.+\_4\.x86_64$'
          containerImage: example.com/test-module
```

Note: The admission control for other object types (such as Pod) will rejects names containing a '.' in their.  This issue came up for us because a change in our scripting was generating Module's with '.'s in their name.  This would have been caught earlier if the admission controller for modules rejected it upon creation.

Proposed Solutions
- Add this to the admission controller check for Modules  (preferred)
- Properly remove the finalizer for Modules with a '.' in their names

GerrySeidman:
Quirky observation
- Daemonsets allow '.' in their names.
  -  The pods they create have '.' in their names
- Deployments allow '.' in their names.
  -  However a warning is printed when they are created
  -  The pods they create have '.' in their names
  
  The important thing for Modules is that they either fail on creation or allow created modules to be properly deleted

PR #969, opened 2025-01-08
Adding a `.` as a valid character for `Module`s' name.

ybettan:
When a `Module` is deleted, the `NMC` controller is removing the labels for this `Module` in the `NMC` object.

As long as those labels for a specific module exists, the module's finalizer will not be removed upon module deletion and the module will keep hanging forever if deleted.

When a module is created with a `.` in its name, the created label for `NMC` doesn't match the expected regexp for this label and, therefore, not removed upon module's deletion.

Adding the `.` will allow the label to match the regexp, be removed upon deletion and deleting the module gracefully.

---

/assign @yevgeny-shnaidman 
/cc @TomerNewman 
Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/967

### DIFF
--- a/internal/nmc/labels.go
+++ b/internal/nmc/labels.go
@@ -6,8 +6,8 @@ import (
 )
 
 var (
-	reConfiguredLabel = regexp.MustCompile(`^beta\.kmm\.node\.kubernetes\.io/([a-zA-Z0-9-]+)\.([a-zA-Z0-9-]+)\.module-configured$`)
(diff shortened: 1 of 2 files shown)


yevgeny-shnaidman:
looks good, but lets add unit-test that verify the new regex

ybettan:
> looks good, but lets add unit-test that verify the new regex

Good catch :)
Done.

PR #971, opened 2025-01-13
adding ModuleImagesConfig definition

yevgeny-shnaidman:
ModuleImagesConfig CR will be used by the KMM operator to execute image existance flow via kubelet. The CR will hold the image to be verified in the spec, and the state of that image( exists/notexists) in the status field


api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type ImageState string
+
+const (
+	// ImageExists means that image exists in the specified repo
+	ImageExists ImageState = "Exists"
+	// ImageNotExists means that image does not exist in the specified repo
+	ImageNotExists ImageState = "NotExists"
```
ybettan, line 29:
```suggestion
	ImageDoesNotExist ImageState = "DoesNotExist"
```

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// ModuleImagesConfigSpec describes the images of the Module whose status needs to be verified
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type ModuleImagesConfigSpec struct {
+	Images []string `json:"images"`
```
ybettan, line 36: How about we call this field `Items`? Also, it is not a slice of strings but a slice of structs (you are missing the `generation` field for each `Item`.

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type ModuleImageStatus struct {
+	// image
+	Image string `json:"image"`
+	// status of the image
+	// one of: Exists, notExists
+	Status ImageState `json:"status"`
+}
```
ybettan, line 45: We are missing the `ObservedGeneration` field in this struct.

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+// ModuleImagesConfigStatus describes the status of the images that need to be verified (defined in the spec)
+// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+// +kubebuilder:validation:Required
+type ModuleImagesConfigStatus struct {
+	ImageStatus []ModuleImageStatus `json:"imageStatus"`
```
ybettan, line 51: How about we call this `Items`? I also think we should keep the same name between the `spec` entries and the `status` entries for clarity.

api/v1beta1/moduleimagesconfig_types.go
ybettan: The inner object can be `ModuleImageSpec` to fit with the following `ModuleImageStatus` later on.

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+
+// ModuleImagesConfig keeps the request for status and status of the images for KMM Module.
```
ybettan, line 57:
```suggestion
// ModuleImagesConfig keeps the record of the images status for a KMM Module.
```

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+// +kubebuilder:resource:path=moduleimagesconfigs,scope=Cluster
```
ybettan, line 58:
```suggestion
// +kubebuilder:resource:path=moduleimagesconfigs,scope=Namespaced
```
This is working at the `Module`'s level which is namespaced and not at the `nmc` level which is more related to the node that doesn't belong to any namespace.

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+// +kubebuilder:resource:path=moduleimagesconfigs,scope=Cluster,shortName=mic
```
ybettan, line 59: Can we just add the `shortName=mic` to the first kubebuilder annotation and remove this line?

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -0,0 +1,83 @@
...
+// +operator-sdk:csv:customresourcedefinitions:displayName="Node Modules Config"
```
ybettan, line 60:
```suggestion
// +operator-sdk:csv:customresourcedefinitions:displayName="Module Images Config"
```

api/v1beta1/moduleimagesconfig_types.go
ybettan: We can also go with `ImageItem`

api/v1beta1/moduleimagesconfig_types.go
yevgeny-shnaidman: Fixed

api/v1beta1/moduleimagesconfig_types.go
yevgeny-shnaidman: fixed

api/v1beta1/moduleimagesconfig_types.go
yevgeny-shnaidman: Fixed

api/v1beta1/moduleimagesconfig_types.go
ybettan: It is not fixed.

api/v1beta1/moduleimagesconfig_types.go
ybettan: I think that `ModuleImageConfig` is too similar to `ModuleImagesConfig`.

api/v1beta1/moduleimagesconfig_types.go
ybettan: Looks like you didn't push your changes.

api/v1beta1/moduleimagesconfig_types.go
yevgeny-shnaidman: Fixed

PR #976, opened 2025-01-19
adding field needed for build and sign flow the MIC object

yevgeny-shnaidman:
MIC reconciler is watching the MIC object. it should receive all the needed info for its action from the MIC, it does not need to access the Module object at all


api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -35,6 +36,18 @@ type ModuleImageSpec struct {
...
+	// +optional
+	Build *Build `json:"build,omitempty"`
+
+	// Sign contains sign instructions, in case image needs signing
+	// +optional
+	Sign *Sign `json:"sign,omitempty"`
+
+	// ImageRepoSecret contains pull secret for the image's repo, if needed
+	// +optional
+	ImageRepoSecret *v1.LocalObjectReference `json:"imageRepoSecret,omitempty"`
```
ybettan, line 50: Let's move this field to the `ModuleImagesConfigSpec`? It is one per module anyway. I don't think we should replicate it in every entry.

PR #980, opened 2025-01-21
Removing cache package and image translation from Hub&Spoke flow

TomerNewman:
Now that we are using the cluster's container-runtime to pull images, maintaining our own cache is redundant since the container runtime is using its own cache on the nodes.

In addition to that, we had some issue invalidating cache entries in some cases, therefore and since we don't need it anymore, removing it completely simplifies the flow.

We should use the images as mentioned in the ManagedClusterModule.
There is no need for additional manipulations to it.

--- 


ybettan:
First comment is about the commit message :)

I would change it to:

> Now that we are using the cluster's container-runtime to pull images, maintaining our own cache is redundant since the container runtime is using its own cache on the nodes.

In addition to that, we had some issue invalidating cache entries in some cases, therefore and since we don't need it anymore, removing it completely simplifies the flow.

We should use the images as mentioned in the `ManagedClusterModule`. The is no need for additional manipulations to it.

Reason:
CRIO, ICSPs, IDMS, ITMS are Openshift objects, the code in this repo is for vanilla k8s only.

TomerNewman:
> First comment is about the commit message :)
> 
> I would change it to:
> 
> > Now that we are using the cluster's container-runtime to pull images, maintaining our own cache is redundant since the container runtime is using its own cache on the nodes.
> > In addition to that, we had some issue invalidating cache entries in some cases, therefore and since we don't need it anymore, removing it completely simplifies the flow.
> > We should use the images as mentioned in the `ManagedClusterModule`. The is no need for additional manipulations to it.
> 
> Reason: CRIO, ICSPs, IDMS, ITMS are Openshift objects, the code in this repo is for vanilla k8s only.

Absolutely right, fixed that + PR description

install.sh
```diff
@@ -0,0 +1,196 @@
+# Copyright Contributors to the Open Cluster Management project
```
ybettan, line 1: What is this script for?

internal/manifestwork/manifestwork_test.go
```diff
@@ -335,7 +254,7 @@ var _ = Describe("SetManifestWorkAsDesired", func() {
 					KernelMappings: []kmmv1beta1.KernelMapping{
 						{
 							Literal:        kernelVersion,
-							ContainerImage: imageName + "@" + digest,
+							ContainerImage: imageName,
```
ybettan, line 257:
```suggestion
							ContainerImage: imageName + ":tag",
```

internal/manifestwork/manifestwork.go
```diff
@@ -233,68 +228,16 @@ func (mwg *manifestWorkGenerator) managedClusterKernelMappings(
 		)
 
 		mapping := mwg.mappingFromModuleLoaderData(mld)
```
yevgeny-shnaidman, line 230: I am missing something: why are we removing the translation from tag to digest?

TomerNewman: > I am missing something: why are we removing the translation from tag to digest?

It seems that the only reason we translated is because Openshift used ICSP as image-mirror objects on disconnected env.
Now we have `ImageDigestMirrorSet` and `ImageTagMirrorSet` resources so we do not need the translation anymore.

ybettan: Exactly, the translation was an constrain of the disconnected flow, this constrain doesn't exist anymore so we don't need the additional code that also add bugs like invalidation of cache entries in time.

PR #981, opened 2025-01-21
adding ModuleBuildSignConfig definition

yevgeny-shnaidman:


api/v1beta1/modulebuildsignconfig_types.go
```diff
@@ -0,0 +1,52 @@
...
+
+// ModuleBuildSignConfig keeps the request for images' build/sign for a KMM Module.
+// +kubebuilder:resource:path=modulebuildsignconfig,scope=Namespaced,shortName=mbsc
+// +operator-sdk:csv:customresourcedefinitions:displayName="Module Build Sign Config"
+type ModuleBuildSignConfig struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ModuleImagesConfigSpec   `json:"spec,omitempty"`
+	Status ModuleImagesConfigStatus `json:"status,omitempty"`
```
ybettan, line 35: If we do this we should add additional image states like `BuildSucceed` and `BuildFailed`.

PR #984, opened 2025-01-23
Adding mbsc internal package

yevgeny-shnaidman:
This package is going to handle the manipulations on the MBSC objects The current PR introduces just one function, which will be used by the MIC controller and unit-test.


internal/mbsc/helper.go
```diff
@@ -0,0 +1,30 @@
+package mbsc
+
+import (
+	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	v1 "k8s.io/api/core/v1"
+)
+
+//go:generate mockgen -source=helper.go -package=mbsc -destination=mock_helper.go
+
+type Helper interface {
```
ybettan, line 10: What do you think about using `ModuleBuildSignConfigAPI` istead? and changing the file name to `mbsc.go`?
I know we are using `helper` for NMC but IMO it is not a great name. 

I am also using the same convention in the MIC package: https://github.com/ybettan/kernel-module-management/blob/37acf32101e6da5bcf71f49610d7e728880de930/internal/mic/mic.go#L19

internal/mbsc/helper.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+type helper struct {
+}
+
+func NewHelper() Helper {
+	return &helper{}
+}
+
+func (h *helper) SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, micImageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference) {
```
ybettan, line 21:
```suggestion
func (h *helper) SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, moduleImageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference) {
```

internal/mbsc/helper.go
```diff
@@ -0,0 +1,30 @@
...
+	mbscObj.Spec.ImageRepoSecret = repoSecret
+	for i, imageSpec := range mbscObj.Spec.Images {
+		if imageSpec.Image == micImageSpec.Image {
+			mbscObj.Spec.Images[i] = *micImageSpec
```
ybettan, line 25: This code is a bit wired to me
1. It is possible that the slice underlying array was created with a default capacity and by accessing it at index `i` - we can access a null ref isn't it?
2. Why do we need line 29 if we already set the image?
3. Why we don't just append it here and return?

internal/mbsc/helper.go
```diff
@@ -0,0 +1,30 @@
...
+
+type helper struct {
+}
+
+func NewHelper() Helper {
+	return &helper{}
+}
+
+func (h *helper) SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, micImageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference) {
+	mbscObj.Spec.ImageRepoSecret = repoSecret
```
ybettan, line 22: Are we going to update the `imageRepoSecret` on each image update? The secret is generic while the specific image isn't.

Should we get a list of `ModuleImageSpec` instead of a single `ModuleImageSpec` to this method?

PR #987, opened 2025-01-23
Adding mbsc internal package

yevgeny-shnaidman:
This package is going to handle the manipulations on the MBSC objects The current PR introduces just one function, which will be used by the MIC controller and unit-test.


internal/mbsc/mbsc.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+//go:generate mockgen -source=mbsc.go -package=mbsc -destination=mock_mbsc.go
+
+type MBSC interface {
+	SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, imageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference)
+}
+
+type mbsc struct {
+}
```
ybettan, line 15:
```suggestion
type mbsc struct {}
```

internal/mbsc/mbsc.go
```diff
@@ -0,0 +1,30 @@
...
+
+func NewMBSC() MBSC {
+	return &mbsc{}
+}
+
+func (m *mbsc) SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, micImageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference) {
```
ybettan, line 21:
```suggestion
func (m *mbsc) SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, moduleImageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference) {
```

internal/mbsc/mbsc.go
```diff
@@ -0,0 +1,30 @@
...
+	mbscObj.Spec.ImageRepoSecret = repoSecret
+	for i, imageSpec := range mbscObj.Spec.Images {
+		if imageSpec.Image == micImageSpec.Image {
+			mbscObj.Spec.Images[i] = *micImageSpec
```
ybettan, line 25: I don't completely understand this code:

It is possible that the slice underlying array was created with a default capacity and by accessing it at index i - we can access a null ref isn't it?
Why do we need line 29 if we already set the image?
Why we don't just append it here and return?

internal/mbsc/mbsc.go
yevgeny-shnaidman: fixed

internal/mbsc/mbsc.go
yevgeny-shnaidman: fixed

internal/mbsc/mbsc.go
yevgeny-shnaidman: we are going over the "range of slice , which means that we are "in the bound" of the allocated array, so  no chance for nul ref.
We cannot just append the imageSpec: if the spec for  that image already exists, we need to update. If it does not exist  - we append. Otherwise we will get 2 spec entries for the same image

internal/mbsc/mbsc.go
```diff
@@ -0,0 +1,30 @@
...
+
+type mbsc struct {
+}
+
+func NewMBSC() MBSC {
+	return &mbsc{}
+}
+
+func (m *mbsc) SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, micImageSpec *kmmv1beta1.ModuleImageSpec, repoSecret *v1.LocalObjectReference) {
+	mbscObj.Spec.ImageRepoSecret = repoSecret
```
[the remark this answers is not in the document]
yevgeny-shnaidman: This function is called per image spec, no need for any list. We just update the structure, using list will just complicate the code. We are updating imageRepoSecret, since we don't know if it changed or not, so we just update it, since it does not cost us anything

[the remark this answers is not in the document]
ybettan: If you don't like sending a list of images then I would split this method into 2 methods:
* One for setting MIC that is called per image
* One for setting the repoSecret once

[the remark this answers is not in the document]
yevgeny-shnaidman: this function is called from the loop that goes over the images in MIC and prepares the MBSC for each one of those images. So we call this function every time. It does not involve communicating with API server, it just updates the structure in memory. If we will see that there is a need for updating the repoSecret only, then we can add that specific function at later stage to the interface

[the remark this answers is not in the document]
ybettan: But this method is doing something unexpected. It is called `SetModuleImageSpec` but it also update an unrelated field which is the `imageRepoSecret`, therefore, I think it should be split into 2 different methods (and both of them called in the loop if needed).

The method from updating the secret should be called outside of the loop and the one for updating the moduleImageSpec to be called from inside the loop IMO.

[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

PR #990, opened 2025-01-26
Adding validation to module toleration effect

TomerNewman:
Users can add a toleration to Module with invalid effects (for example "blabla").
This commit adds a webhook validation to the toleration effect value, ensuring only:
`NoExecute`,  `NoSchedule`, `PreferNoSchedule` effects can be applied.

---


internal/webhook/module.go
```diff
@@ -207,3 +212,14 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
 
 	return nil
 }
+func validateModuleTolerarations(mod *kmmv1beta1.Module) error {
+	for _, toleration := range mod.Spec.Tolerations {
+		switch toleration.Effect {
+		case corev1.TaintEffectNoSchedule, corev1.TaintEffectNoExecute, corev1.TaintEffectPreferNoSchedule:
+			continue
+		default:
+			return fmt.Errorf("invalid toleration effect: %s. allowed values are NoSchedule, NoExecute, PreferNoSchedule", toleration.Effect)
```
ybettan, line 221:
```suggestion
			return fmt.Errorf("invalid toleration effect %s. allowed values are NoSchedule, NoExecute, PreferNoSchedule", toleration.Effect)
```

We use `:` to wrap errors, having `:` in the string error will let us think that everything before the `:` is the wrapping error and everything after is the wrapped error.

PR #993, opened 2025-01-27
Generating `ModuleConfigImages` objects during `Module` reconciliation.

ybettan:
The `Module` reconciler will now generate and update
`ModuleImagesConfig`s objects in the cluster.

At this point, those objects aren't reconciled by any controller but
they will in the upcoming commits. For this reason, the `Module` controller
isn't checking the `MIC`'s status but instead keep checking if images exists
using direct HTTP calls.

---

/assign @yevgeny-shnaidman 

Here is the `ModuleImagesConfig` created for a simple `Module` on a minikube cluster:
```
ybettan kernel-module-management (module-reconciler) $  oc get mic/kmm-ci -o yaml | yq
apiVersion: kmm.sigs.x-k8s.io/v1beta1
kind: ModuleImagesConfig
metadata:
  creationTimestamp: "2025-02-06T14:19:13Z"
  generation: 1
  name: kmm-ci
  namespace: default
  ownerReferences:
    - apiVersion: kmm.sigs.x-k8s.io/v1beta1
      kind: Module
      name: kmm-ci
      uid: 3d1efaf1-5389-4ecd-bc3d-89ff94bb7bec
  resourceVersion: "1207250"
  uid: 4b887fe7-2115-4442-82f8-bacfb554e5d9
spec:
  images:
    - image: quay.io/ybettan/kmm-kmod:6.11.9-100.fc39.x86_64
  regeneration: 0
```

PR #995, opened 2025-01-29
Adding the `mic` package.

ybettan:
This package will be used by the `Module` reconciler in the following
commits.

---

/assign @yevgeny-shnaidman


internal/mic/mic.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create or patch %s/%s: %v", ns, name, err)
+	}
+
+	logger.Info("Applied MIC", "name", name, "namespace", ns, "result", opRes)
+
+	return nil
+}
+
+func (mica *moduleImagesConfigAPI) getGeneration(ctx context.Context, name, ns string) (int64, error) {
```
yevgeny-shnaidman, line 73: Why do we need this function? we no longer care about generation

internal/mic/mic.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+//go:generate mockgen -source=mic.go -package=mic -destination=mock_mic.go
+
+type ModuleImagesConfigAPI interface {
+	HandleModuleImagesConfig(ctx context.Context, name, ns string, images []kmmv1beta1.ModuleImageSpec,
```
yevgeny-shnaidman, line 21: I think that it is better to have a function that updates one entry in the in-memory MIc object, and let the caller handle CreatOrPatch as needed.

yevgeny-shnaidman:
General comment: the commit should be changed to something like:
adding mic package etc'/ 
"ModuleImageConfig API" is confusing, at first i thought we were talking about CRD

internal/mic/mic.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+//go:generate mockgen -source=mic.go -package=mic -destination=mock_mic.go
+
+type ModuleImagesConfigAPI interface {
```
yevgeny-shnaidman, line 20: what do you say we changed it to 
type MIC interface

ybettan: We have a lot of interfaces with the API suffix. What about `MICApi`?

yevgeny-shnaidman: API should be all in capital letters and so is MIC, but MICAPI is ugly. I think just MIC should be enough.

ybettan: This was exactly my reasoning when picking the name :joy: I will change it to `MIC` as you suggested.

ybettan:
> General comment: the commit should be changed to something like:
> adding mic package etc'/
> "ModuleImageConfig API" is confusing, at first i thought we were talking about CRD

Fixed the commit message and the PR title+description.

internal/mic/mic.go
ybettan: We indeed agreed that we will remove the generation field and the future MIC controller will continue to pull images with an exponential waiting time (with a max waiting time of a few minutes) as long as the status is `ImageDoesNotExist` and there is no build section in the MIC object.

Will update the CRD in a separate PR.

ybettan: https://github.com/kubernetes-sigs/kernel-module-management/pull/996

internal/mic/mic.go
ybettan: I think differently on this one.

I think those internal packages should be as smart as possible **as long as they stay in their scope** to reduce as max as possible responsibility from the controllers. Those controllers are getting very complicated very fast (and then we start writing helper interfaces for them).

In this case, the `MIC` interface is responsible for everything related to `ModuleImagesConfig` objects, therefore, updating the object is in its scope and it reduce a lot of functionality from the `Module` controller which should not care about how `ModuleImagesConfig`s are "internally" handled.

ybettan:
/hold

Should be rebased on top of https://github.com/kubernetes-sigs/kernel-module-management/pull/996

internal/mic/mic.go
yevgeny-shnaidman: The controllers usually want to update the images one at a time, since they are looping around the input and decide upon each image. Better interface would be  for controller to say: this is the image i want to query about, and let the package prepare the needed structure and find the correct entry.  Controller will look like:
loop over input:
    add image to mic if needed
commit changes into etcd

internal/mic/mic.go
ybettan: Rebased.

internal/mic/mic.go
ybettan: IBut how does the controller know what image has changed (or if what was changed is metadata/imageRepoSecret)?

This is how I envision the `Module` controller: https://github.com/kubernetes-sigs/kernel-module-management/pull/993/files#diff-3e13aac4e240cd59f1a0bd375f519a14e5846cc38980d3b5018f87e133a6b56fR111-R122

PR #997, opened 2025-02-04
Adding validation to module toleration operator value

TomerNewman:
Users can add a toleration to Module with invalid operator values (for example "invalid123").
This commit adds a webhook validation to the toleration operator value, ensuring only: `Exists` and `Equal` operator values can be applied.

---


internal/webhook/module.go
```diff
@@ -212,8 +212,13 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
 
 	return nil
 }
-func validateModuleTolerarations(mod *kmmv1beta1.Module) error {
+func validateModuleTolerations(mod *kmmv1beta1.Module) error {
 	for _, toleration := range mod.Spec.Tolerations {
+
```
yevgeny-shnaidman, line 217: nit: just to make it a more consistent code: we are using "if" to check the operator and "switch" to check the Effect... WDYT about using "if" for both?

TomerNewman: yeah sure

internal/webhook/module.go
```diff
@@ -212,12 +212,16 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
...
-func validateModuleTolerarations(mod *kmmv1beta1.Module) error {
+func validateModuleTolerations(mod *kmmv1beta1.Module) error {
 	for _, toleration := range mod.Spec.Tolerations {
-		switch toleration.Effect {
-		case corev1.TaintEffectNoSchedule, corev1.TaintEffectNoExecute, corev1.TaintEffectPreferNoSchedule:
-			continue
-		default:
+
+		if toleration.Operator != corev1.TolerationOpExists && toleration.Operator != corev1.TolerationOpEqual {
+			return fmt.Errorf("invalid toleration operator %s. allowed values are Exists, Equal", toleration.Operator)
```
ybettan, line 219: Can we try to find an error message without a `.` in it? When we see a full error in the code that has be wrapped multiple time it is easier when each error message is a single, short error message.

internal/webhook/module.go
```diff
@@ -212,12 +212,16 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
...
+		}
+
+		if toleration.Effect != corev1.TaintEffectNoSchedule &&
+			toleration.Effect != corev1.TaintEffectNoExecute &&
+			toleration.Effect != corev1.TaintEffectPreferNoSchedule {
 			return fmt.Errorf("invalid toleration effect %s. allowed values are NoSchedule, NoExecute, PreferNoSchedule", toleration.Effect)
```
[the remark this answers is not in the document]
TomerNewman: same as above ^

PR #999, opened 2025-02-05
Adding new functions to MIC interface

yevgeny-shnaidman:
Adding 2 function to the interface, that will be later used by controllers, among then MIC controller

### DIFF
--- a/internal/mic/mic.go
+++ b/internal/mic/mic.go
@@ -18,6 +18,9 @@ import (
 type MIC interface {
 	ApplyMIC(ctx context.Context, name, ns string, images []kmmv1beta1.ModuleImageSpec,
 		imageRepoSecret *v1.LocalObjectReference, owner metav1.Object) error
+	GetModuleImageSpec(micObj *kmmv1beta1.ModuleImagesConfig, image string) *kmmv1beta1.ModuleImageSpec
+	SetImageStatus(micObj *kmmv1beta1.ModuleImagesConfig, image string, status kmmv1beta1.ImageState)
+	GetImageState(micObj *kmmv1beta1.ModuleImagesConfig, image string) kmmv1beta1.ImageState
 }
 
 type micImpl struct {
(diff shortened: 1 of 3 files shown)


internal/mic/mic.go
```diff
@@ -61,3 +64,35 @@ func (mici *micImpl) ApplyMIC(ctx context.Context, name, ns string, images []kmm
 
 	return nil
 }
+
+func (mici *micImpl) GetModuleImageSpec(micObj *kmmv1beta1.ModuleImagesConfig, image string) *kmmv1beta1.ModuleImageSpec {
+	for i, imageSpec := range micObj.Spec.Images {
+		if imageSpec.Image == image {
+			return &micObj.Spec.Images[i]
+		}
+	}
```
ybettan, line 73:
```suggestion
	for _, imageSpec := range micObj.Spec.Images {
		if imageSpec.Image == image {
			return &imageSpec
		}
	}
```

internal/mic/mic.go
```diff
@@ -61,3 +64,35 @@ func (mici *micImpl) ApplyMIC(ctx context.Context, name, ns string, images []kmm
 
 	return nil
 }
+
+func (mici *micImpl) GetModuleImageSpec(micObj *kmmv1beta1.ModuleImagesConfig, image string) *kmmv1beta1.ModuleImageSpec {
+	for i, imageSpec := range micObj.Spec.Images {
+		if imageSpec.Image == image {
+			return &micObj.Spec.Images[i]
```
ybettan, line 71: Do we want to return the actual `ModuleImageSpec` (to be updated outside of this package) or just a copy of it?
If it is the later I would go with something like:
```
for _, imageSpec := range micObj.Spec.Images {
		if imageSpec.Image == image {
			return &imageSpec
                }
}
```

internal/mic/mic.go
```diff
@@ -61,3 +64,35 @@ func (mici *micImpl) ApplyMIC(ctx context.Context, name, ns string, images []kmm
 
 	return nil
 }
+
+func (mici *micImpl) GetModuleImageSpec(micObj *kmmv1beta1.ModuleImagesConfig, image string) *kmmv1beta1.ModuleImageSpec {
```
ybettan, line 68: What does `micObj` needs to be a pointer? Looks like we are not modifying it in the method.

ybettan, line 68: What does `micObj` needs to be a pointer? Looks like we are not modifying it in the method.

internal/mic/mic.go
yevgeny-shnaidman: Actually, since we only use imageSpec for field inspection, i am switching to your suggestion

internal/mic/mic.go
ybettan: I won't say it is always better. It is a trade of between spending time on copying the object and having more control making sure no one is updating an object it shouldn't had.

Anyway, I am OK with keeping it as a pointer.

PR #1001, opened 2025-02-06
[WIP] Module reconciler refactor

ybettan:
---

/cc @yevgeny-shnaidman

### DIFF
--- a/internal/controllers/module_reconciler.go
+++ b/internal/controllers/module_reconciler.go
@@ -104,32 +104,10 @@ func (mr *ModuleReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Modul
 		return ctrl.Result{}, fmt.Errorf("failed to handle MIC: %v", err)
 	}
 
-	currentNMCs, err := mr.reconHelper.getNMCsByModuleSet(ctx, mod)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to get NMCs for Module %s/%s: %v", mod.Namespace, mod.Name, err)
+	if err := mr.reconHelper.handleNMC(ctx, mod, targetedNodes); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to handle NMC: %v", err)
 	}
 
-	sdMap, prepareErrs := mr.reconHelper.prepareSchedulingData(ctx, mod, targetedNodes, currentNMCs)
-	errs := make([]error, 0, len(sdMap)+1)
-	errs = append(errs, prepareErrs...)
-
-	for nodeName, sd := range sdMap {
-		if sd.action == actionAdd {
-			err = mr.reconHelper.enableModuleOnNode(ctx, sd.mld, sd.node)
-		}
-		if sd.action == actionDelete {
-			err = mr.reconHelper.disableModuleOnNode(ctx, mod.Namespace, mod.Name, nodeName)
-		}
-		errs = append(errs, err)
-	}
-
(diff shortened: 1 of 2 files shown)


internal/controllers/module_reconciler.go
```diff
@@ -75,17 +81,8 @@ func (mr *ModuleReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Modul
...
-		if err := mr.nsLabeler.tryRemovingLabel(ctx, mod.Namespace, mod.Name); err != nil {
-			return ctrl.Result{}, fmt.Errorf("error while trying to remove the label on namespace %s: %v", mod.Namespace, err)
-		}
-
-		err := mr.reconHelper.finalizeModule(ctx, mod)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to finalize Module %s/%s: %v", mod.Namespace, mod.Name, err)
-		}
-		return ctrl.Result{}, nil
+	if err := mr.reconHelper.handleModuleInDeletionProcess(ctx, mod); err != nil {
```
yevgeny-shnaidman, line 84: i think that handleModuleInDeletionProcess should not check if the mod is getting deleted. The "if should stay in Reconciler, and the handling of "if" should be done in handleModuleInDeletionProcess

internal/controllers/module_reconciler.go
```diff
@@ -103,32 +100,14 @@ func (mr *ModuleReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Modul
 		return ctrl.Result{}, fmt.Errorf("failed to get list of nodes by selector: %v", err)
 	}
 
-	currentNMCs, err := mr.reconHelper.getNMCsByModuleSet(ctx, mod)
```
yevgeny-shnaidman, line 106: i would leave getNMCsByModuleSet and prepareSchedulingData in the reconciler.  Reconciler should be clear, but not empty. In addition gathering all those functions in one function will make unit-testing harder

ybettan: > i would leave getNMCsByModuleSet and prepareSchedulingData in the reconciler

But those 2 function are only used for NMC handling and not for anything else.

> In addition gathering all those functions in one function will make unit-testing harder

On the contrary, they make them easier because each function is smaller, and all sub-fucntions that were previously used are still used so no complexity is added (it is just breaking the controller logic into multiple function but all the rest stays the same.

internal/controllers/module_reconciler.go
ybettan: And if the method is renamed to `handleModuleInDeletionProcessIfNeeded`? The fact that the comment saying `the module is being deleted` is there is because the check of the timestamp isn't clear. Putting it in a method that its name describe the logic is a good wrapping IMO.

internal/controllers/module_reconciler.go
yevgeny-shnaidman: once you unite those functions in one function, your unit-test needs to test all those functions together, since they are in the same interface. there is nothing wrong in Reconciler containing code, as long as it is clear. Having those functions in the Reconciler allows others to look at the reconciler function and understand the flow, without the need to descend into all those helper functions

ybettan:
* `handleMIC` was implemented in another PR
* `handleNMC` will use other methods from the same interface which will make unit-test harder - ditched it for now
* `handleModuleInDeletionProcess` is also using other methods from the same interface, it requires the creation of a `Module` interface, moving the `finalizeModule` method to it and then use it - ditching for now as it is a big change.

PR #1002, opened 2025-02-06
Creating MIC controller

yevgeny-shnaidman:
Initial PR for MIc controller. The flow:
1. check the status of pull pods
2. check the status of mbsc
3. create MBSC or pull pods based on the MIC Spec and the statuses collected in steps 1 and 2

The MIC deletion flow is not handled in the PR, will be handled later


internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return res, fmt.Errorf("failed to get the image pods for mic %s: %v", micObj.Name, err)
+	}
+
+	err = r.micReconHelper.updateStatusByPullPods(ctx, micObj, pods)
+	if err != nil {
+		return res, fmt.Errorf("failed tp update the status for MIC %s based on pull pods: %v", micObj.Name, err)
+	}
+
+	// update the status of the current images in the Spec
```
ybettan, line 88: Do we need this comment? Line 83 also update the status of the current image in the spec

ybettan:
> The MIC deletion flow is not handled in the PR, will be handled later

Why does it has to be handled here. It is only supposed to be deleted by the `Module` reconciler once a `Module` is being deleted isn't it?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+		phase := p.Status.Phase
+		switch phase {
+		case v1.PodFailed:
+			if imageSpec.Build != nil || imageSpec.Sign != nil {
+				logger.Info("failed pod with build or sign spec, updating image status to NeedsBulding")
+				mrhi.micHelper.SetImageStatus(micObj, image, kmmv1beta1.ImageNeedsBuilding)
+			} else {
+				logger.Info(utils.WarnString("failed pod without build or sign spec, shoud not have happened"), "image", image)
+			}
+			podsToDelete = append(podsToDelete, p)
```
ybettan, line 156: Don't we want to keep the pod for debugging?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+		case v1.PodSucceeded:
+			logger.Info("successful pod, updating image status to ImageExists")
+			mrhi.micHelper.SetImageStatus(micObj, image, kmmv1beta1.ImageExists)
+			podsToDelete = append(podsToDelete, p)
+		}
+	}
+	// patch the status in the MIC object
+	err := mrhi.client.Status().Patch(ctx, micObj, patchFrom)
```
ybettan, line 164: What are we patching here? We didn't make any update to the object.

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to patch the status of mic %s: %v", micObj.Name, err)
+	}
+
+	errs := make([]error, 0, len(podsToDelete))
+	for _, pod := range podsToDelete {
+		err = mrhi.podHelper.deletePod(ctx, &pod)
```
ybettan, line 171: If we are deleting each pod separately in a different client call, then why don't we directly delete each pod instead of appending appending it to the list of pods to be deleted?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	if mbsc == nil {
+		return nil
+	}
+
+	patchFrom := client.MergeFrom(micObj.DeepCopy())
+	for _, imageState := range mbsc.Status.ImagesStates {
+		imageSpec := mrhi.micHelper.GetModuleImageSpec(micObj, imageState.Image)
+		if imageSpec == nil {
+			// image not found in spec, ignore
+			continue
```
ybettan, line 193: Don't we need to delete its status if the spec is gone?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+		// image State does not exists either new image or pull pod is still running
+		switch imageState {
+		case "":
+			// image State is not set: either new image or pull pod is still running
+			if mrhi.podHelper.getPullPodForImage(pullPods, imageSpec.Image) == nil {
+				// no pull pod- create it, otherwise we wait for it to finish
+				err := mrhi.podHelper.createPullPod(ctx, &imageSpec, micObj)
+				errs = append(errs, err)
+			}
+		case kmmv1beta1.ImageDoesNotExist:
```
ybettan, line 219: I though we decided to to set the status to `ImageDoesNotExist` in this case. Didn't we say that if the pull pod has failed but there is a build/sign section we just set the status to `NeedBuilding`?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+			if imageSpec.Build == nil && imageSpec.Sign == nil {
+				break
+			}
+			fallthrough
```
ybettan, line 223: I though we decided to to set the status to ImageDoesNotExist in this case. Didn't we say that if the pull pod has failed but there is a build/sign section we just set the status to NeedBuilding?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+		case kmmv1beta1.ImageNeedsBuilding:
+			// image needs to be built - patching MBSC
+			mbscObj := &kmmv1beta1.ModuleBuildSignConfig{
+				ObjectMeta: metav1.ObjectMeta{Name: micObj.Name, Namespace: micObj.Namespace},
+			}
+			_, err := controllerutil.CreateOrPatch(ctx, mrhi.client, mbscObj, func() error {
+				mrhi.mbscHelper.SetModuleImageSpec(mbscObj, &imageSpec)
+				mbscObj.Spec.ImageRepoSecret = micObj.Spec.ImageRepoSecret
+				return controllerutil.SetOwnerReference(micObj, mbscObj, mrhi.scheme)
+			})
```
ybettan, line 233: How about we move all of this to the MBSC interface since we already have this package and this logic doesn't contain any logic outside of the MBSC scope?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	return errors.Join(errs...)
+}
+
+//go:generate mockgen -source=mic_reconciler.go -package=controllers -destination=mock_mic_reconciler.go pullPodManager
+type pullPodManager interface {
+	listImagesPullPods(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig) ([]v1.Pod, error)
+	deletePod(ctx context.Context, pod *v1.Pod) error
+	createPullPod(ctx context.Context, imageSpec *kmmv1beta1.ModuleImageSpec, micObj *kmmv1beta1.ModuleImagesConfig) error
+	getPullPodForImage(pods []v1.Pod, image string) *v1.Pod
+}
```
ybettan, line 246: How about we create a public interface for pods instead of another reconciler helper API?
* listImagesPullPods only needs mic for name and namespace
* createPullPod can get parameters like name,ns, image (string) instead of MIC/MICSpec

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	imageLabelKey       = "kmm.node.kubernetes.io/module-image"
+)
+
+// micReconciler reconciles a MIC (moduleimagesconfig) object
+type micReconciler struct {
+	micReconHelper micReconcilerHelper
+	podHelper      pullPodManager
+}
+
+func NewMICReconciler(client client.Client, micHelper mic.MIC, mbscHelper mbsc.MBSC, scheme *runtime.Scheme) *micReconciler {
```
ybettan, line 50:
```suggestion
func NewMICReconciler(client client.Client, micAPI mic.MIC, mbscAPI mbsc.MBSC, scheme *runtime.Scheme) *micReconciler {
```

WDYT?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	return nil
+}
+
+func (ppmi *pullPodManagerImpl) createPullPod(ctx context.Context, imageSpec *kmmv1beta1.ModuleImageSpec, micObj *kmmv1beta1.ModuleImagesConfig) error {
+	restartPolicy := v1.RestartPolicyOnFailure
+	if imageSpec.Build != nil || imageSpec.Sign != nil {
+		restartPolicy = v1.RestartPolicyNever
+	}
+	imagePullSecrets := []v1.LocalObjectReference{}
```
ybettan, line 294:
```suggestion
	var imagePullSecrets []v1.LocalObjectReference
```
There is no point of initializing it if we override it [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/1002/files#diff-9101b15d22c840a3bae15419ccf8c1aaaae06da4d9d15f59e798045d152e2866R296).

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+			Namespace:    micObj.Namespace,
+			Labels: map[string]string{
+				moduleImageLabelKey: micObj.Name,
+				imageLabelKey:       imageSpec.Image,
+			},
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
+				{
+					Name:    workerContainerName,
```
ybettan, line 311: Why do we call this container `worker`? Isn't it confusing with the actual worker pods?

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed the comment so that it will be more explanatory

yevgeny-shnaidman:
> > The MIC deletion flow is not handled in the PR, will be handled later
> 
> Why does it has to be handled here. It is only supposed to be deleted by the `Module` reconciler once a `Module` is being deleted isn't it?

We will probably need to add finalizer and handle pull pods deletion and MBSC deletion.

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: Pod by itself won't provide us with any additional info. We have it label (image), and that is the only interesting thing about it. We won't get any additional info from it, that will help us to understand how pod without RestartOnFailure was created when build or sign are not present

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: in case updating the MIC status on line 165 fails ( for any reason), we need those pods to be present in the cluster , in order to be able to execute the next reconciliation loop correctly

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+	errs := make([]error, 0, len(podsToDelete))
+	for _, pod := range podsToDelete {
+		err = mrhi.podHelper.deletePod(ctx, &pod)
+		errs = append(errs, err)
+	}
+
+	return errors.Join(errs...)
+}
+
+func (mrhi *micReconcilerHelperImpl) updateStatusByMBSC(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig) error {
```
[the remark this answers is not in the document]
yevgeny-shnaidman: the MBSC objet is extracted inside the function, based on the MIC object

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: we did. How does this code contradict this statement?

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: This API is targeted specifically for MIC.  Creating it as a public interface which is used only by one controller does not make much of a difference.We have another helper pod interface , in NMC controller. We might consider unifying them , but in any case it should be in a different PR.

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
ybettan: > Pod by itself won't provide us with any additional info

I think it is very helpful just to see the failing pod (as you expect a pull pod to be created). You can also see why the pod failed (because the node isn't healthy or because it couldn't pull the image). It can also help us see that the pod failed to pull the image because of access denied and not just because the image doesn't exist.

IMO, it is helpful to see failing objects and not immediately remove them. We got the same feedback from multiple directions when we were directly deleting build pod. Although build pod logs are more helpful, it is still benefitial to have a failing pull pods.

internal/controllers/mic_reconciler.go
ybettan: Makes sense. Can we put your explanation in a comment? I will probably ask the same question each time I see this code :)

internal/controllers/mic_reconciler.go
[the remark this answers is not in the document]
ybettan: Why don't we do this outside this method and pass an MBSC to it? I think it will be a bit more clear.

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (mrhi *micReconcilerHelperImpl) processImagesSpecs(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig, pullPods []v1.Pod) error {
+	errs := make([]error, len(micObj.Spec.Images))
+	for _, imageSpec := range micObj.Spec.Images {
+		imageState := mrhi.micHelper.GetImageState(micObj, imageSpec.Image)
+
+		// image State does not exists either new image or pull pod is still running
+		switch imageState {
+		case "":
```
ybettan, line 212:
```suggestion
		switch imageState {
		
		// image State does not exists either new image or pull pod is still running
		case "":
```

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+			// image State is not set: either new image or pull pod is still running
+			if mrhi.podHelper.getPullPodForImage(pullPods, imageSpec.Image) == nil {
+				// no pull pod- create it, otherwise we wait for it to finish
+				err := mrhi.podHelper.createPullPod(ctx, &imageSpec, micObj)
+				errs = append(errs, err)
+			}
```
ybettan, line 218: I don't know how I feel about the next suggestion, but WDYT about always applying the pull pod (createOrPatch)? If the pod already exists, then k8s will do nothing, otherwise, it will create it.

internal/controllers/mic_reconciler.go
ybettan: I remember we said that if a pull pod has failed but there is a build/sign section, we set it directly to `ImageNeedsBuilding`. In the current code, it is first set to `ImageDoesNotExist` and only in the next reconciler loop it is being updated to `ImageNeedsBuilding`.

I think we should check if a build/sign section exists in the case where the status is `""` an not here.

I see few issues with the current approach:
1. We temporarily set the status to a "wrong" status (that will be overwritten soon) - an operator using KMM, may catch the wrong status and don't check it again later.
2. We can't guarantee that after updating the status to `ImageDoesNotExist` a new reconciliation is directly being triggered.

internal/controllers/mic_reconciler.go
ybettan: Yes, I do think we should unify them and create a public interface for them. I will open a task for it.

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: but here we are not talking about failing pod, we are talking about invalid pod that should not have been created in the first place. This is not an issue for the customer, it is an issue for us , and the solution won't be provided by looking at pod, but by looking at operator log. In case the we do have build or sign section, then we go to build flow. If we start keeping failing pods (which we expect to happen anyway, and it is a normal flow, it will be very confusing to look at the namepsace

internal/controllers/mic_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: MBSC is used only inside this function.  Eventually every function we have is using MIC object, since it contains the input needed by every function.

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: mmm... this code does not deal with setting the status of the images in the MIC. This code deals with either creating a pull pod or triggering MBSC based on the status of the image in the MIC

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,336 @@
...
+
+	return mrhi.client.Status().Patch(ctx, micObj, patchFrom)
+}
+
+func (mrhi *micReconcilerHelperImpl) processImagesSpecs(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig, pullPods []v1.Pod) error {
+	errs := make([]error, len(micObj.Spec.Images))
+	for _, imageSpec := range micObj.Spec.Images {
+		imageState := mrhi.micHelper.GetImageState(micObj, imageSpec.Image)
+
+		// image State does not exists either new image or pull pod is still running
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: createOrPatch does not make sense in case of Pods: pods can not be updated, they can only be created or deleted

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,332 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	MICReconcilerName = "MICReconciler"
```
ybettan, line 37: Do we have to export this const?

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,332 @@
...
+		// [TODO] delete check image and build/sign pods
+		return res, nil
+	}
+
+	pods, err := r.podHelper.listImagesPullPods(ctx, micObj)
+	if err != nil {
+		return res, fmt.Errorf("failed to get the image pods for mic %s: %v", micObj.Name, err)
+	}
+
+	// update the status of the current images in the Spec based on pull pods
```
ybettan, line 84: I think this comment is redundant. The method's name is self-explanatory.

internal/controllers/mic_reconciler.go
```diff
@@ -0,0 +1,332 @@
...
+	err = r.micReconHelper.updateStatusByPullPods(ctx, micObj, pods)
+	if err != nil {
+		return res, fmt.Errorf("failed tp update the status for MIC %s based on pull pods: %v", micObj.Name, err)
+	}
+
+	// update the status of the current images in the Spec based on MBSC status
```
ybettan, line 90: I think this comment is redundant. The method's name is self-explanatory.

internal/controllers/mic_reconciler.go
ybettan: The only reason the status should be `ImagedoesNotExist` is for images that doesn't exist and can't be built, therefore, I don't think this is the correct place to check if a build/sign section exists. At this point of the code we should assume that there is not a build/sign section (otherwise, we would have never set the status to `ImageDoesNotExist` to begin with.

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: Yes, it will be used in the main

internal/controllers/mic_reconciler.go
ybettan: > we are talking about invalid pod that should not have been created in the first place.

If we have a special status reserved for such cases, then those are valid pods IMO.

> the solution won't be provided by looking at pod, but by looking at operator log.

Looking at the failing pod is much simpler. When a build pod fails, the first thing that we check is the build pod events/logs.

> In case the we do have build or sign section, then we go to build flow.

I think only those pods should be deleted because those are valid failures and we will see:
* the image status becoming `NeedsBuilding`
* an MBSC being created

> If we start keeping failing pods (which we expect to happen anyway, and it is a normal flow, it will be very confusing to look at the namepsace

I don't think it will if the only failing pods we keep are the one that fits the `ImageDoesNotExist` scenario.

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: ImageDoesNotExists also applies to images whose Build has failed.  In case ImageDoesNotExist but has build or sign, we want to patch he MBSC, in case it is part of fixing the build

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mic_reconciler.go
ybettan: Shouldn't the status be [build-failed](https://github.com/kubernetes-sigs/kernel-module-management/blob/6ecab10fd511f12b6704cc0a417238e450350d2b/api/v1beta1/moduleimagesconfig_types.go#L34) in case the build has failed?!

PR #1003, opened 2025-02-09
Adding Patch API to mbsc package

yevgeny-shnaidman:


internal/mbsc/mbsc.go
```diff
@@ -48,3 +45,31 @@ func (m *mbsc) GetMBSC(ctx context.Context, name, namespace string) (*kmmv1beta1
...
+	mbscObj := &kmmv1beta1.ModuleBuildSignConfig{
+		ObjectMeta: metav1.ObjectMeta{Name: name, Namespace: namespace},
+	}
+
+	_, err := controllerutil.CreateOrPatch(ctx, m.client, mbscObj, func() error {
+		setModuleImageSpec(mbscObj, moduleImageSpec)
+		mbscObj.Spec.ImageRepoSecret = imageRepoSecret
+		return controllerutil.SetOwnerReference(owner, mbscObj, m.scheme)
+	})
+	return err
```
ybettan, line 64:
```suggestion
	if err != nil {
	    return fmt.Errorf("failed to create or patch...", err)
	}
	
	return nil
```

internal/mbsc/mbsc.go
```diff
@@ -5,38 +5,35 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 //go:generate mockgen -source=mbsc.go -package=mbsc -destination=mock_mbsc.go
 
 type MBSC interface {
-	SetModuleImageSpec(mbscObj *kmmv1beta1.ModuleBuildSignConfig, moduleImageSpec *kmmv1beta1.ModuleImageSpec)
 	GetMBSC(ctx context.Context, name, namespace string) (*kmmv1beta1.ModuleBuildSignConfig, error)
+	PatchMBSC(ctx context.Context, name string, namespace string, moduleImageSpec *kmmv1beta1.ModuleImageSpec,
```
ybettan, line 21: WDYT about renaming to `ApplyMBSC`? Sometime this will create a new object and not just patch. Also it will be more consistent with the MIC package.

yevgeny-shnaidman: it does not apply the whole object, as in case with MBSC

internal/mbsc/mbsc.go
yevgeny-shnaidman: no need to add that additional error info: the code is simple, we know precisely which function produces the error, and the error itself will be self explanatory. In addition, the code that will be calling this function will append its own error message, so we will have no problem to debug it

internal/mbsc/mbsc.go
yevgeny-shnaidman: I think that Patch is more explanatory, since it means that not the whole object is updated, but only a part of it

ybettan: But it is possible that the whole object is being created since you are calling `CreateOrPatch`. If we wan't to patch only then why don't you use `client.Patch()`?

yevgeny-shnaidman: The object will be created once, but after that every time there is only a Patch

ybettan: So the function is Creating or Patching :) not just patching. Someone looking at the function call may think no new object will be created when a Patch function is called.

yevgeny-shnaidman: How about CreateOrPatch

ybettan: I am OK with that, we if we choose `CreateOrPatch` can you please update it in the `MIC` package as well to be consistent?

yevgeny-shnaidman: ok, but in a different PR

PR #1005, opened 2025-02-11
Changing the names of the APIs of mic package

yevgeny-shnaidman:
1) changing ApplyMIC to CreateOrPatch
2) chaning NewModuleImagesConfigAPI to New

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -123,7 +123,7 @@ func main() {
 	buildHelperAPI := build.NewHelper()
 	nodeAPI := node.NewNode(client)
 	kernelAPI := module.NewKernelMapper(buildHelperAPI, sign.NewSignerHelper())
-	micAPI := mic.NewModuleImagesConfigAPI(client, scheme)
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -123,7 +123,7 @@ func main() {
 	buildHelperAPI := build.NewHelper()
 	nodeAPI := node.NewNode(client)
 	kernelAPI := module.NewKernelMapper(buildHelperAPI, sign.NewSignerHelper())
-	micAPI := mic.NewModuleImagesConfigAPI(client, scheme)
+	micAPI := mic.New(client, scheme)
```
ybettan, line 126: I am not sure how I feel about this one. I agree in theory, but in practice is is easier to grep `NewModuleImagesConfigAPI` and we use this convention to all out other interfaces...

yevgeny-shnaidman: maybe we need to start changing it. The controller-runtime is doing this also. I think it is easy to find, since we know what is the package

PR #1008, opened 2025-02-12
deleting orphan worker pods

TomerNewman:
Previously, when worker pods failed for any reason and the user deleted the associated Module CR,
the failing worker pods would remain in the cluster instead of being removed. 
This commit introduces a garbage collection mechanism to ensure that orphaned worker pods are properly deleted when
the corresponding Module CR is removed.

---


internal/controllers/nmc_reconciler.go
```diff
@@ -261,6 +274,39 @@ func newNMCReconcilerHelper(client client.Client, pm podManager, recorder record
...
+
+	var errs []error
+	for _, pod := range podsList {
+		podModuleName, err := getModuleNameFromWorkerPod(pod.Name)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("could not get module name from worker pod %s: %v", pod.Name, err))
+			continue
+		}
+		if !specModules.Has(podModuleName) {
+			if err = h.RemovePodFinalizers(ctx, nmcObj.Name); err != nil {
```
ybettan, line 296: Why do we have a finalizer on the worker pod?

internal/controllers/nmc_reconciler.go
```diff
@@ -261,6 +274,39 @@ func newNMCReconcilerHelper(client client.Client, pm podManager, recorder record
...
+				errs = append(errs, fmt.Errorf("could not remove worker pod %s finalizers: %v", pod.Name, err))
+			}
+			if err = h.pm.DeletePod(ctx, &pod); err != nil {
+				errs = append(errs, fmt.Errorf("could not delete pod %s: %v", pod.Name, err))
+			}
+		}
+	}
+	if len(errs) == 0 {
+		return nil
+	}
```
ybettan, line 306: I think that `errors.Join()` return `nil` if all the errors in the list are `nil`.

internal/controllers/nmc_reconciler.go
```diff
@@ -178,6 +178,10 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 		errs = append(errs, fmt.Errorf("failed to GC in-use labels for NMC %s: %v", req.NamespacedName, err))
 	}
 
+	if err := r.helper.GarbageCollectWorkerPods(ctx, &nmcObj); err != nil {
```
ybettan, line 181: I am wondering if it won't be better to add a finalizer to the NMC object to make sure it is not being deleted before the worker pods are and if the NMC is being delete we go over all its worker pod and make sure we remove all failed pods instead of garbage collecting them a-syncronically.

 WDYT @yevgeny-shnaidman @TomerNewman

internal/controllers/nmc_reconciler.go
```diff
@@ -235,6 +247,7 @@ func GetContainerStatus(statuses []v1.ContainerStatus, name string) v1.Container
 
 type nmcReconcilerHelper interface {
 	GarbageCollectInUseLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
+	GarbageCollectWorkerPods(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
```
ybettan, line 250: I think this methods belongs in the https://github.com/kubernetes-sigs/kernel-module-management/blob/946be3a4a12a417cb0d2486b510023edbc80d13f/internal/controllers/nmc_reconciler.go#L743 interface instead of the `nmcReconcilerHelper` interface.

The pod manager will probably move to its own package as a followup to https://github.com/kubernetes-sigs/kernel-module-management/pull/1010 so I think it will better to move it there so it moves with the rest of the interface to the `pod` package once merged.

ybettan:
This bug will be solved without this PR once we start using the new MIC because no NMC entry will be created for images that doesn't exist yet.

It is still good to keep this fix in case the worker pod fails due to other reasons such a kmod that was compiled for the wrong kernel and similar issues.

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: I would keep the function in the nmcReconcilerHelper interface.This is the logic of controller, and not of the pod. It will use the pod interface functions , like ListPods and DeletePods, but the decision which pods to delete should be in the controller

internal/controllers/nmc_reconciler.go
```diff
@@ -261,6 +274,39 @@ func newNMCReconcilerHelper(client client.Client, pm podManager, recorder record
 	}
 }
 
+func (h *nmcReconcilerHelperImpl) GarbageCollectWorkerPods(ctx context.Context, nmcObj *kmmv1beta1.NodeModulesConfig) error {
+	podsList, err := h.pm.ListWorkerPodsOnNode(ctx, nmcObj.Name)
+	if err != nil {
+		return fmt.Errorf("could not list worker Pods: %v", err)
+	}
+
+	specModules := sets.New[string]()
```
yevgeny-shnaidman, line 283: I think you need to check also for module in the status. Possible scenario: kernel module is loaded successfully, then Module is deleted. In that case Module is present in status, but not in spec, and you will end up deleting a valid unload worker pod

internal/controllers/nmc_reconciler.go
TomerNewman: correct, fixing

internal/controllers/nmc_reconciler.go
TomerNewman: correct, fixing

internal/controllers/nmc_reconciler.go
ybettan: I don't think there is a point of having a "pure pod package" as it is not different than what is already offered in https://pkg.go.dev/k8s.io/kubernetes/pkg/api/v1/pod.

This function is no different in its skope than any other method in https://github.com/kubernetes-sigs/kernel-module-management/blob/946be3a4a12a417cb0d2486b510023edbc80d13f/internal/controllers/nmc_reconciler.go#L743-L750 so I think it should be there.

yevgeny-shnaidman: but those functions are not making decisions for the controller: they use the data the controller passes them to do something with the actual pods: create, create templates, delete. GarbageCollectWorkerPods needs to decide which pods to delete based on the NMC, Module etc', and this decision must belong to the controller

ybettan: You convinced me.

internal/controllers/nmc_reconciler.go
ybettan: This is not relevant to this PR.

internal/controllers/nmc_reconciler.go
```diff
@@ -261,6 +266,36 @@ func newNMCReconcilerHelper(client client.Client, pm podManager, recorder record
...
+	statusModules := sets.New[string]()
```
yevgeny-shnaidman, line 276: no need to define a new variable. We just want to know which module is present in either spec or status, so we need only one Set, and even if module is present in Spec and Status, it will still hold only one entry in Spec

TomerNewman: I know that we can do it in 1 set variable, it would not be more organized to separate the cases though?

internal/controllers/nmc_reconciler.go
TomerNewman: @ybettan why not? in order to delete the worker pod we have to delete his finelizers

TomerNewman: oh wait I need to remove only for the relevant pod, fixing now

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: IMHO no, we just need to check that the module is present in either spec or module, if you have a variable that is called "modulePresentInNMC", it will be self-explanatory

PR #1010, opened 2025-02-13
Creating the `pod` package.

ybettan:
The MIC controller is defining an helper interface for manipulating pods. The NMC controller is doing the same. Moving the MIC's pod interface to its own package for simplicity and code re-use.

The NMC's pod interface will move to this package as well in an upcoming commit.

---

/assign @yevgeny-shnaidman


yevgeny-shnaidman:
lets split this PR in 2:
1) new pod package and it unit-tests.
2) updating the controllers , once the new pod package is pushed

internal/pod/pod.go
```diff
@@ -0,0 +1,141 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:generate mockgen -source=pod.go -package=pod -destination=mock_pod.go
+
+type ImagePuller interface {
```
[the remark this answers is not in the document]
ybettan: Second interface will come in a followup PR to make this PR smaller and easier to review.

ybettan:
> lets split this PR in 2:
> 
> new pod package and it unit-tests.
> updating the controllers , once the new pod package is pushed

This will leave the codebase with 100% code duplication between step 1 and step 2. This is 99% the same code just moving from one place to another. I don't think a smaller PR will justify duplicating the code instead of moving it from one package to another.

We still have a followup PR to move the NMC `podManager` interface in the same manner as a followup PR to this one.

internal/pod/pod.go
```diff
@@ -0,0 +1,141 @@
...
+	CreatePullPod(ctx context.Context, imageSpec *kmmv1beta1.ModuleImageSpec, micObj *kmmv1beta1.ModuleImagesConfig) error
+	DeletePod(ctx context.Context, pod *v1.Pod) error
+	ListPullPods(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig) ([]v1.Pod, error)
+	GetPullPodForImage(pods []v1.Pod, image string) *v1.Pod
+}
+
+const (
+	moduleImageLabelKey = "kmm.node.kubernetes.io/module-image-config"
+	ImageLabelKey       = "kmm.node.kubernetes.io/module-image"
```
yevgeny-shnaidman, line 26: WDYT if we make the ImageLabelKey also private, and add an interface function that returns the image of the pullPod?

internal/pod/pod.go
```diff
@@ -0,0 +1,141 @@
...
+
+	for i, pod := range pods {
+		if image == pod.Labels[ImageLabelKey] {
+			return &pods[i]
+		}
+	}
+	return nil
+}
+
+func (ipi *imagePullerImpl) pullPodTemplate(ctx context.Context, imageSpec *kmmv1beta1.ModuleImageSpec,
```
yevgeny-shnaidman, line 84: why do we need this function? it does not make it easier to write unit-test, and only one function is calling it. Why not make its code to be part of CreatePullPod?

ybettan: I though it is a bit simpler but it is indeed not helping unit tests and indeed no-one else is using it.
I will revert it to be how it was.

internal/pod/pod.go
ybettan: Done.

internal/pod/pod.go
ybettan: Done.

PR #1018, opened 2025-02-17
Changing MSBC image status structure

yevgeny-shnaidman:
Currently MBSC image's spec contains the action that MBSC controller
    should execute(sign or build). But since the action does not exists in
    the image's status, the MIC controller cannot know if the success status
    refers to the previous action (i.e build) or the current requested
    action( i.e sign). In addition, since action will appear in status,
    there is no need to have explicit statuses like BuildFailed: we change
    it to ActionFailed or ActionSuccess.
    This PR contains the following:
    1) adding action field to the MBSC image' status
    2) chaning the type of Status field: now it will be only ActionFailure
       or ActionSuccess
    2) update the MIC controller MBSC lookup function to decide what is the
       current status of image in MIC, based on MBSC image's action and
    status fields
    3) unit-tests


ybettan:
> Currently MBAX

Currently MBSC...

yevgeny-shnaidman:
> > Currently MBAX
> 
> Currently MBSC...

Fixed

internal/controllers/mic_reconciler.go
```diff
@@ -193,23 +193,27 @@ func (mrhi *micReconcilerHelperImpl) updateStatusByMBSC(ctx context.Context, mic
...
-		imageSpec := mrhi.micHelper.GetModuleImageSpec(micObj, imageState.Image)
-		if imageSpec == nil {
-			// image not found in spec, ignore
+	for _, mbscImageState := range mbsc.Status.Images {
+		micImageSpec := mrhi.micHelper.GetModuleImageSpec(micObj, mbscImageState.Image)
+		mbscImageSpec := mrhi.mbscHelper.GetImageSpec(mbsc, mbscImageState.Image)
+		if micImageSpec == nil || mbscImageSpec == nil || mbscImageSpec.Action != mbscImageState.Action {
+			// spec for image does not exists in either MIC or MBCS, which is not normal, or
+			// the action in MBSC status not equal to action in MBSC spec, which means that the status
+			// was not yet updated. In any case - ignore
```
ybettan, line 202: Isn't it possible that the image was removed in the module and it was removed from MIC spec and MBSC spec but still present in MIC status and MBSC status because the controllers didn't update it yet?

yevgeny-shnaidman: it is , but it still means that we can ignore it

internal/controllers/mic_reconciler.go
```diff
@@ -188,23 +188,29 @@ func (mrhi *micReconcilerHelperImpl) updateStatusByMBSC(ctx context.Context, mic
...
+			mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageDoesNotExist)
+		case mbscStatus == kmmv1beta1.ActionSuccess && mbscAction == kmmv1beta1.SignImage:
+			// sign action succeeded - image exists, nothing more to do
+			mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageExists)
+		case mbscStatus == kmmv1beta1.ActionSuccess && micImageSpec.Sign != nil:
+			// build succeeded and sign exists - image needs to be signed
+			mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageNeedsSigning)
+		case mbscStatus == kmmv1beta1.ActionSuccess:
+			// build succeeded, no sign - image exists
+			mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageExists)
```
ybettan, line 213:
```suggestion
switch {

case mbscStatus == kmmv1beta1.ActionFailure:
    // any failure (build or sign) - image does not exists
    mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageDoesNotExist)

case mbscStatus == kmmv1beta1.ActionSuccess:

    if (mbscAction == kmmv1beta1.BuildImage && micImageSpec.Sign == nil) || mbscAction == kmmv1beta1.SignImage {
        // image exists, nothing more to do
        mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageExists)
    } else if mbscAction == kmmv1beta1.BuildImage && micImageSpec.Sign != nil {
        // build succeeded and sign exists - image needs to be signed                                           
        mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageNeedsSigning)
    }
```

I think it is easier to understand and to make sure we didn't forget any optional combination.

ybettan: Since github diff make it hard to read then this is the case:
```
switch {

case mbscStatus == kmmv1beta1.ActionFailure:
    // any failure (build or sign) - image does not exists
    mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageDoesNotExist)

case mbscStatus == kmmv1beta1.ActionSuccess:

    if (mbscAction == kmmv1beta1.BuildImage && micImageSpec.Sign == nil) || mbscAction == kmmv1beta1.SignImage {
        // image exists, nothing more to do
        mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageExists)
    } else if mbscAction == kmmv1beta1.BuildImage && micImageSpec.Sign != nil {
        // build succeeded and sign exists - image needs to be signed                                           
        mrhi.micHelper.SetImageStatus(micObj, mbscImageState.Image, kmmv1beta1.ImageNeedsSigning)
    }
}
```

yevgeny-shnaidman: I wrote it especially this way in order to avoid using "switch" and "if", which in my opinion makes code a little bit harder to read. With only switch case we can see exactly what is going on

PR #1020, opened 2025-02-18
Moving the `podManager` NMC helper interface to the new `pod` package.

ybettan:
Now that we have the pod package, we can move pod-related interfaces to this package.

---

/assign @yevgeny-shnaidman


internal/pod/kmodloader.go
```diff
@@ -0,0 +1,551 @@
...
+	"k8s.io/kubectl/pkg/cmd/util/podcmd"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/yaml"
+)
+
+//go:generate mockgen -source=kmodloader.go -package=pod -destination=mock_kmodloader.go KmodLoader
+
+type KmodLoader interface {
```
yevgeny-shnaidman, line 31: "KmodLoader" is not a good name for this interface. I was looking at the main.go and got really confused when i encountered that interface's initialization function. I think something like "WorkerPodManager/Handler" would be more appropriate and clear.

internal/pod/kmodloader.go
```diff
@@ -0,0 +1,551 @@
...
+	CreateUnloaderPod(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleStatus) error
+	DeletePod(ctx context.Context, pod *v1.Pod) error
+	GetWorkerPod(ctx context.Context, podName, namespace string) (*v1.Pod, error)
+	ListWorkerPodsOnNode(ctx context.Context, nodeName string) ([]v1.Pod, error)
+	LoaderPodTemplate(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleSpec) (*v1.Pod, error)
+	UnloaderPodTemplate(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleStatus) (*v1.Pod, error)
+}
+
+const (
+	WorkerActionLoad           = "Load"
```
yevgeny-shnaidman, line 42: can we hide those exported variables, making them private and defining Getter/Setter interface functions if needed?

internal/controllers/nmc_reconciler.go
```diff
@@ -78,13 +58,14 @@ func NewNMCReconciler(
 	workerCfg *config.Worker,
 	recorder record.EventRecorder,
 	nodeAPI node.Node,
+	kmodLoader pod.KmodLoader,
```
yevgeny-shnaidman, line 61: lets keep podManager variable name, much more clearer

ybettan: I will use `WorkerPodManager` - as you suggested. `podManager` is too generic and we have another `podManager` interface coming in this package in a following PR - I try to give them meaningful names.

internal/pod/kmodloader.go
ybettan: Do you mean all the exported variables in this file?
```
        WorkerActionLoad = "Load"
	WorkerActionUnload         = "Unload"
	NodeModulesConfigFinalizer = "kmm.node.kubernetes.io/nodemodulesconfig-reconciler"
	ActionLabelKey             = "kmm.node.kubernetes.io/worker-action"
	ConfigAnnotationKey        = "kmm.node.kubernetes.io/worker-config"
	WorkerContainerName        = "worker"
	HashAnnotationKey          = "kmm.node.kubernetes.io/worker-hash"
```

ybettan: I will investigate possible options in a future PR. Let's keep this one a simple as possible (just moving code from one place to another).

internal/controllers/nmc_reconciler.go
ybettan: Fixed.

PR #1021, opened 2025-02-19
Adding KernelVersion to the MIC image spec

yevgeny-shnaidman:
MIC image spec is being propgated as is to the MBSC image spec. As part of building sign and build pod, one of the inputs is the kernel version that this image is targeted for. This PR does the following:
1. Add KernelVersion field to the ModuleImageSpec
2. module reconciler sets the KernelVersion in the imageSpec variable
3. changing unit-tests

### DIFF
--- a/api/v1beta1/moduleimagesconfig_types.go
+++ b/api/v1beta1/moduleimagesconfig_types.go
@@ -39,6 +39,9 @@ type ModuleImageSpec struct {
 	// image
 	Image string `json:"image"`
 
+	// kernel version for which this image is targeted
+	KernelVersion string `json:"kernelVersion"`
+
 	// Build contains build instructions, in case image needs building
 	// +optional
 	Build *Build `json:"build,omitempty"`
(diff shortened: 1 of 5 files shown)


api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -39,6 +39,9 @@ type ModuleImageSpec struct {
 	// image
 	Image string `json:"image"`
 
+	// kernel version for which this image is targeted
```
ybettan, line 42:
```suggestion
	// kernel version for which this image is targeting
```

ybettan:
> As part of building sign and build pod

Can we fix this sentence in the commit message?

api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -39,6 +39,9 @@ type ModuleImageSpec struct {
...
+	KernelVersion string `json:"kernelVersion"`
```
ybettan, line 43: If it is needed by the MBSC then lets add it to the MBSC object as well in the same PR as this is the same logical change (and the same commit message).

PR #1022, opened 2025-02-19
Adding Selector field to the MIC image spec

yevgeny-shnaidman:
MIC image spec is being propagated as is to the MBSC image spec. As part of creating sign/build pod, one of the inputs is the selector field, which defines where the pod should be scheduled. This PR does the following:
1. Add Selector field to the ModuleImageSpec
2. module reconciler sets the Selector in the imageSpec variable
3. changing unit-tests


api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -42,6 +42,9 @@ type ModuleImageSpec struct {
 	// kernel version for which this image is targeted
 	KernelVersion string `json:"kernelVersion"`
 
+	// selector for pod that should run builds
+	Selector map[string]string `json:"selector,omitempty"`
```
ybettan, line 46: We already have a selector in the build object. Why do we need another one?

yevgeny-shnaidman: the selector should be taken from mld: its value is decided based on Selector in Spec and Selector in Build. If Selector in Build is present, then it is taken, otherwise the Spec's Selector is used

ybettan: Then I think its the `Module` controller's job to update the MIC.build.selector (to be passed to MBSC) with the correct value (from module's spec/selector).

I don't think we should add a new selector in MIC and MBSC just to take that decision in MBSC controller. Let's just create the correct build object in MIC to begging with.

WDYT?

PR #1025, opened 2025-02-20
Un-exporting some of `WorkerPodManager` interface env variables.

ybettan:
When this interface was transitioned to the `pod` package, it was copied as is (as much as possible) and some variables had to be exported to be used as before.

Once the transition done, it was simpler to review some of those new exported env variables and convert them to usefull methods inside the `WorkerPodManager` interface.

---

/assign @yevgeny-shnaidman


internal/pod/workerpodmanager.go
```diff
@@ -36,16 +36,15 @@ type WorkerPodManager interface {
 	ListWorkerPodsOnNode(ctx context.Context, nodeName string) ([]v1.Pod, error)
 	LoaderPodTemplate(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleSpec) (*v1.Pod, error)
 	UnloaderPodTemplate(ctx context.Context, nmc client.Object, nms *kmmv1beta1.NodeModuleStatus) (*v1.Pod, error)
+	IsLoaderPod(p v1.Pod) bool
```
yevgeny-shnaidman, line 39: I think all those function can receive a Pod pointer, instead of a Pod structure. Passing it as a struct mean copying it, and it is a large structure. There are a lot of functions , even in kuberenetes packages, that receive parameters as pointers, event though they are not changing them

ybettan: Fixed.

PR #1029, opened 2025-02-24
Prevent "Exists" tolerations from having a value

TomerNewman:
This commit introduces a validation check in the webhook, to ensure that tolerations with operator: "Exists" do not specify a value, as this is invalid in k8s.

---

/cc @ybettan  @yevgeny-shnaidman

### DIFF
--- a/internal/webhook/module.go
+++ b/internal/webhook/module.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"regexp"
 	"strings"
 
@@ -110,7 +111,7 @@ func validateModule(mod *kmmv1beta1.Module) (admission.Warnings, error) {
 		return nil, fmt.Errorf("failed to validate kernel mappings: %v", err)
 	}
 
-	if err := validateModuleTolerations(mod); err != nil {
+	if err := validateTolerations(mod.Spec.Tolerations); err != nil {
 		return nil, fmt.Errorf("failed to validate Module's tolerations: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


yevgeny-shnaidman:
@TomerNewman please take a look at this function: [ValidateTolerations](https://github.com/kubernetes/kubernetes/blob/3985b78f062af889f10a0982226913eda9149c48/pkg/apis/core/validation/validation.go#L3969)
Maybe we can use it, so that we won;t need to update the webhook each time @cdvultur finds another bug

TomerNewman:
> @TomerNewman please take a look at this function: [ValidateTolerations](https://github.com/kubernetes/kubernetes/blob/3985b78f062af889f10a0982226913eda9149c48/pkg/apis/core/validation/validation.go#L3969) Maybe we can use it, so that we won;t need to update the webhook each time @cdvultur finds another bug

Good idea,I see it does not covers everything we did but it sure covers a lot

TomerNewman:
> > > @TomerNewman please take a look at this function: [ValidateTolerations](https://github.com/kubernetes/kubernetes/blob/3985b78f062af889f10a0982226913eda9149c48/pkg/apis/core/validation/validation.go#L3969) Maybe we can use it, so that we won;t need to update the webhook each time @cdvultur finds another bug
> > 
> > 
> > Good idea, I see it does not cover everything we did but it sure covers a lot
> 
> What is not covered by that function?

for example
```
if toleration.Effect != corev1.TaintEffectNoSchedule &&
			toleration.Effect != corev1.TaintEffectNoExecute &&
			toleration.Effect != corev1.TaintEffectPreferNoSchedule {
			return fmt.Errorf("toleration effect can be only {NoSchedule, NoExecute, PreferNoSchedule} but got %s", toleration.Effect)
		}
```

yevgeny-shnaidman:
> > > > @TomerNewman please take a look at this function: [ValidateTolerations](https://github.com/kubernetes/kubernetes/blob/3985b78f062af889f10a0982226913eda9149c48/pkg/apis/core/validation/validation.go#L3969) Maybe we can use it, so that we won;t need to update the webhook each time @cdvultur finds another bug
> > > 
> > > 
> > > Good idea, I see it does not cover everything we did but it sure covers a lot
> > 
> > 
> > What is not covered by that function?
> 
> for example
> 
> ```
> if toleration.Effect != corev1.TaintEffectNoSchedule &&
> 			toleration.Effect != corev1.TaintEffectNoExecute &&
> 			toleration.Effect != corev1.TaintEffectPreferNoSchedule {
> 			return fmt.Errorf("toleration effect can be only {NoSchedule, NoExecute, PreferNoSchedule} but got %s", toleration.Effect)
> 		}
> ```

I think it is covered here, if i am not mistaken: [EffectCoverage](https://github.com/kubernetes/kubernetes/blob/3985b78f062af889f10a0982226913eda9149c48/pkg/apis/core/validation/validation.go#L4007)

TomerNewman:
#### Note
ValidateTolerations is not importable so I implemented our own version based on that + added more unit tests

yevgeny-shnaidman:
@TomerNewman WDYT about not aggregating the errors, but exiting the function immediately  whenever you encounter the first error? it will simplify the code and make it more unique :)

TomerNewman:
> @TomerNewman WDYT about not aggregating the errors, but exiting the function immediately  whenever you encounter the first error? it will simplify the code and make it more unique :)

But lets say the user made multiple mistakes in multiple tolerations. 
Wouldn't it be better if he would get all of the errors from the webhook? Insread of only one?

Issue #1006, opened 2025-02-11
worker.setFirmwareClassPath needs to be updated to worker.firmwareHostPath in kmm.sigs.k8s.io/documentation/configure/


quic-vmellemp:
[https://github.com/kubernetes-sigs/kernel-module-management/commit/4728244e37f463f301d7864b6868e4369719d2bb](https://github.com/kubernetes-sigs/kernel-module-management/commit/4728244e37f463f301d7864b6868e4369719d2bb)

After the release of v2.2.0(above mentioned change to be specific), worker.setFirmwareClassPath is no longer valid. Could you please update the documentation in https://kmm.sigs.k8s.io/documentation/configure/

ybettan:
Thank you for pointing that our @quic-vmellemp. I have created an internal task to update the docs. Hopefully, we will change it ASAP.

PR #1031, opened 2025-02-26
Fixing the firmware configuration documentation.

ybettan:
The config fields to be used in order to modify the firmware path on the host but the user was changed rom `setFirmwareClassPath` to `firmwareHostPath` but the documentation wasn't update accordingly. This commit is about to fix that.

---

Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/1006 
/assign @yevgeny-shnaidman


yevgeny-shnaidman:
lgtm

PR #1032, opened 2025-02-26
Adding manager and maker/signer to buildsign package

yevgeny-shnaidman:
With the introduction of the MBSC controller, the flow is changing and
    there is no need for 2 different managers. There is no need to decide if
    the build/sign should be executed, since that decision is taken by the MIC controller.
    The new package must provide only the following capabilities:
    1. Get the running pods (build or sign), in order for MBSC controller to
       update its status
    2. Sync the current build/sign pods based on action parameter
    3. Delete build/sign successfull pods based on the action parameter
    
    The following files are added:
    1. manager - it will decide, based on the action field, which action
       should be taken: check pod, create pod etc'. It will use a dedicated
    maker/signer interfaces when creating/deleting relevant pod. In addition
    it will return the status of the running build/sign pods
    2. maker - will be responsible for creating the Build pod. Copied almost
       as is from the current build package
    3. signer - will be responsible for creating the Sign pod. Copied almost
       as is from the current sign package


internal/buildsign/helper.go
```diff
@@ -0,0 +1,67 @@
...
+
+		if o, ok := overridesMap[argName]; ok {
+			args[i] = o
+			unusedOverrides.Delete(argName)
+		}
+	}
+
+	for _, overrideName := range unusedOverrides.List() {
+		args = append(args, overridesMap[overrideName])
+	}
```
ybettan, line 42:
```suggestion
		args = append(args, unusedOverrides.List()...)
```
I may have a syntax error here but you get the point :)

internal/buildsign/helper.go
```diff
@@ -0,0 +1,67 @@
...
+
+	return args
+}
+
+func (m *helper) GetRelevantBuild(moduleBuild *kmmv1beta1.Build, mappingBuild *kmmv1beta1.Build) *kmmv1beta1.Build {
+	if moduleBuild == nil {
+		return mappingBuild.DeepCopy()
```
ybettan, line 49: What happens if `mappingBuild` is also `nil`? You will have a nil memory access.

internal/buildsign/helper.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	buildConfig := moduleBuild.DeepCopy()
+	if mappingBuild.DockerfileConfigMap != nil {
+		buildConfig.DockerfileConfigMap = mappingBuild.DockerfileConfigMap
+	}
+
+	buildConfig.BuildArgs = m.ApplyBuildArgOverrides(buildConfig.BuildArgs, mappingBuild.BuildArgs...)
+
+	// [TODO] once MGMT-10832 is consolidated, this code must be revisited. We will decide which
```
ybettan, line 63: This Jira is an assisted-installer Jira.

internal/buildsign/helper.go
yevgeny-shnaidman: that condition is checked in the caller code, when you reach this function, at least one of them must be present

internal/buildsign/helper.go
yevgeny-shnaidman: that won't work, we need to use the output of the list t they the value of the map, not just concatinate the output of the list

internal/buildsign/helper.go
yevgeny-shnaidman: removed

yevgeny-shnaidman:
> I think this PR should be splitted in a different way. There is no point of reviewing all those methods that are just copied from the current code in a new package.
> 
> WDYT about having a different split in such way that each PR is "atomic" on its own. For example, the `Helper` interface in `interfnal/build/helper.go` just move to the new `buildsign` package + anyone using those methods now uses them from the new package + remove old files.
> 
> IMO it is clearer to review a PR moving code from one place to another instead of 1 PR to add a bunch of code (that was already reviewed in the past) and then another PR to remove the same code from another package.
> 
> We can then have smaller PRs that each one of them is "complete" on its own.

I have move the Helper interface to a different [PR](https://github.com/kubernetes-sigs/kernel-module-management/pull/1032#issuecomment-2684927567), but all the rest will have to be in the same PR, since only MBSC controller is using those new interfaces

internal/buildsign/pod/maker.go
```diff
@@ -0,0 +1,329 @@
...
+	buildsignpod "github.com/kubernetes-sigs/kernel-module-management/internal/pod"
+)
+
+const (
+	dockerfileVolumeName = "dockerfile"
+)
+
+//go:generate mockgen -source=maker.go -package=pod -destination=mock_maker.go
+
+type Maker interface {
```
ybettan, line 32: How about we rename it to `Builder` (and the filename as well)?

internal/buildsign/pod/maker.go
```diff
@@ -0,0 +1,329 @@
...
+	MakePodTemplate(
```
ybettan, line 33: I am wondering why don't we move that method to the `pod` package in the `BuildSignPodManager` interface instead of a separate interface in a different package to also manager build/sign pods?

internal/buildsign/pod/signer.go
```diff
@@ -0,0 +1,266 @@
...
+var templateFS embed.FS
+
+var tmpl = template.Must(
+	template.ParseFS(templateFS, "templates/Dockerfile.gotmpl"),
+)
+
+//go:generate mockgen -source=signer.go -package=pod -destination=mock_signer.go
+
+type Signer interface {
+	MakePodTemplate(
```
ybettan, line 42: I am wondering why don't we move that method to the pod package in the BuildSignPodManager interface instead of a separate interface in a different package to also manager build/sign pods?

internal/buildsign/pod/maker.go
ybettan: We have agreed that `Maker` and `Signer` will be removed in a later PR and their method will be merge into the `BuildSignPodManager` interface instead. Some refactor is needed so it will be handled in a separate PR.

internal/buildsign/pod/maker.go
ybettan: We have agreed that `Maker` and `Signer` will be removed in a later PR and their method will be merge into the `BuildSignPodManager` interface instead. Some refactor is needed so it will be handled in a separate PR.

internal/buildsign/pod/signer.go
ybettan: We have agreed that `Maker` and `Signer` will be removed in a later PR and their method will be merge into the `BuildSignPodManager` interface instead. Some refactor is needed so it will be handled in a separate PR.

internal/buildsign/pod/manager.go
```diff
@@ -0,0 +1,133 @@
...
+		return kmmv1beta1.BuildOrSignStatus(""), fmt.Errorf("failed to get status from the pod %s/%s, action %s: %v",
+			foundPod.Namespace, foundPod.Name, action, err)
+	}
+	switch status {
+	case StatusCompleted:
+		return kmmv1beta1.ActionSuccess, nil
+	case StatusFailed:
+		return kmmv1beta1.ActionFailure, nil
+	}
+	return kmmv1beta1.BuildOrSignStatus(""), nil
```
ybettan, line 61: Since we should never reach this line, shouldn't we remove an error instead of `nil`?

internal/buildsign/pod/manager.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (pm *podManager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushImage bool, owner metav1.Object, action kmmv1beta1.BuildOrSignAction) error {
+	logger := log.FromContext(ctx)
+
+	var podType string
+	var podTemplate *v1.Pod
+	var err error
```
ybettan, line 69:
```suggestion
	var (
	    podType string
	    podTemplate *v1.Pod
	    err error
	)
```

internal/buildsign/pod/manager.go
```diff
@@ -0,0 +1,133 @@
...
+	if action == kmmv1beta1.BuildImage {
+		logger.Info("Building in-cluster")
+		podType = PodTypeBuild
+		podTemplate, err = pm.maker.MakePodTemplate(ctx, mld, owner, pushImage)
+	} else {
+		logger.Info("Signing in-cluster")
+		podType = PodTypeSign
+		podTemplate, err = pm.signer.MakePodTemplate(ctx, mld, owner, pushImage)
+	}
```
ybettan, line 78: I think that when we are `if`ing over enums, switch-case is cleaner. Don't you think so?

internal/buildsign/pod/manager.go
```diff
@@ -0,0 +1,133 @@
...
+		return fmt.Errorf("could not make Pod template: %v", err)
+	}
+
+	p, err := pm.buildSignPodManager.GetModulePodByKernel(ctx, mld.Name, mld.Namespace,
+		mld.KernelNormalizedVersion, podType, owner)
+
+	if err != nil {
+		if !errors.Is(err, ErrNoMatchingPod) {
+			return fmt.Errorf("error getting the %s pod: %v", podType, err)
+		}
```
ybettan, line 89: This is the second time we are running this code (first one was in the previous method). How about we move the check to see if the error is `ErrNoMatchingPod` inside the `GetModulePodByKernel`)?

internal/buildsign/pod/manager.go
yevgeny-shnaidman: we will reach this line in case the Pod is still running or preparing to be run (pending). The error handling is done on line 51

internal/buildsign/pod/manager.go
yevgeny-shnaidman: Fixed

internal/buildsign/pod/manager.go
yevgeny-shnaidman: Fixed

internal/buildsign/pod/manager.go
yevgeny-shnaidman: ErrNoMatchingPod is actually returned by the GetModulePodByKernel on purpose, so that the callee code will know what to do.

internal/buildsign/pod/manager.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get %s pods for mbsc %s/%s: %v", podType, namespace, name, err)
+	}
+
+	deleteNames := make([]string, 0, len(pods))
+	for _, pod := range pods {
+		if pod.Status.Phase == v1.PodSucceeded {
+			err = pm.buildSignPodManager.DeletePod(ctx, &pod)
+			if err != nil {
+				return nil, fmt.Errorf("failed to delete %s pod %s: %v", podType, pod.Name, err)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/buildsign/pod/manager.go
ybettan: How about we add a comment for clarity?

PR #1035, opened 2025-02-27
Moving BuildSignPodManager from the `pod` to the `buildsign` package

yevgeny-shnaidman:
BuildSignPodManager interface is used only by the build and sign packages.
 Since we are moving into buildsign package, it make sense to move that
 interface there, especially as it is a porting layer


ybettan:
I would rename the commit title to
```
Moving BuildSignPodManager from the `pod` to the `buildsign` package.
```

And also updating the commit message accordingly.

ybettan:
Do you plan to remove the `Maker` and `Signer` interfaces and add their methods to the `BuildSignPodManager` interface in a separate PR?

yevgeny-shnaidman:
> Do you plan to remove the `Maker` and `Signer` interfaces and add their methods to the `BuildSignPodManager` interface in a separate PR?

Yes

PR #1037, opened 2025-03-04
Changing the GarbageCollector API of the buildsign manager

yevgeny-shnaidman:
Currently the GarbageCollector API receives the PodType as an input parameter. The PodType  is defined in the pod package, which should not be used by the callee, since it is a porting layer. This commit changes the API to use kmmv1beta1.BuildOrSignAction, and the translation to the PodType will be done inside the manager code

### DIFF
--- a/internal/buildsign/pod/manager.go
+++ b/internal/buildsign/pod/manager.go
@@ -120,7 +120,11 @@ func (pm *podManager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushI
 	return nil
 }
 
-func (pm *podManager) GarbageCollect(ctx context.Context, name, namespace, podType string, owner metav1.Object) ([]string, error) {
+func (pm *podManager) GarbageCollect(ctx context.Context, name, namespace string, action kmmv1beta1.BuildOrSignAction, owner metav1.Object) ([]string, error) {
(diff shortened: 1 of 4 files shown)


internal/buildsign/pod/manager.go
```diff
@@ -120,7 +120,11 @@ func (pm *podManager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushI
 	return nil
 }
 
-func (pm *podManager) GarbageCollect(ctx context.Context, name, namespace, podType string, owner metav1.Object) ([]string, error) {
+func (pm *podManager) GarbageCollect(ctx context.Context, name, namespace string, action kmmv1beta1.BuildOrSignAction, owner metav1.Object) ([]string, error) {
+	podType := PodTypeBuild
+	if action == kmmv1beta1.SignImage {
+		podType = PodTypeSign
+	}
```
ybettan, line 127: Why do we need a translation, why can't we just add the action as the label instead of having another label for the pod type (which will always be the same as the action)?

yevgeny-shnaidman: it means changing a lot of code that right now we inherit as is. We can do it later, but for now i would prefer a simple translation

PR #1044, opened 2025-03-10
Updating MIC and MBSC definitions

yevgeny-shnaidman:
MIC and MBSC objects should be namespaced, since they are the direct derivitive of the Module object

### DIFF
--- a/api/v1beta1/moduleimagesconfig_types.go
+++ b/api/v1beta1/moduleimagesconfig_types.go
@@ -80,8 +80,9 @@ type ModuleImagesConfigStatus struct {
 	ImagesStates []ModuleImageState `json:"imagesStates"`
 }
 
-// +kubebuilder:object:root=true
-// +kubebuilder:subresource:status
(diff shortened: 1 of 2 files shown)


api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -80,8 +80,9 @@ type ModuleImagesConfigStatus struct {
 	ImagesStates []ModuleImageState `json:"imagesStates"`
 }
 
-// +kubebuilder:object:root=true
-// +kubebuilder:subresource:status
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:resource:scope=Namespaced
```
ybettan, line 85: It is already namespaced https://github.com/kubernetes-sigs/kernel-module-management/blob/ef3e3172497a9428014c561cc463ff153cef71da/api/v1beta1/moduleimagesconfig_types.go#L87

PR #1045, opened 2025-03-10
Adding GetImageStatus API to mbsc package

yevgeny-shnaidman:
This API return the status of the an image in the MBSC object based on the image and the action. Is required by the MBSC controller


internal/mbsc/mbsc.go
```diff
@@ -85,6 +86,15 @@ func (m *mbsc) SetImageStatus(mbscObj *kmmv1beta1.ModuleBuildSignConfig, image s
 	mbscObj.Status.Images = append(mbscObj.Status.Images, imageState)
 }
 
+func (m *mbsc) GetImageStatus(mbscObj *kmmv1beta1.ModuleBuildSignConfig, image string, action kmmv1beta1.BuildOrSignAction) kmmv1beta1.BuildOrSignStatus {
+	for _, imageState := range mbscObj.Status.Images {
+		if imageState.Image == image && imageState.Action == action {
```
ybettan, line 91: Why do we need to pass the action to get the status?

yevgeny-shnaidman: to make sure that we get the correct one. MIC is sending request for image and action, if we don't pass action, we might get the status for the previous action

yevgeny-shnaidman: the status is updated atomically by MBSC controller, but it is updated once the MBSC controller has the result. MIC should never update the status of MBSC object.

PR #1048, opened 2025-03-11
Updating image puller package with the new pod flow

yevgeny-shnaidman:
This commit contains the following changes:
1. Removing image label - image will be obtained from the Spec directly
2. Adding pod type label - one-time pod that should be considered as Failed once the container fails on image pull, and run-until-success, that is considered as InProgess, in case container ails on image pull
3. Adding GetPodStatus API - this api returns pod status based on the following:
   - the pod that fails to pull image stays in the Pending state, and the container status is updated with appropriate reason.So, if the pod is one-time pod, and container failed on image pull - status is Failed. Otherwise status is Running, since we let the pod restart the container with back-offs
4. Pod is created with restartPolicy Never, since if the pod managed to pull image - we consider it a success

### DIFF
--- a/internal/pod/imagepuller.go
+++ b/internal/pod/imagepuller.go
@@ -12,6 +12,26 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type PullPodStatus string
+
+const (
+	PullImageFailed        PullPodStatus = "pullFailed"
+	PullImageSuccess       PullPodStatus = "pullSuccess"
+	PullImageInProcess     PullPodStatus = "pullInProcess"
+	PullImageUnexpectedErr PullPodStatus = "unexpectedError"
+
+	imagePullBackOffReason = "ImagePullBackOff"
+	errImagePullReason     = "ErrImagePull"
+
+	moduleImageLabelKey = "kmm.node.kubernetes.io/module-image-config"
+	pullPodTypeLabelKey = "kmm.node.kubernetes.io/pull-pod-type"
+
+	pullerContainerName = "puller"
+
+	pullPodTypeOneTime  = "one-time-pull"
+	pullPodUntilSuccess = "until-success"
+)
+
 //go:generate mockgen -source=imagepuller.go -package=pod -destination=mock_imagepuller.go
 
 type ImagePuller interface {
@@ -20,13 +40,8 @@ type ImagePuller interface {
 	ListPullPods(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig) ([]v1.Pod, error)
 	GetPullPodForImage(pods []v1.Pod, image string) *v1.Pod
 	GetPullPodImage(pod v1.Pod) string
+	GetPullPodStatus(pod *v1.Pod) PullPodStatus
 }
-
-const (
-	moduleImageLabelKey = "kmm.node.kubernetes.io/module-image-config"
-	imageLabelKey       = "kmm.node.kubernetes.io/module-image"
-	pullerContainerName = "puller"
-)
 
 type imagePullerImpl struct {
 	client client.Client
@@ -43,9 +58,9 @@ func NewImagePuller(client client.Client, scheme *runtime.Scheme) ImagePuller {
 func (ipi *imagePullerImpl) CreatePullPod(ctx context.Context, imageSpec *kmmv1beta1.ModuleImageSpec,
 	micObj *kmmv1beta1.ModuleImagesConfig) error {
 
-	restartPolicy := v1.RestartPolicyOnFailure
+	pullPodTypeLabeValue := pullPodUntilSuccess
 	if imageSpec.Build != nil || imageSpec.Sign != nil {
-		restartPolicy = v1.RestartPolicyNever
+		pullPodTypeLabeValue = pullPodTypeOneTime
 	}
 
 	imagePullSecrets := []v1.LocalObjectReference{}
@@ -59,7 +74,7 @@ func (ipi *imagePullerImpl) CreatePullPod(ctx context.Context, imageSpec *kmmv1b
 			Namespace:    micObj.Namespace,
 			Labels: map[string]string{
 				moduleImageLabelKey: micObj.Name,
-				imageLabelKey:       imageSpec.Image,
+				pullPodTypeLabelKey: pullPodTypeLabeValue,
 			},
 		},
 		Spec: v1.PodSpec{
@@ -70,7 +85,7 @@ func (ipi *imagePullerImpl) CreatePullPod(ctx context.Context, imageSpec *kmmv1b
 					Command: []string{"/bin/sh", "-c", "exit 0"},
 				},
 			},
-			RestartPolicy:    restartPolicy,
+			RestartPolicy:    v1.RestartPolicyNever,
 			ImagePullSecrets: imagePullSecrets,
 		},
 	}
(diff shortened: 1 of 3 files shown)


internal/pod/imagepuller.go
```diff
@@ -107,13 +122,47 @@ func (ipi *imagePullerImpl) ListPullPods(ctx context.Context, micObj *kmmv1beta1
...
+	case v1.PodRunning:
+		return PullImageInProcess
+	case v1.PodPending:
+		// no container statuses yet, the pod is just starting to pull images
+		if pod.Status.ContainerStatuses == nil {
+			return PullImageInProcess
+		}
+
+		// no wating status, the pull process is still in progress
+		if pod.Status.ContainerStatuses[0].State.Waiting == nil {
```
TomerNewman, line 151: A question, I see that when we go to this line we assume already `ContainerStatuses` is not nil, but can it be an empty slice? and if not, can we assume `State` has already `Waiting` initialized?

internal/pod/imagepuller.go
```diff
@@ -12,6 +12,26 @@ import (
...
+	imagePullBackOffReason = "ImagePullBackOff"
+	errImagePullReason     = "ErrImagePull"
+
+	moduleImageLabelKey = "kmm.node.kubernetes.io/module-image-config"
+	pullPodTypeLabelKey = "kmm.node.kubernetes.io/pull-pod-type"
+
+	pullerContainerName = "puller"
+
+	pullPodTypeOneTime  = "one-time-pull"
+	pullPodUntilSuccess = "until-success"
```
ybettan, line 32: How about we enum those as well?

internal/pod/imagepuller.go
```diff
@@ -43,9 +58,9 @@ func NewImagePuller(client client.Client, scheme *runtime.Scheme) ImagePuller {
 func (ipi *imagePullerImpl) CreatePullPod(ctx context.Context, imageSpec *kmmv1beta1.ModuleImageSpec,
 	micObj *kmmv1beta1.ModuleImagesConfig) error {
 
-	restartPolicy := v1.RestartPolicyOnFailure
+	pullPodTypeLabeValue := pullPodUntilSuccess
 	if imageSpec.Build != nil || imageSpec.Sign != nil {
-		restartPolicy = v1.RestartPolicyNever
+		pullPodTypeLabeValue = pullPodTypeOneTime
```
ybettan, line 63: Shouldn't it be the opposite? That we run once if there are not build/sign and continue trying if we can build?

internal/pod/imagepuller.go
```diff
@@ -107,13 +122,47 @@ func (ipi *imagePullerImpl) ListPullPods(ctx context.Context, micObj *kmmv1beta1
...
 	for i, pod := range pods {
-		if image == pod.Labels[imageLabelKey] {
+		if image == pod.Spec.Containers[0].Image {
 			return &pods[i]
 		}
 	}
 	return nil
 }
 
 func (ipi *imagePullerImpl) GetPullPodImage(pod v1.Pod) string {
```
ybettan, line 132: Do we really need this public method? It doesn't give us anything that we can have access to in a single like outside of this package.

internal/pod/imagepuller.go
yevgeny-shnaidman: they are just values for the labels, so i don't see any need to define a type for them

internal/pod/imagepuller.go
```diff
@@ -12,6 +12,26 @@ import (
...
+type PullPodStatus string
+
+const (
+	PullImageFailed        PullPodStatus = "pullFailed"
+	PullImageSuccess       PullPodStatus = "pullSuccess"
+	PullImageInProcess     PullPodStatus = "pullInProcess"
+	PullImageUnexpectedErr PullPodStatus = "unexpectedError"
+
+	imagePullBackOffReason = "ImagePullBackOff"
+	errImagePullReason     = "ErrImagePull"
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Those are also just values of the reason field inside container statuses

internal/pod/imagepuller.go
yevgeny-shnaidman: actually  - no. In case there is a buld/sign present, then once the pull fails, we can stop trying to pull and go to building/signing images. In case there are no build/sign, then we need to try the pulling indefinitly

internal/pod/imagepuller.go
yevgeny-shnaidman: image puller hides the implementation of the pull pod from the calling layer (in this case MBSC). Why should MBSC care about the details of how to get the status from pod?

internal/pod/imagepuller.go
```diff
@@ -107,13 +122,47 @@ func (ipi *imagePullerImpl) ListPullPods(ctx context.Context, micObj *kmmv1beta1
...
+		pullPodType := pod.GetLabels()[pullPodTypeLabelKey]
+		// if pod is targeted to wait till the end - return the InProgress
+		if pullPodType == pullPodUntilSuccess {
+			return PullImageInProcess
+		}
+
+		if pod.Status.ContainerStatuses[0].State.Waiting.Reason == imagePullBackOffReason ||
+			pod.Status.ContainerStatuses[0].State.Waiting.Reason == errImagePullReason {
+			return PullImageFailed
+		}
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

internal/pod/imagepuller.go
ybettan: If those are the only 2 values allowed for this label, then I think it should be validated through enums.

internal/pod/imagepuller.go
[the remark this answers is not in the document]
ybettan: If those are the only 2 values allowed for this label, then I think it should be validated through enums.

internal/pod/imagepuller.go
ybettan: Ohh right, and then we update the MIC status based on the MBSC status and not based on the pull pods.

internal/pod/imagepuller.go
ybettan: We are not getting the status from the pod but the image from the spec. If you prefer to keep it as is I don't mind but you know how I feel about primitive getters/setters :)

internal/pod/imagepuller.go
```diff
@@ -107,13 +122,46 @@ func (ipi *imagePullerImpl) ListPullPods(ctx context.Context, micObj *kmmv1beta1
...
+
+		// no wating status, the pull process is still in progress
+		if pod.Status.ContainerStatuses[0].State.Waiting == nil {
+			return PullImageInProcess
+		}
+
+		pullPodType := pod.GetLabels()[pullPodTypeLabelKey]
+		// if pod is targeted to wait till the end - return the InProgress
+		if pullPodType == pullPodUntilSuccess {
+			return PullImageInProcess
```
ybettan, line 158: How will the customer knows that he forgot to push the image for example if we return a "in-progress" value. Shouldnt we return a better status for him to understand that he forgot to push the image?

internal/pod/imagepuller.go
yevgeny-shnaidman: what do you mean by enums? in golang  it is just consts.  defining a type for a label value will means conversions, which will make the code harder to read

internal/pod/imagepuller.go
yevgeny-shnaidman: sorry, my mistake! It makes the code simplier in the mbsc controller

internal/pod/imagepuller.go
yevgeny-shnaidman: we can add a more explanatory value into the MBSC status ("image is being pulled"), but it should not be in this package

ybettan: We will merge it as is for now because the is the behavior in our current code and we will have a task to check if replacing it with `imageDoesNotExist` is a better option.

PR #1049, opened 2025-03-12
Introducing MBSC controller

yevgeny-shnaidman:
MBSC controller is responsible for handling building and signing of the images based on the Spec of the MBSC object
The flow is as following:
1. get the running build or sign pods and update the MBSC status based on those pods
2. process the spec of the MBSC object. For each image and action verify if it has already been completed successfully. If not - call the buildsign package, which will know how to handle new requests, or updates to the existing pods
3. run garbage-collect - delete succeesfully completed build/sign pods.


internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+	v1 "k8s.io/api/core/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const MBSCReconcilerName = "MBSCReconciler"
+
+// mbscReconciler reconciles a Module object
```
ybettan, line 39:
```suggestion
// mbscReconciler reconciles a ModuleBuildSignConfig object
```

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modulebuildsignconfigs,verbs=get;list;watch;update;patch;create
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modulebuildsignconfigs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=pods,verbs=create;list;watch;delete
+
+// Reconcile lists all nodes and looks for kernels that match its mappings.
+// For each mapping that matches at least one node in the cluster, it creates a DaemonSet running the container image
+// on the nodes with a compatible kernel.
```
ybettan, line 74: I believe this comment was copied by mistake :)

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+func (r *mbscReconciler) Reconcile(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) (ctrl.Result, error) {
+	res := ctrl.Result{}
+
+	logger := log.FromContext(ctx)
+
+	if mbscObj.GetDeletionTimestamp() != nil {
+		// [TODO] delete build/sign pods
```
ybettan, line 81: Why don't we do it as part of this PR?

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+		buildSignAPI: buildSignAPI,
+		mbscAPI:      mbscAPI,
+	}
+}
+
+func (mrh *mbscReconcilerHelper) updateStatusByBuilds(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
+	errs := make([]error, 0, len(mbscObj.Spec.Images))
+	patchFrom := client.MergeFrom(mbscObj.DeepCopy())
+	for _, imageSpec := range mbscObj.Spec.Images {
+		status, err := mrh.buildSignAPI.GetStatus(ctx, mbscObj.Name, mbscObj.Namespace, imageSpec.ModuleImageSpec.KernelVersion, imageSpec.Action, mbscObj)
```
ybettan, line 130: How about instead passing `mbscObj, `mbscObj.Name` and `mbscObj.Namespace` we just pass `mbscObj`?

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+		errs = append(errs, err)
+		if err != nil || status == kmmv1beta1.BuildOrSignStatus("") {
```
ybettan, line 132: Although the logic is the same, let's put line 131 inside the `if` as there is no point of appending all `nil` errors by default.

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+		imageStatus := mrh.mbscAPI.GetImageStatus(mbscObj, imageSpec.Image, imageSpec.Action)
+		if imageStatus == kmmv1beta1.ActionSuccess {
+			// in case action succeeded - skip to the next image. Otherwise, action had not been handled yet, or is being handled, or already failed.
+			// in that case the Sync API will take care of what is needed to be done
+			continue
+		}
+		mld := createMLD(mbscObj, &imageSpec.ModuleImageSpec)
+		err := mrh.buildSignAPI.Sync(ctx, mld, true, mbscObj, imageSpec.Action)
+		errs = append(errs, err)
+		if err != nil {
```
ybettan, line 157: If we already have a `if err != nil` block then let's put the previous line in it.

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+			logger.Info(utils.WarnString("sync for image %s, action %s failed: %v"), imageSpec.Image, imageSpec.Action, err)
+		}
+		errs = append(errs, err)
```
ybettan, line 160:
```suggestion
```

Dup of line 156

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return errors.Join(errs...)
+}
+
+func (mrh *mbscReconcilerHelper) garbageCollect(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
+	logger := log.FromContext(ctx)
+
+	// Garbage collect for successfully finished build pods
+	deleted, err := mrh.buildSignAPI.GarbageCollect(ctx, mbscObj.Name, mbscObj.Namespace, kmmv1beta1.BuildImage, mbscObj)
```
ybettan, line 169: Can we just pass `mbscObj` instead of sending the name/namespace separately?

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: we have the same issue in the MIC controller also. I would prefer to split PRs between 2 different flows:
1) the normal flow
2) deletion flow

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: There are pro and cons, but in this case i think that having a more "elaborate" API allows you better understand what the functions needs. Looking at API you can see that it will get the Status based on name, namespace, kernel version etc'. IF we just pass MBSC object, then developer will need to go into the code to understand the parameters that will be used.

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: i will fix it, but notice that the nil errors will still get appended, since the condition is either an error or empty status, which will probably be more common

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (mrh *mbscReconcilerHelper) processImagesSpecs(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
+	logger := log.FromContext(ctx)
+	errs := make([]error, 0, len(mbscObj.Spec.Images))
+	for _, imageSpec := range mbscObj.Spec.Images {
+		imageStatus := mrh.mbscAPI.GetImageStatus(mbscObj, imageSpec.Image, imageSpec.Action)
+		if imageStatus == kmmv1beta1.ActionSuccess {
+			// in case action succeeded - skip to the next image. Otherwise, action had not been handled yet, or is being handled, or already failed.
+			// in that case the Sync API will take care of what is needed to be done
+			continue
```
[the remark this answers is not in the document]
yevgeny-shnaidman: Then the customer will have to define a different image in the KMM Module.

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mbsc_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: fixed

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: same comment as for previous suggestion. In addition, notice that the last parameter (mbscObj) is actually an owner, so i feel that passing mbsc object for the sake of its data, and then using the object in itself as an "owner" will be confusing

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+	// Garbage collect for successfully finished build pods
+	deleted, err := mrh.buildSignAPI.GarbageCollect(ctx, mbscObj.Name, mbscObj.Namespace, kmmv1beta1.BuildImage, mbscObj)
+	if err != nil {
+		return fmt.Errorf("could not garbage collect build objects: %v", err)
+	}
+
+	logger.Info("Garbage-collected Build objects", "names", deleted)
+
+	// Garbage collect for successfully finished sign pods
+	deleted, err = mrh.buildSignAPI.GarbageCollect(ctx, mbscObj.Name, mbscObj.Namespace, kmmv1beta1.SignImage, mbscObj)
```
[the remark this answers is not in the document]
yevgeny-shnaidman: same answer as above

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,196 @@
...
+	err := mrh.client.Status().Patch(ctx, mbscObj, patchFrom)
+	errs = append(errs, err)
+	return errors.Join(errs...)
+}
+
+func (mrh *mbscReconcilerHelper) processImagesSpecs(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
+	logger := log.FromContext(ctx)
+	errs := make([]error, 0, len(mbscObj.Spec.Images))
+	for _, imageSpec := range mbscObj.Spec.Images {
+		imageStatus := mrh.mbscAPI.GetImageStatus(mbscObj, imageSpec.Image, imageSpec.Action)
```
TomerNewman, line 145: where `GetImageStatus` is implemented? can not find it

yevgeny-shnaidman: it is part of MBSC package API

TomerNewman: @yevgeny-shnaidman I only saw

```
type mbscReconcilerHelperAPI interface {
	updateStatusByBuilds(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error
	processImagesSpecs(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error
	garbageCollect(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error
}
```

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,196 @@
...
+		Named(MBSCReconcilerName).
+		Complete(
+			reconcile.AsReconciler[*kmmv1beta1.ModuleBuildSignConfig](mgr.GetClient(), r),
+		)
+}
+
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modulebuildsignconfigs,verbs=get;list;watch;update;patch;create
+//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modulebuildsignconfigs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
+//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list;watch
```
ybettan, line 69: Can we combine those 2 resources since they have the same verbs?

internal/controllers/mbsc_reconciler.go
ybettan: Then how about we make it a bit more clear, like:
```
if err != nil {
     errs = append(errs, err)
      continue
}

if status == kmmv1beta1.buildOrSignStatus("") {
     continue
}
```

?

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,196 @@
...
+
+func newMBSCReconcilerHelper(client client.Client, buildSignAPI buildsign.Manager, mbscAPI mbsc.MBSC) mbscReconcilerHelperAPI {
+	return &mbscReconcilerHelper{
+		client:       client,
+		buildSignAPI: buildSignAPI,
+		mbscAPI:      mbscAPI,
+	}
+}
+
+func (mrh *mbscReconcilerHelper) updateStatusByBuilds(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
```
ybettan, line 123:
```suggestion
func (mrh *mbscReconcilerHelper) updateStatusByPods(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
```

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (mrh *mbscReconcilerHelper) processImagesSpecs(ctx context.Context, mbscObj *kmmv1beta1.ModuleBuildSignConfig) error {
+	logger := log.FromContext(ctx)
+	errs := make([]error, 0, len(mbscObj.Spec.Images))
+	for _, imageSpec := range mbscObj.Spec.Images {
+		imageStatus := mrh.mbscAPI.GetImageStatus(mbscObj, imageSpec.Image, imageSpec.Action)
+		if imageStatus == kmmv1beta1.ActionSuccess {
+			// in case action succeeded - skip to the next image. Otherwise, action had not been handled yet, or is being handled, or already failed.
+			// in that case the Sync API will take care of what is needed to be done
```
ybettan, line 148: Took me some time and investigation to understand those comments. In case the pod has failed then it just hang until the use make something that changes the pod?

WDYT about moving the check if the action has already succeeded inside `Sync` for simplicity?

internal/controllers/mbsc_reconciler.go
ybettan: Same comment as before.

The function is garbage collecting pods created for the MBSC so I don't see any issue to only pass the MBSC and inside the function use it for data and as the owner. I think it makes more sense then how it is written now.

internal/controllers/mbsc_reconciler.go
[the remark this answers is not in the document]
ybettan: https://github.com/kubernetes-sigs/kernel-module-management/pull/1049/files#r2009086077

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: I would prefer to keep them separate, for readability sake. It already happened to me that i missed the resource, because it was in the same line as another resource

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: Sync can only look at the pods. In case build succeeded, by the time Sync will look at it,  the pod might already have been deleted, and in that case the Sync will run the build/sign again, which is not what we want

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: in Module controller we pass the module largely to the helper interface, which is tightly connected to the module reconciler. In any other interface it is always better to pass specific parameter, and not some encapsulating object  (unless it is a LOT of parameters). Lets say we will have some other package trying to use buildisign package: does it mean that it will have to create an MBSC object, instead of just passing the data it already has?

internal/controllers/mbsc_reconciler.go
```diff
@@ -0,0 +1,196 @@
...
+			continue
+		}
+		mld := createMLD(mbscObj, &imageSpec.ModuleImageSpec)
+		err := mrh.buildSignAPI.Sync(ctx, mld, true, mbscObj, imageSpec.Action)
```
ybettan, line 152: Change `mbscObj` to the owner part only.

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: Fixed

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: Fixed

internal/controllers/mbsc_reconciler.go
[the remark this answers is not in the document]
yevgeny-shnaidman: Fixed

internal/controllers/mbsc_reconciler.go
yevgeny-shnaidman: Fixed

PR #1060, opened 2025-03-27
Switching KMM to work with the new build/sign flow

yevgeny-shnaidman:
changing the manager's main to use MIC and MBSC reconcilers instead of BuildSignReconciler

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -22,26 +22,23 @@ import (
 	"os"
 	"strconv"
 
+	"github.com/kubernetes-sigs/kernel-module-management/internal/mbsc"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/mic"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/pod"
 
 	"github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/api/v1beta2"
-	buildpod "github.com/kubernetes-sigs/kernel-module-management/internal/build/pod"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/buildsign"
 	buildsignpod "github.com/kubernetes-sigs/kernel-module-management/internal/buildsign/pod"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/cmd"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/config"
-	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/controllers"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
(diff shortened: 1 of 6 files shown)


cmd/manager/main.go
```diff
@@ -168,6 +167,10 @@ func main() {
 		cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.NodeLabelModuleVersionReconcilerName)
 	}
 
+	if err = controllers.NewMICReconciler(client, micAPI, mbscAPI, imagePullerAPI, scheme).SetupWithManager(mgr); err != nil {
```
ybettan, line 170: Why does the MIC controller needs the `mbscAPI` interface? I don't see it being used in the MIC controller.

cmd/manager/main.go
```diff
@@ -179,30 +182,38 @@ func main() {
...
+				client,
+				buildAPI,
+				signAPI,
+				kernelAPI,
+				filterAPI,
+				nodeAPI)
+			if err = bsc.SetupWithManager(mgr, constants.KernelLabel); err != nil {
+				cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.BuildSignReconcilerName)
+			}
+		*/
```
ybettan, line 211: How about we just remove it?

cmd/manager/main.go
```diff
@@ -215,12 +226,15 @@ func main() {
...
+			preflightStatusUpdaterAPI := preflight.NewStatusUpdater(client)
+			preflightAPI := preflight.NewPreflightAPI(client, buildAPI, signAPI, registryAPI, preflightStatusUpdaterAPI, kernelAPI)
 
-		if err = controllers.NewPreflightValidationReconciler(client, filterAPI, metricsAPI, preflightStatusUpdaterAPI, preflightAPI).SetupWithManager(mgr); err != nil {
-			cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.PreflightValidationReconcilerName)
-		}
+			if err = controllers.NewPreflightValidationReconciler(client, filterAPI, metricsAPI, preflightStatusUpdaterAPI, preflightAPI).SetupWithManager(mgr); err != nil {
+				cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.PreflightValidationReconcilerName)
+			}
+		*/
```
ybettan, line 237: Can we do it in the same PR? If you insist on keeping small changes then I would do a multi-commit PR but I think it makes sense to put all of the following inside the same PR and it will be easier for me to review as well:
* Create the new controllers
* Remove the old controllers
* Use the new buildsign pkg everywhere needed
* Remove the old build/sign packages

cmd/manager/main.go
yevgeny-shnaidman: Fixed

cmd/manager/main.go
yevgeny-shnaidman: it would be better to do it in a separate PR: preflight is not a major priority, and i am not yet sure how many changes i would need to make in order to make it work.

cmd/manager/main.go
yevgeny-shnaidman: it reads the MBSC status

PR #1061, opened 2025-03-27
BootID support for KMM

sriram-30:
KMM loads driver again on node reboot once it detects Node moving from Ready to NotReady back to Ready.
In some cases, like VM node reboot, reboot happens very fast and k8s doesn't report NotReady because of its reporting frequency. Node status, however, has bootID which changes after reboot.
Added support for KMM to detect node reboot using this bootID in cases where reboot happens fast and NotReady state is not reached

Verified that kmm is detecting the boot id change and triggering worker after node is rebooted

```
vm@ubuntu2204:~/kernel-module-management$ kubectl logs -n kube-amd-gpu   amd-gpu-operator-kmm-controller-6fd784d4b4-nfx4z | grep "Detected"
I0328 02:43:46.316244       1 nmc_reconciler.go:348] "Detected BootId change" logger="kmm" controller="NodeModulesConfig" controllerGroup="kmm.sigs.x-k8s.io" controllerKind="NodeModulesConfig" NodeModulesConfig="worker-2" namespace="" name="worker-2" reconcileID="5d8875ad-ed80-4496-a006-49b6502eaacd" module="kube-amd-gpu/test-deviceconfig"
I0328 02:43:56.734541       1 nmc_reconciler.go:348] "Detected BootId change" logger="kmm" controller="NodeModulesConfig" controllerGroup="kmm.sigs.x-k8s.io" controllerKind="NodeModulesConfig" NodeModulesConfig="worker-1" namespace="" name="worker-1" reconcileID="5f2a7d67-b7d9-4596-96d4-6800598add7d" module="kube-amd-gpu/test-deviceconfig"
```

BootId present in nmc:

```
status:
  modules:
  - bootId: ab95e5f6-97e9-4d78-bff6-060f11f15d87
    config:
      containerImage: registry.test.pensando.io:5000/sriram:ubuntu-22.04-5.15.0-134-generic-6.3.2
      imagePullPolicy: ""
      insecurePull: true
      kernelVersion: 5.15.0-134-generic
      modprobe:
        args: {}
        dirName: /opt
        firmwarePath: firmwareDir/updates
        moduleName: amdgpu
        modulesLoadingOrder:
        - amdgpu
        - amdttm
        - amdkcl
    lastTransitionTime: "2025-03-28T02:47:04Z"
    name: test-deviceconfig
    namespace: kube-amd-gpu
    serviceAccountName: amd-gpu-operator-kmm-module-loader
```

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -90,7 +90,12 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 		return reconcile.Result{}, fmt.Errorf("could not get NodeModuleState %s: %v", req.NamespacedName, err)
 	}
 
-	if err := r.helper.SyncStatus(ctx, &nmcObj); err != nil {
+	node := v1.Node{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: nmcObj.Name}, &node); err != nil {
+		return ctrl.Result{}, fmt.Errorf("could not get node %s: %v", nmcObj.Name, err)
+	}
+
+	if err := r.helper.SyncStatus(ctx, &nmcObj, &node); err != nil {
 		return reconcile.Result{}, fmt.Errorf("could not reconcile status for NodeModulesConfig %s: %v", nmcObj.Name, err)
 	}
 
@@ -101,11 +106,6 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	for i := 0; i < len(nmcObj.Status.Modules); i++ {
 		status := nmcObj.Status.Modules[i]
 		statusMap[status.Namespace+"/"+status.Name] = &nmcObj.Status.Modules[i]
-	}
-
-	node := v1.Node{}
-	if err := r.client.Get(ctx, types.NamespacedName{Name: nmcObj.Name}, &node); err != nil {
-		return ctrl.Result{}, fmt.Errorf("could not get node %s: %v", nmcObj.Name, err)
 	}
 
 	errs := make([]error, 0, len(nmcObj.Spec.Modules)+len(nmcObj.Status.Modules))
@@ -220,7 +220,7 @@ type nmcReconcilerHelper interface {
 	ProcessModuleSpec(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, spec *kmmv1beta1.NodeModuleSpec, status *kmmv1beta1.NodeModuleStatus, node *v1.Node) error
 	ProcessUnconfiguredModuleStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, status *kmmv1beta1.NodeModuleStatus, node *v1.Node) error
 	RemovePodFinalizers(ctx context.Context, nodeName string) error
-	SyncStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig) error
+	SyncStatus(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) error
 	UpdateNodeLabels(ctx context.Context, nmc *kmmv1beta1.NodeModulesConfig, node *v1.Node) ([]types.NamespacedName, []types.NamespacedName, error)
 	RecordEvents(node *v1.Node, loadedModules, unloadedModules []types.NamespacedName)
 }
@@ -374,7 +374,7 @@ func (h *nmcReconcilerHelperImpl) ProcessModuleSpec(
 			return h.podManager.CreateLoaderPod(ctx, nmcObj, spec)
 		}
 
-		if h.nodeAPI.NodeBecomeReadyAfter(node, status.LastTransitionTime) {
+		if h.nodeAPI.IsNodeRebooted(node, status.BootId) {
 			logger.Info("node has been rebooted and become ready after kernel module was loaded; creating loader Pod")
 			return h.podManager.CreateLoaderPod(ctx, nmcObj, spec)
 		}
@@ -425,7 +425,7 @@ func (h *nmcReconcilerHelperImpl) ProcessUnconfiguredModuleStatus(
 	/* node was rebooted, spec not set so no kernel module is loaded, no need to unload.
 	   it also fixes the scenario when node's kernel was upgraded, so unload pod will fail anyway
 	*/
-	if h.nodeAPI.NodeBecomeReadyAfter(node, status.LastTransitionTime) {
+	if h.nodeAPI.IsNodeRebooted(node, status.BootId) {
 		logger.Info("node was rebooted and spec is missing: delete the status to allow Module CR unload, if needed")
 		patchFrom := client.MergeFrom(nmcObj.DeepCopy())
 		nmc.RemoveModuleStatus(&nmcObj.Status.Modules, status.Namespace, status.Name)
@@ -493,7 +493,7 @@ func (h *nmcReconcilerHelperImpl) RemovePodFinalizers(ctx context.Context, nodeN
 	return errors.Join(errs...)
 }
 
-func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1beta1.NodeModulesConfig) error {
+func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1beta1.NodeModulesConfig, node *v1.Node) error {
 	logger := ctrl.LoggerFrom(ctx)
 
 	logger.Info("Syncing status")
@@ -579,6 +579,7 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 				FinishedAt
 
 			status.LastTransitionTime = podLTT
+			status.BootId = node.Status.NodeInfo.BootID
 
 			nmc.SetModuleStatus(&nmcObj.Status.Modules, *status)
 
--- a/api/v1beta1/nodemodulesconfig_types.go
+++ b/api/v1beta1/nodemodulesconfig_types.go
@@ -68,6 +68,8 @@ type NodeModuleStatus struct {
 	Config ModuleConfig `json:"config,omitempty"`
 	//+optional
 	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+	//+optional
+	BootId string `json:"bootId,omitempty"`
 }
 
 // NodeModuleConfigStatus is the most recently observed status of the KMM modules on node.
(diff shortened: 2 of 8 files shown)


internal/controllers/nmc_reconciler.go
```diff
@@ -374,7 +374,12 @@ func (h *nmcReconcilerHelperImpl) ProcessModuleSpec(
 			return h.podManager.CreateLoaderPod(ctx, nmcObj, spec)
 		}
 
-		if h.nodeAPI.NodeBecomeReadyAfter(node, status.LastTransitionTime) {
+		currentBootId := node.Status.NodeInfo.BootID
+		bootIdChanged := status.BootId != currentBootId
+		if bootIdChanged {
+			logger.Info("Detected BootId change")
+		}
+		if bootIdChanged || h.nodeAPI.NodeBecomeReadyAfter(node, status.LastTransitionTime) {
```
yevgeny-shnaidman, line 382: 1. I don't think that we need to check h.nodeAPI.NodeBecomeReadyAfter:  bootId is surely was changed in case h.nodeAPI.NodeBecomeReadyAfter, so the second condition of "if" will never be reached
2. lets move the verification of bootId change to the node pkg:  we can create an interface function like:
isNodeRebooted(node, status.BootID)
and it will check 2 conditions: that the bootId was changes and that the node is in the ready state

internal/controllers/nmc_reconciler.go
```diff
@@ -578,6 +583,12 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 				Terminated.
 				FinishedAt
 
+			node := &v1.Node{}
```
yevgeny-shnaidman, line 586: we are actually obtaining the node related to the NMC at:
https://github.com/kubernetes-sigs/kernel-module-management/blob/56eddaf7b61821ccf769a55ddb874cab9796abb9/internal/controllers/nmc_reconciler.go#L106
So we  can obtain it before calling SyncStatus, and pass it as an input parameter to the SyncStatus function

internal/controllers/nmc_reconciler.go
```diff
@@ -374,7 +374,12 @@ func (h *nmcReconcilerHelperImpl) ProcessModuleSpec(
 			return h.podManager.CreateLoaderPod(ctx, nmcObj, spec)
 		}
 
-		if h.nodeAPI.NodeBecomeReadyAfter(node, status.LastTransitionTime) {
+		currentBootId := node.Status.NodeInfo.BootID
+		bootIdChanged := status.BootId != currentBootId
+		if bootIdChanged {
+			logger.Info("Detected BootId change")
```
ybettan, line 380:
```suggestion
			logger.Info("Detected BootId change - node has rebooted")
```

internal/controllers/nmc_reconciler.go
```diff
@@ -578,6 +583,12 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 				Terminated.
 				FinishedAt
 
+			node := &v1.Node{}
+			if err := h.client.Get(ctx, types.NamespacedName{Name: nmcObj.Name}, node); err != nil {
+				logger.Error(err, "Failed to get node object", "node", nmcObj.Name)
+			} else {
+				status.BootId = node.Status.NodeInfo.BootID
```
ybettan, line 590: Since `SyncStatus` is called before `ProcessModuleSpec` - if we update the status here then when we get to `ProcessModuleSpec` then https://github.com/kubernetes-sigs/kernel-module-management/pull/1061/files#diff-4abf41a7395eeba9459df6b2fe59c88e93e33e83dd683d40878c61651de912f6R378 will always be false. Am I missing something?

internal/controllers/nmc_reconciler.go
ybettan: We should probably also remove `NodeBecomeReadyAfter` from that interface as it won't be used anymore.

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: @ybettan it will be false in the Reconcile loop , when we are handling the successful termination of the worker pod. But when the reconcile loop will be called due to the Node event(node was rebooted), the bootId in the status will be the old one (since the pod does not exists), and then the condition in the ProcessModuleSpec will be true

internal/controllers/nmc_reconciler.go
sriram-30: Sure, that makes sense, will make that change

internal/controllers/nmc_reconciler.go
sriram-30: Thanks for the suggestions! Will make those changes as well

yevgeny-shnaidman:
@sriram-30 looks good, just a couple of comments:
1) I think we have missed one more place where the new IsNodeRebooted function should be used:
    [need to be replaced](https://github.com/kubernetes-sigs/kernel-module-management/blob/b25257518ddabe147dcc81b1700159c0b6dcf770/internal/controllers/nmc_reconciler.go#L428)
2)  please squash your commits into one

sriram-30:
> @sriram-30 looks good, just a couple of comments:
> 
> 1. I think we have missed one more place where the new IsNodeRebooted function should be used:
>    [need to be replaced](https://github.com/kubernetes-sigs/kernel-module-management/blob/b25257518ddabe147dcc81b1700159c0b6dcf770/internal/controllers/nmc_reconciler.go#L428)
> 2. please squash your commits into one


Thanks!
1) is already taken care of. As you can see, that line was already updated to use the new function:
https://github.com/kubernetes-sigs/kernel-module-management/pull/1061/files#diff-4abf41a7395eeba9459df6b2fe59c88e93e33e83dd683d40878c61651de912f6R428
2) Have taken care of this now, it is one commit

yevgeny-shnaidman:
/approve
@ybettan if no further comments, please lgtm

ybettan:
/lgtm
@sriram-30 in order for the PR to merge you need to pass the `EasyCLA` job - you can click on the job link and it will re-direct you to a short guide.

sriram-30:
> /lgtm @sriram-30 in order for the PR to merge you need to pass the `EasyCLA` job - you can click on the job link and it will re-direct you to a short guide.

Done now, thanks!

sriram-30:
> Sorry, I just noticed the commit message. @sriram-30 Can you please update the commit message to a descriptive one and please (I will re-approve) ? Something like
> 
> > KMM is checking if a node has rebooted by inspecting the `Ready` timestamp on the node and check if it newer than the last Ready timestamp recorded.
> > Kubernetes has a grace period in which if a node stop reporting heartbeats it is then marked by the k8s API server as not ready.
> > In some cases, the reboot is so fast that the node become Ready again before the k8s API has even noticed it went down, and in those cases we need to make sure KMM catches it and reload the kmod to the node - this is being done by comparing the node's `status.nodeInfo.bootID` ,which is unique per reboot, with the last recorded value.

Sure, I've added that description to the commit message now and repushed

sriram-30:
> Sorry, I just noticed the commit message. @sriram-30 Can you please update the commit message to a descriptive one and please (I will re-approve) ? Something like
> 
> > KMM is checking if a node has rebooted by inspecting the `Ready` timestamp on the node and check if it newer than the last Ready timestamp recorded.
> > Kubernetes has a grace period in which if a node stop reporting heartbeats it is then marked by the k8s API server as not ready.
> > In some cases, the reboot is so fast that the node become Ready again before the k8s API has even noticed it went down, and in those cases we need to make sure KMM catches it and reload the kmod to the node - this is being done by comparing the node's `status.nodeInfo.bootID` ,which is unique per reboot, with the last recorded value.

Sure, I've added that description to the commit message now and repushed

sriram-30:
> @sriram-30 I am sorry I am bugging. Can you please keep short lines in the commit message to be more readable? I promise this is the last piece :)
> 
> ```
> commit 377de92b3213a32497cc7119649eb11202d95fc3 (refs/heads/test)
> Author: vm <sriramr2230@gmail.com>
> Date:   Thu Mar 27 16:54:03 2025 +0000
> 
>     BootID support for KMM
> 
>     BootID KMM support:
>     KMM is checking if a node has rebooted by inspecting the Ready timestamp on the node and check if it newer than the last Ready timestamp recorded.
> 
>     Kubernetes has a grace period in which if a node stop reporting heartbeats it is then marked by the k8s API server as not ready.
> 
>     In some cases, the reboot is so fast that the node become Ready again before the k8s API has even noticed it went down, and in those cases we need to make sure KMM catches it and reload the kmod to the node - this is being done by comparing the node's status.nodeInfo.bootID ,which is unique per reboot, with the last recorded value.
> 
> commit 56eddaf7b61821ccf769a55ddb874cab9796abb9
> Author: Yoni Bettan <yonibettan@gmail.com>
> Date:   Thu Mar 27 09:45:02 2025 +0200
> 
>     Filtering out node heartbeats events for the NMC controller.
> 
>     Node heartbeats are there to let the k8s API server that the node is
>     still connected and functional and if not filtered, it will spam the
>     events the NMC controller gets.
> 
>     The NMC controller, is trying to garbage collect pods for NMCs that were
>     removed - it causes a constant reconciliation even when no Module, and
>     therefore no NMC, are applied to the cluster.
> 
>     This commit is fixing this issue.
> 
>     Signed-off-by: Yoni Bettan <yonibettan@gmail.com>
> 
> commit 9c4a309da7dc214a59c7cbf48bae88758227aade
> Author: Yevgeny Shnaidman <yshnaidm@redhat.com>
> Date:   Sun Mar 23 12:37:15 2025 +0200
> 
>     Hiding the buildsign porting layer initialization
> 
>     buildsign/pod is a porting layer, and as such we would like to minimize
>     itsexposure to the outer code. This PR make buildsign manager initialize
>     the maker, signer and buildisngpodmanager interfaces inside its own init
>     function, which will remove the exposure to the porting layer from main
> 
> commit 0ad7b3847d109ae772281e6d14b52598a6c043f4
> Author: Yevgeny Shnaidman <yshnaidm@redhat.com>
> Date:   Mon Mar 24 13:06:24 2025 +0200
> 
>     Aligning Sync API of the buildsign manager
> 
>     Moving the "owner" input parameter to be the last parameter in order to
>     be aligned with the rest of buildsign manager APIs
> 
> commit 6ef41572a32603b280dd0ae13220c05df4ba0d07
> Author: dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
> Date:   Thu Mar 13 02:16:49 2025 +0000
> 
>     Bump golang.org/x/net from 0.35.0 to 0.36.0
> ```

No problem :)
Have made that change now to the commit message!

ybettan:
Github runners are being under maintenance - hopefully, CI will be working again soon.
https://github.com/actions/runner-images/issues/11101

ybettan:
Fixes https://github.com/kubernetes-sigs/kernel-module-management/issues/1062

PR #1066, opened 2025-04-01
Removing `LastTransitionTime` from the NMC status.

ybettan:
Since we are now checking if a node has rebooted using the node's
`status.nodeInfo.bootID` then we don't need to compare the node's ready
time and the NMC last recorded node ready time anymore.

---

/assign @yevgeny-shnaidman (only the last commit is relevant)
/hold
It should only be merged after https://github.com/kubernetes-sigs/kernel-module-management/pull/1061

I have tested the upgrade flow and I didn't inspect an issue during the upgrade.


ybettan:
Github runners are being under maintenance - hopefully, CI will be working again soon.
https://github.com/actions/runner-images/issues/11101

PR #1070, opened 2025-04-03
Changing ImagePuller interface to more elaborate parameters

yevgeny-shnaidman:
Since ImagePuller interface is going to be used not only by MIC, but also by Preflight controller, the APIs should be changed to receive specific parameters, and not use ModuleImagesConfig or ImageSpec structures, that are alient to Preflight

### DIFF
--- a/internal/pod/imagepuller.go
+++ b/internal/pod/imagepuller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


internal/pod/imagepuller.go
```diff
@@ -55,33 +55,33 @@ func NewImagePuller(client client.Client, scheme *runtime.Scheme) ImagePuller {
...
 	pullPod := v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: micObj.Name + "-pull-pod-",
-			Namespace:    micObj.Namespace,
+			GenerateName: name + "-pull-pod-",
+			Namespace:    namespace,
 			Labels: map[string]string{
-				moduleImageLabelKey: micObj.Name,
-				pullPodTypeLabelKey: pullPodTypeLabeValue,
+				moduleImageLabelKey: name,
```
ybettan, line 76: `moduleImageLabelKey` should be changed to something more generic as it will now always be a MIC.

internal/pod/imagepuller.go
```diff
@@ -103,17 +103,17 @@ func (ipi *imagePullerImpl) DeletePod(ctx context.Context, pod *v1.Pod) error {
...
 }
 
-func (ipi *imagePullerImpl) ListPullPods(ctx context.Context, micObj *kmmv1beta1.ModuleImagesConfig) ([]v1.Pod, error) {
+func (ipi *imagePullerImpl) ListPullPods(ctx context.Context, name, namespace string) ([]v1.Pod, error) {
 
 	pl := v1.PodList{}
 
 	hl := client.HasLabels{pullPodTypeLabelKey}
-	ml := client.MatchingLabels{moduleImageLabelKey: micObj.Name}
+	ml := client.MatchingLabels{moduleImageLabelKey: name}
```
ybettan, line 111: Same here regarding `moduleImageLabelKey`

yevgeny-shnaidman: Fixed

PR #1077, opened 2025-04-09
Combined the Maker and Signer interfaces into BuildSignPodManager.

ybettan:
---

/assign @yevgeny-shnaidman

This is a multi commit (2 commits) PR. Each commit contains its own commit message and the commits will be merged as is (not squashed by Github).


ybettan:
/close

This is a dup created for testing purposes.

PR #1078, opened 2025-04-10
Refactor the way we get hash and build args for the build/sign flows.

ybettan:
---

/assign @yevgeny-shnaidman

This is a multi commit (2 commits) PR. Each commit contains its own commit message and the commits will be merged as is (not squashed by Github).

It contains 3 commits:
* Refactoring the way we get the hash annotation for build/sign.
* Refactoring the way we generate the container args for the build pod.
* Renaming the method object name for BuildSignPodManager.

### DIFF
--- a/internal/buildsign/pod/common.go
+++ b/internal/buildsign/pod/common.go
@@ -35,6 +35,22 @@ var tmpl = template.Must(
 func (pm *buildSignPodManager) podSpec(mld *api.ModuleLoaderData, containerImage string, pushImage bool) v1.PodSpec {
 
 	buildConfig := mld.Build
+
+	args := pm.containerArgs(mld, containerImage, mld.Build.BaseImageRegistryTLS, pushImage)
+	overrides := []kmmv1beta1.BuildArg{
+		{Name: "KERNEL_VERSION", Value: mld.KernelVersion},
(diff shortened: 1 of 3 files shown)


ybettan:
/override pull-kernel-module-management-check-commits-count

This is a multi-commit PR.

internal/buildsign/pod/common.go
```diff
@@ -35,6 +35,22 @@ var tmpl = template.Must(
 func (pm *buildSignPodManager) podSpec(mld *api.ModuleLoaderData, containerImage string, pushImage bool) v1.PodSpec {
```
yevgeny-shnaidman, line 35: we have "pm" here, and "mp" in buildsignpodmanager, which is very confusing. Lets reach the common name, it would be better to call it bspm

ybettan:
> why do we need 2 commits in this PR? They are unrelated.

Because it feels like having ~10 PRs to refactor a single file (cause it easier to refactor function by function) will bring a log of overhead. So I am trying to group them as much as possible cause they all fit in the "refactor the buildsign package for code re-usage" change.

internal/buildsign/pod/common.go
ybettan: Fixed.

ybettan:
/override pull-kernel-module-management-check-commits-count

This is a multi-commit PR.

PR #1079, opened 2025-04-10
Refactoring the way we generate volume/volumeMounts in build/sign flows.

ybettan:
To be used in the build flow as well in the build-sign flow.

---

/assign @yevgeny-shnaidman

### DIFF
--- a/internal/buildsign/pod/buildsignpodmanager.go
+++ b/internal/buildsign/pod/buildsignpodmanager.go
@@ -16,7 +16,6 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/api"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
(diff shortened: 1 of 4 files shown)


yevgeny-shnaidman:
1) i like it that the volume and volume mount are explicit :)
2) WDYT about moving the volumes/volumesmount to a common_volumes.go file? the common.go becomes a large file
3) why do we need to change sign package and delete the utils in this PR? It is a little bit confusing since we are already not using sign package. We can leave till we the PR where we delete build and sign packages

ybettan:
> i like it that the volume and volume mount are explicit :)

I am happy you liked it :)

> WDYT about moving the volumes/volumesmount to a common_volumes.go file? the common.go becomes a large file

I moved it to a new `volumes.go` file in the same directory.

> why do we need to change sign package and delete the utils in this PR? It is a little bit confusing since we are already not using sign package. We can leave till we the PR where we delete build and sign packages

Because at first I was using methods (instead of explicit volume/volumeMount list) so I have created new methods in the buildsign/pod package and removed them from utils.

PR #1091, opened 2025-04-24
Remove non KMM related tolerations from the NMC status

TomerNewman:
Until now, the  nmc status reflected all the worker-pod tolerations, including those unrelated to KMM.
This commit updates the behavior to get tolerations from a dedicated annotation instead, ensuring only KMM-specific tolerations appear in the status.

--- 

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -563,7 +563,14 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 					errs,
 					fmt.Errorf("%s: could not unmarshal the ModuleConfig from YAML: %v", podNSN, err),
 				)
-
+				continue
+			}
+			tolerationsAnnotation := h.podManager.GetTolerationsAnnotation(&p)
+			if err = yaml.UnmarshalStrict([]byte(tolerationsAnnotation), &status.Tolerations); err != nil {
+				errs = append(
+					errs,
+					fmt.Errorf("%s: could not unmarshal the ModuleConfig from YAML: %v", podNSN, err),
+				)
 				continue
 			}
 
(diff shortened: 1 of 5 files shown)


ybettan:
Can we please rename the commit/PR title to something like `Remove non KMM related tolerations from the NMC status`

internal/pod/workerpodmanager.go
```diff
@@ -220,6 +225,9 @@ func (wpmi *workerPodManagerImpl) UnloaderPodTemplate(ctx context.Context, nmc c
 	if err = setWorkerConfigAnnotation(pod, nms.Config); err != nil {
 		return nil, fmt.Errorf("could not set worker config: %v", err)
 	}
+	if err = setWorkerTolerationsAnnotation(pod, nms.Tolerations); err != nil {
```
ybettan, line 228: Something doesn't sums up for me. If we always set the worker's annotation based on the NMC spec, then why do we need the annotation at all? We can take it from the NMC spec and it will be the same. 

Neither one of the option doesn't reflect the "real" pod's tolerations.

internal/controllers/nmc_reconciler.go
```diff
@@ -563,15 +563,23 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 					errs,
 					fmt.Errorf("%s: could not unmarshal the ModuleConfig from YAML: %v", podNSN, err),
 				)
-
 				continue
 			}
+			tolerationsAnnotation := h.podManager.GetTolerationsAnnotation(&p)
+			if tolerationsAnnotation != "" {
```
yevgeny-shnaidman, line 569: do we need to check if tolerationsAnnotation != "" ?  won't Unmarshal knows to handle it by default?

internal/pod/workerpodmanager.go
TomerNewman: If the tolerations in nmc spec changes for some reason while the worker pod runs from A to B, then later the status will be based on tolerations B (because we are doing status=spec), even though the worker pod ran with tolerations A.

internal/controllers/nmc_reconciler.go
TomerNewman: @yevgeny-shnaidman you are right, when I applied Module without tolerations it made `null` on the pod's annotation.
This should have be the fix but it is not correct, so now I changed `setWorkerTolerationsAnnotation`.

internal/pod/workerpodmanager.go
ybettan: You are correct :)

PR #1092, opened 2025-04-27
Removed the `PodType` consts in the `buildsign` package.

ybettan:
We can directly use the API's `BuildImage` and `SignImage`. No need to use local consts and do the translation.

---

/assign @yevgeny-shnaidman


ybettan:
/hold

We will first remove the old `build`/`sign` packages and only then make this change in order to simplify it.

PR #1093, opened 2025-05-05
Update nmc node predicate to detect Ready transitions and meaningful changes

TomerNewman:
Previously, nmc controller filtered out all heartbeat-related updates by ignoring changes to node `status.conditions[*].lastHeartbeatTime`.

With this change:
- We now trigger reconciliation when:
  - `node.Spec` changes
  - `node.metadata.labels` change
  - `node.Status.NodeInfo `changes
  - The Node condition status transitions.
  
  ---
  
### DIFF
--- a/internal/filter/filter.go
+++ b/internal/filter/filter.go
@@ -2,16 +2,14 @@ package filter
 
 import (
 	"context"
-	"reflect"
-
 	"github.com/go-logr/logr"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 	v1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/kubectl/pkg/util/podutils"
 	clusterv1 "open-cluster-management.io/api/cluster/v1"
+	"reflect"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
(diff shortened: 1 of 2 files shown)


internal/filter/filter.go
```diff
@@ -150,48 +148,49 @@ func ListModulesForNMC(_ context.Context, obj client.Object) []reconcile.Request
...
-			for i, mf := range oldNodeCopy.ManagedFields {
-				if mf.Manager == "kubelet" {
-					oldNodeCopy.ManagedFields[i].Time = nil
-				}
-			}
-			for i, mf := range newNodeCopy.ManagedFields {
-				if mf.Manager == "kubelet" {
-					newNodeCopy.ManagedFields[i].Time = nil
-				}
+			if (oldReady == v1.ConditionFalse || oldReady == v1.ConditionUnknown) &&
```
yevgeny-shnaidman, line 174: We are only interested in a change in the Condition, so we can just compare old and new Status, and if they differ, then we want the even

TomerNewman: @yevgeny-shnaidman do we want to allow reconcile when Node becomes NotReady?

internal/filter/filter.go
```diff
@@ -150,48 +148,49 @@ func ListModulesForNMC(_ context.Context, obj client.Object) []reconcile.Request
...
+
+			if !reflect.DeepEqual(oldNode.Status.NodeInfo, newNode.Status.NodeInfo) {
+				return true
 			}
 
-			// clear the resource version as it gets updated with every heartbeat
-			oldNodeCopy.ResourceVersion = ""
-			newNodeCopy.ResourceVersion = ""
+			oldReady := getNodeReadyConditionStatus(oldNode)
+			newReady := getNodeReadyConditionStatus(newNode)
```
ybettan, line 172:
```suggestion
			oldReadyCondition := getNodeReadyConditionStatus(oldNode)
			newReadyCondition := getNodeReadyConditionStatus(newNode)
```

internal/filter/filter.go
```diff
@@ -150,48 +148,49 @@ func ListModulesForNMC(_ context.Context, obj client.Object) []reconcile.Request
 }
 
 func skipNodeHeartbeat() predicate.Predicate {
```
ybettan, line 150: We need to rename this function to something that reflects more of what it is doing.

internal/filter/filter.go
yevgeny-shnaidman: I think that change in the Ready status should be evaluated.

PR #1094, opened 2025-05-06
Updated the MCM controller to work with the new MIC implementation.

ybettan:
The MCM controller doesn't have to create build/sign pods manually anymore, all it needs to do is to create 1 MIC object in the cluster per MCM object per managed cluster.

If all images in an MIC are ready, then the controller can proceed with creating the manifestWork because it guarantees that the targeted managed cluster can pull all kmod images successfully.

---

/assign @yevgeny-shnaidman


yevgeny-shnaidman:
can we move all the changes to CRDs (config directory and Makefile to a separate PR)?
Do we need the changes to workflow?

internal/controllers/hub/managedclustermodule_reconciler.go
```diff
@@ -173,5 +177,69 @@ func (r *ManagedClusterModuleReconciler) SetupWithManager(mgr ctrl.Manager) erro
...
+	for _, kver := range kernelVersions {
+
+		kver = strings.TrimSuffix(kver, "+")
+		mld, err := rh.clusterAPI.GetModuleLoaderDataForKernel(mcm, kver)
+		if err != nil {
+			if !errors.Is(err, module.ErrNoMatchingKernelMapping) {
+				errs = append(errs, fmt.Errorf("failed to get MLD for kernel %s: %v", kver, err))
+			}
+			// error getting kernelVersion or kernelVersion is not targeted by the managedClusterModule
+			continue
```
yevgeny-shnaidman, line 219: We should exist in case we failed to get mld for one of the version, otherwise a MIC will be created and it might even sat that all images exist

ybettan: Fixed

internal/controllers/hub/managedclustermodule_reconciler.go
```diff
@@ -65,55 +76,62 @@ func NewManagedClusterModuleReconciler(
...
+		err = r.reconHelper.setMicAsDesired(ctx, mcm, cluster.Name, kernelVersions)
+		if err != nil {
+			logger.Info(utils.WarnString(
+				fmt.Sprintf("Failed to set MIC as desired: %v", err),
+			))
 			continue
 		}
 
-		kernelVersions, err := r.clusterAPI.KernelVersions(cluster)
+		allImagesReady, err := r.reconHelper.isMicReady(ctx, mcm.Name, cluster.Name)
```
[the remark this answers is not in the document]
ybettan: Changed to `areImagesReady`

PR #1097, opened 2025-05-08
Making KMM's configuration be persist during upgrade

TomerNewman:
Until now, whenever a user upgraded KMM, the ConfigMap containing the operator configuration was redeployed, causing any user changes to be lost.
This commit removes the ConfigMap from KMM's deploy process, meaning KMM will take hard-coded default configuration values and the user can apply manually the ConfigMap, meaning it will not be re-deployed during upgrades.

---

/assign @ybettan


ybettan:
@TomerNewman This PR requires updating the documentation at https://github.com/kubernetes-sigs/kernel-module-management/blob/main/docs/mkdocs/documentation/configure.md

config/manager-base/manager_config_patch.yaml
```diff
@@ -10,6 +10,7 @@ spec:
       - name: manager
         args:
         - "--config=controller_config.yaml"
+        - "--user-config-name=kmm-user-override-config"
```
ybettan, line 13: I think having 2 configMaps to configure KMM can be confusing for customers. Especaially for the one that already uses `controller_config` to configure KMM as mentioned in the docs.

How about we set all the default in the code (at the upper layer as possible) and just extract `controller_config` configMap from the bundle. The benefits are:
* User will still use the same configMap as before
* It will persist during upgrades
* Only small documentation changes are required
* We don't have 2 configMap to do the same (and only one of them is persistent which is even worst)

internal/config/config.go
```diff
@@ -42,6 +46,44 @@ type Config struct {
 	Worker                 Worker         `yaml:"worker"`
 }
 
+type UserConfig struct {
+	SELinuxType      *string `yaml:"seLinuxType"`
+	FirmwareHostPath *string `yaml:"firmwareHostPath,omitempty"`
```
ybettan, line 51: We also need the `gcDelay` option to be configurable.

cmd/manager-hub/main.go
```diff
@@ -69,9 +71,9 @@ func main() {
 	logConfig := textlogger.NewConfig()
 	logConfig.AddFlags(flag.CommandLine)
 
-	var configFile string
+	var userConfigName string
 
-	flag.StringVar(&configFile, "config", "", "The path to the configuration file.")
```
ybettan, line 74: I think we can keep the old names. A config file can be applied by users as well. No need to call it user's config file IMO. This will also narrow down this PR to the bare minimum required to solve the issue.

internal/config/config.go
```diff
@@ -42,20 +42,48 @@ type Config struct {
...
+}
 
-	cfg := Config{}
+func decodeStrictYAMLIntoConfig(yamlData []byte, config *Config) error {
+	decoder := yaml.NewDecoder(bytes.NewReader(yamlData))
+	decoder.KnownFields(true)
 
-	if err = yaml.NewDecoder(fd).Decode(&cfg); err != nil {
-		return nil, fmt.Errorf("could not decode configuration file: %v", err)
+	if err := decoder.Decode(config); err != nil {
```
yevgeny-shnaidman, line 58: what happens if the yamlData does not contain all the fields? will the values in the config be overriden empty values, or stay the same?

cmd/manager-hub/main.go
```diff
@@ -104,6 +104,18 @@ func main() {
 
 	client := mgr.GetClient()
 
+	ctx := ctrl.SetupSignalHandler()
+	reader := mgr.GetAPIReader()
```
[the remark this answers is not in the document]
TomerNewman: because the cache has not started yet, and the API reader talks directly with the API server

internal/config/config.go
TomerNewman: it will not be overriden, for example:
```yaml
apiVersion: v1
data:
  controller_config.yaml: |
    healthProbeBindAddress: :9999
    metrics:
      secureServing: true
    worker:
      firmwareHostPath: /test/test
kind: ConfigMap
metadata:
  name: kmm-operator-manager-config
  namespace: kmm-operator-system

```
this is the config:

```json
{
    "HealthProbeBindAddress": ":9999",
    "Job": {
        "GCDelay": 0
    },
    "LeaderElection": {
        "Enabled": true,
        "ResourceID": "kmm.sigs.x-k8s.io"
    },
    "Metrics": {
        "BindAddress": "0.0.0.0:8443",
        "EnableAuthnAuthz": true,
        "SecureServing": true
    },
    "WebhookPort": 9443,
    "Worker": {
        "RunAsUser": 0,
        "SELinuxType": "spc_t",
        "FirmwareHostPath": "/test/test"
    }
}
```

cmd/manager-hub/main.go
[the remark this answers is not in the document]
yevgeny-shnaidman: do we care if the cache is started? Will using the client produce an error, as opposed to using the reader?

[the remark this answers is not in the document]
ybettan: But the cache is created with the cluster isn't it? When customer installs KMM, the cluster (and cache) are already running.

cmd/manager-hub/main.go
```diff
@@ -104,6 +104,18 @@ func main() {
...
+	cm := &corev1.ConfigMap{}
```
ybettan, line 109:
```suggestion
	managerConfig := &corev1.ConfigMap{}
```

cmd/manager-hub/main.go
```diff
@@ -104,6 +104,18 @@ func main() {
...
+	if err = reader.Get(ctx, types.NamespacedName{Namespace: operatorNamespace, Name: userConfigName}, cm); err != nil {
+		setupLogger.Info("could not get user ConfigMap, using the default one", "namespace", operatorNamespace, "name", userConfigName)
```
ybettan, line 111:
```suggestion
		setupLogger.Info("no custom ConfigMap configuring the manager was found in the namespace, using the default configuration", "namespace", operatorNamespace, "name", userConfigName)
```

cmd/manager-hub/main.go
```diff
@@ -123,8 +135,6 @@ func main() {
 	ctrlLogger := setupLogger.WithValues("name", hub.ManagedClusterModuleReconcilerName)
 	ctrlLogger.Info("Adding controller")
 
-	operatorNamespace := cmd.GetEnvOrFatalError(constants.OperatorNamespaceEnvVar, setupLogger)
-	ctx := ctrl.SetupSignalHandler()
```
ybettan, line 127: What does this do? Why did we remove it?

cmd/manager/main.go
```diff
@@ -66,9 +69,9 @@ func main() {
 	logConfig := textlogger.NewConfig()
 	logConfig.AddFlags(flag.CommandLine)
 
-	var configFile string
+	var userConfigName string
```
ybettan, line 72: Same as https://github.com/kubernetes-sigs/kernel-module-management/pull/1097/files#r2100366505

cmd/manager/main.go
```diff
@@ -109,6 +110,18 @@ func main() {
 
 	client := mgr.GetClient()
 
+	ctx := ctrl.SetupSignalHandler()
+	reader := mgr.GetAPIReader()
+	cm := &corev1.ConfigMap{}
```
ybettan, line 115:
```suggestion
	managerConfig := &corev1.ConfigMap{}
```

cmd/webhook-server/main.go
```diff
@@ -33,14 +36,14 @@ func main() {
 	logConfig.AddFlags(flag.CommandLine)
 
 	var (
-		configFile                 string
 		enableModule               bool
 		enableManagedClusterModule bool
 		enableNamespaceDeletion    bool
 		enablePreflightValidation  bool
+		userConfigName             string
```
ybettan, line 43: Let's use the same name as in `manager` and `manager-hub`

cmd/webhook-server/main.go
```diff
@@ -33,14 +36,14 @@ func main() {
...
 	)
 
-	flag.StringVar(&configFile, "config", "", "The path to the configuration file.")
+	flag.StringVar(&userConfigName, "config", "", "Name of the ConfigMap containing user config.")
```
ybettan, line 46:
```suggestion
	flag.StringVar(&userConfigName, "config", "", "Name of the ConfigMap containing the configuration file.")
```

config/manager/kustomization.yaml
```diff
@@ -6,6 +6,7 @@ resources:
 
 patches:
 - path: manager_worker_image_patch.yaml
+- path: manager_config_patch.yaml
```
ybettan, line 9: What is the difference between this and what we were doing before?

internal/config/config.go
```diff
@@ -42,20 +42,48 @@ type Config struct {
...
+		return fmt.Errorf("error unmarshaling YAML: %w", err)
```
ybettan, line 59:
```suggestion
		return fmt.Errorf("error unmarshaling YAML: %v", err)
```

We had a discussion long ago about it. tl;dr, we don't wrap the error unless we unwrap it in the caller using `errors.Is()` or `errors.As()`.

config/manager-hub/kustomization.yaml
```diff
@@ -9,10 +9,5 @@ images:
...
 
-configMapGenerator:
-- files:
-  - controller_config.yaml
-  name: manager-config
-
-generatorOptions:
-  disableNameSuffixHash: true
+patches:
+- path: manager_hub_config_patch.yaml
```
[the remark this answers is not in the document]
TomerNewman: @ybettan  until now we had 1 base patch file that was being used for both manager and manager-hub.
now that we have different argument values (1 for the regular and 1 for the hub manager) I assumed it would be better to split the patch files rather than keeping the same file in the manager-base.

internal/config/config.go
```diff
@@ -42,20 +42,48 @@ type Config struct {
...
+		},
+		Metrics: Metrics{
+			EnableAuthnAuthz: true,
+			BindAddress:      "0.0.0.0:8443",
+			SecureServing:    true,
+		},
+		Worker: Worker{
+			RunAsUser:        &runAsUser,
+			SELinuxType:      "spc_t",
+			FirmwareHostPath: &firmwarePath,
```
ybettan, line 84:
```suggestion
			FirmwareHostPath: ptr.To("/lib/firmware"),
```
And remove the variable.

internal/config/config.go
```diff
@@ -42,20 +42,48 @@ type Config struct {
...
+			Enabled:    true,
+			ResourceID: "kmm.sigs.x-k8s.io",
+		},
+		Metrics: Metrics{
+			EnableAuthnAuthz: true,
+			BindAddress:      "0.0.0.0:8443",
+			SecureServing:    true,
+		},
+		Worker: Worker{
+			RunAsUser:        &runAsUser,
```
ybettan, line 82:
```suggestion
			RunAsUser:        ptr.To[int64](0),
```

and remove the variable.

cmd/manager/main.go
TomerNewman: Even though it is not a file?

cmd/manager-hub/main.go
ybettan: Update: Let's call it `userConfigMapName`?

internal/config/config.go
```diff
@@ -41,21 +47,79 @@ type Config struct {
 	WebhookPort            int            `yaml:"webhookPort"`
 	Worker                 Worker         `yaml:"worker"`
 }
+type ConfigGetter interface {
+	GetConfig(ctx context.Context, logger logr.Logger, userConfigMapName, userConfigMapNamespace string) (*Config, error)
+}
+
+//go:generate mockgen -source=config.go -package=config -destination=mock_config.go
```
ybettan, line 54: Let's have a single `generate` annotation before `ConfigGetter` interface definition with both interfaces in it: like we do at https://github.com/kubernetes-sigs/kernel-module-management/blob/925409699c648c0eb747b34bc7b2e06b425a1c23/internal/module/kernelmapper.go#L16

internal/config/config.go
```diff
@@ -41,21 +47,79 @@ type Config struct {
...
-func ParseFile(path string) (*Config, error) {
-	fd, err := os.Open(path)
+func (ch *configHelper) getClient() (client.Client, error) {
+	clnt, err := client.New(ctrl.GetConfigOrDie(), client.Options{})
 	if err != nil {
-		return nil, fmt.Errorf("could not open the configuration file: %v", err)
+		return nil, fmt.Errorf("failed to create new client %v", err)
 	}
-	defer fd.Close()
+	return clnt, err
```
ybettan, line 81:
```suggestion
	return clnt, nil
```

internal/config/config.go
```diff
@@ -41,21 +47,79 @@ type Config struct {
...
+}
+
+func (ch *configHelper) loadConfigFromCM(cm *corev1.ConfigMap, cfg *Config) error {
+	const configKey = "controller_config.yaml"
+	rawConfig, ok := cm.Data[configKey]
```
ybettan, line 86: What happens if `cm == nil`? Can we return an error in case `cm == nil` ?

internal/config/config.go
```diff
@@ -41,21 +47,79 @@ type Config struct {
...
+	if !ok {
+		return fmt.Errorf("key %q not found in ConfigMap %s/%s", configKey, cm.Namespace, cm.Name)
+	}
+	return decodeStrictYAMLIntoConfig([]byte(rawConfig), cfg)
+}
 
-	cfg := Config{}
+func decodeStrictYAMLIntoConfig(yamlData []byte, config *Config) error {
```
ybettan, line 93: Any reason this is not an helper interface method as well?

internal/config/config.go
```diff
@@ -76,3 +140,22 @@ func (c *Config) ManagerOptions() *manager.Options {
 		WebhookServer:          webhook.NewServer(webhook.Options{Port: c.WebhookPort}),
 	}
 }
+
+func (cg *configGetter) GetConfig(ctx context.Context, logger logr.Logger, userConfigMapName, userConfigMapNamespace string) (*Config, error) {
+	cfg := newDefaultConfig()
+	managerConfig := &corev1.ConfigMap{}
+	clnt, err := cg.configHelper.getClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to getClient %v", err)
```
ybettan, line 149:
```suggestion
		return nil, fmt.Errorf("failed to get client %v", err)
```

internal/config/config.go
TomerNewman: you are right

internal/config/config.go
```diff
@@ -41,21 +47,79 @@ type Config struct {
...
-		return nil, fmt.Errorf("could not decode configuration file: %v", err)
+	if err := decoder.Decode(config); err != nil {
+		return fmt.Errorf("error unmarshaling YAML: %v", err)
 	}
 
-	return &cfg, nil
+	return nil
+}
+
+func newDefaultConfig() *Config {
```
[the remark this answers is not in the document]
TomerNewman: @ybettan  The reason I did not put it there is because I dont this function to be mocked (its only returns an initialized object)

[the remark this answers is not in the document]
ybettan: Maybe we can still put it in the helper interface for simplicity, that way we know all helper methods are part of the helper interface. It is not crucial if you don't like it but a bit cleaner IMO.

internal/config/config.go
```diff
@@ -42,20 +49,103 @@ type Config struct {
...
 
-	cfg := Config{}
+type configHelperAPI interface {
+	getClient() (client.Client, error)
+	overrideConfigFromCM(cm *corev1.ConfigMap, cfg *Config) error
+	decodeStrictYAMLIntoConfig(yamlData []byte, config *Config) error
+	newDefaultConfig() *Config
+}
+type configGetter struct {
+	configHelper configHelperAPI
```
ybettan, line 83: We can add the logger here.

PR #1099, opened 2025-05-12
move GetRelevantBuild and GetRelevantSign from Combiner to kernelMapperHelperAPI

TomerNewman:
Refactor the `module.Combiner` interface by removing the `GetRelevantBuild` and `GetRelevantSign` methods.
These methods are now part of the `module.kernelMapperHelperAPI` interface.
As part of the change, renaming `Combiner` interface to `BuildArgOverrider`.

---


internal/build/helper.go
```diff
@@ -10,7 +10,6 @@ import (
 
 type Helper interface {
 	ApplyBuildArgOverrides(args []kmmv1beta1.BuildArg, overrides ...kmmv1beta1.BuildArg) []kmmv1beta1.BuildArg
-	GetRelevantBuild(moduleBuild *kmmv1beta1.Build, mappingBuild *kmmv1beta1.Build) *kmmv1beta1.Build
```
ybettan, line 13: No need to modify this package. `internal/build` and `internal/sign` are being replaced by `internal/buildsign`. Once we finish to tie all loose ends we will completely remove bot `internal/build` and `internal/sign` packages entirely.

Let's keep this PR focused on the code that is here to stay (`internal/buildsign`)

internal/module/kernelmapper.go
```diff
@@ -51,15 +51,17 @@ type kernelMapperHelperAPI interface {
 	findKernelMapping(mappings []kmmv1beta1.KernelMapping, kernelVersion string) (*kmmv1beta1.KernelMapping, error)
 	prepareModuleLoaderData(mapping *kmmv1beta1.KernelMapping, mod *kmmv1beta1.Module, kernelVersion string) (*api.ModuleLoaderData, error)
 	replaceTemplates(mld *api.ModuleLoaderData) error
+	GetRelevantBuild(moduleBuild *kmmv1beta1.Build, mappingBuild *kmmv1beta1.Build) *kmmv1beta1.Build
+	GetRelevantSign(moduleSign *kmmv1beta1.Sign, mappingSign *kmmv1beta1.Sign, kernel string) (*kmmv1beta1.Sign, error)
```
ybettan, line 55: I think we can unexport those methods. No other package is using them (excluding the build package which uses the other methods from the build package.

internal/module/kernelmapper.go
```diff
@@ -51,15 +51,17 @@ type kernelMapperHelperAPI interface {
...
 }
 
 type kernelMapperHelper struct {
-	buildSignCombiner Combiner
+	buildArgOverrider BuildArgOverrider
```
ybettan, line 59: This feels a bit redundant. Can't we just put the `ApplyBuildArgOverrides` inside the `kernelMapperHelper` interface instead?

internal/build/helper.go
TomerNewman: @ybettan take a look at [cmd/manager-hub/main.go:116](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/cmd/manager-hub/main.go#L116).

We use buildArgOverrider (previosly Combiner) as `Helper` so we have to remove GetRelevantBuild also from the Helper interface (btw no one uses Helper.GetRelevantBuild() currently)

internal/module/kernelmapper.go
TomerNewman: Do you mean `kernelMapperHelperAPI`? I thought the all point is to leave `ApplyBuildArgOverrides` in a new Interface `BuildArgOverrider`

internal/build/helper.go
ybettan: The code you pointed out will very soon be removed https://github.com/kubernetes-sigs/kernel-module-management/pull/1094 :)

internal/module/kernelmapper.go
ybettan: > Do you mean kernelMapperHelperAPI?

This is indeed what I meant.

> I thought the all point is to leave ApplyBuildArgOverrides in a new Interface BuildArgOverrider

The whole point was to move `GetRelevant*` methods to the package they belong. What is then left in the "BuildArgOverrider" interface is just left-overs.

But while taking another look, `BuildArgOverrider` is also used in the `buildsign` package which I missed before so let's leave it as you did.

ybettan:
/hold

This PR should be merged after https://github.com/kubernetes-sigs/kernel-module-management/pull/1094 - it will simplify this PR as well.

TomerNewman:
/unhold

#1094 is merged and this branch is rebased

PR #1101, opened 2025-05-12
Update kmod_image.md

reavessm:
Keep module location consistent.  This isn't the most critical thing because `/lib/modules` should be a symlink to `/usr/lib/modules`, but might as well be consistent with the rest of the docs.

### DIFF
--- a/docs/mkdocs/documentation/kmod_image.md
+++ b/docs/mkdocs/documentation/kmod_image.md
@@ -141,7 +141,7 @@ COPY --from=builder /usr/src/kernel-module-management/ci/kmm-kmod/kmm_ci_a.ko /o
 COPY --from=builder /usr/src/kernel-module-management/ci/kmm-kmod/kmm_ci_b.ko /opt/lib/modules/${KERNEL_FULL_VERSION}/
 
 # Create the symbolic link
-RUN ln -s /lib/modules/${KERNEL_FULL_VERSION} /opt/lib/modules/${KERNEL_FULL_VERSION}/host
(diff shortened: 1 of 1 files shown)


docs/mkdocs/documentation/kmod_image.md
```diff
@@ -141,7 +141,7 @@ COPY --from=builder /usr/src/kernel-module-management/ci/kmm-kmod/kmm_ci_a.ko /o
 COPY --from=builder /usr/src/kernel-module-management/ci/kmm-kmod/kmm_ci_b.ko /opt/lib/modules/${KERNEL_FULL_VERSION}/
 
 # Create the symbolic link
-RUN ln -s /lib/modules/${KERNEL_FULL_VERSION} /opt/lib/modules/${KERNEL_FULL_VERSION}/host
+RUN ln -s /usr/lib/modules/${KERNEL_FULL_VERSION} /opt/lib/modules/${KERNEL_FULL_VERSION}/host
```
ybettan, line 144: I suspect this may not work in diestros in which `/lib` isn't linked to `/usr/lib`.

AFAIK, `/lib/modules` will always work and if there is an internal linkage to something else then it still work.

@yevgeny-shnaidman WDYT about this?

yevgeny-shnaidman: since this is an example i also don't think this line should be changed.  /lib/module is the conventional place , in most of the systems it is linked to /usr/lib/modules, and in those that those directories are separate - then most of the drivers' files will be present in the /lb/modules

PR #1102, opened 2025-05-14
Gathering kmm RBAC annotations in Module controller

TomerNewman:
Until now, RBAC annotations under internal/controllers and internal/controllers/hub were scattered across multiple controller files.

With this commit, the annotations are now centralized only in the Module controller and the MCM controller.

This change addresses the issue where annotations spread across various controllers are applied only to KMM and not to KMM-hub—because all controllers except MCM reside in internal/controllers, and RBAC generation for KMM-hub only scans internal/controllers/hub.

---

/assign @ybettan 
/cc @yevgeny-shnaidman

### DIFF
--- a/internal/controllers/build_sign_reconciler.go
+++ b/internal/controllers/build_sign_reconciler.go
@@ -64,13 +64,6 @@ func NewBuildSignReconciler(
 	}
 }
 
-//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules,verbs=get;list;watch;update;patch
-//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules/status,verbs=get;update;patch
-//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;list;watch
-//+kubebuilder:rbac:groups="core",resources=secrets,verbs=get;list;watch
-//+kubebuilder:rbac:groups="core",resources=configmaps,verbs=get;list;watch
-//+kubebuilder:rbac:groups="core",resources=pods,verbs=create;list;watch;delete
-
 // Reconcile lists all nodes and looks for kernels that match its mappings.
 // For each mapping that matches at least one node in the cluster, it creates a DaemonSet running the container image
 // on the nodes with a compatible kernel.
--- a/internal/controllers/module_reconciler.go
+++ b/internal/controllers/module_reconciler.go
@@ -32,9 +32,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-//+kubebuilder:rbac:groups="core",resources=namespaces,verbs=get;list;patch;watch
-//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;watch
+//+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=delete;patch;update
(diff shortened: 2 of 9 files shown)


ybettan:
@TomerNewman Can you please extend the commit message to describe what is the issue? (That all the controller annotations will only generate RBAC for KMM and not for KMM-hub as all contronllers but MCM are in `internal/controllers`).

So next time, if someone wish to split it again, he can see why it was gathered to begin with.

internal/controllers/build_sign_reconciler.go
```diff
@@ -64,13 +64,6 @@ func NewBuildSignReconciler(
 	}
 }
 
-//+kubebuilder:rbac:groups=kmm.sigs.x-k8s.io,resources=modules,verbs=get;list;watch;update;patch
```
ybettan, line 67: All the changes needs to also be applied to `internal/controller/hub/managedclustermodule_recondiler.go`

internal/controllers/module_reconciler.go
```diff
@@ -32,9 +32,25 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-//+kubebuilder:rbac:groups="core",resources=namespaces,verbs=get;list;patch;watch
-//+kubebuilder:rbac:groups="core",resources=nodes,verbs=get;watch
+//+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=create;delete;get;list;patch;watch
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=delete;patch;update
+//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=create;get;list;watch
```
ybettan, line 37: Can we split those based on the resource? One line for `clusterclaims` and one for `kernel-versions.kmm.node.kubernetes.io`.

The `clusterclaim` line will need to aggregate all verbs from both lines.

internal/controllers/build_sign_reconciler.go
ybettan: I take it back. Everything needed for KMM-hub is already present in the MCM controller...

internal/controllers/module_reconciler.go
ybettan: No. 

First line says:
* give `delete;patch;update` capabilities to `clusterclaims` and to `kernel-versions.kmm.node.kubernetes.io`

Second line says:
* give `create;get;list;watch` capabilities to `clusterclaims`.

My suggestion was to have one line for each resource:
* One for `clusterclaims` with the `delete;patch;update;create;get;list;watch` (combined from the 2 lines)
* One for `kernel-versions.kmm.node.kubernetes.io` with the `delete;patch;update` capabilities.

So it basically looks like:
```
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=delete;patch;update
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=create;get;list;watch;delete;patch;update
```

TomerNewman: but
```
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,resourceNames=kernel-versions.kmm.node.kubernetes.io,verbs=delete;patch;update
//+kubebuilder:rbac:groups=cluster.open-cluster-management.io,resources=clusterclaims,verbs=create;get;list;watch
```
means that `ClusterClaims` can be `create;get;list;watch` unless the `ClusterClaim` name is `kernel-versions.kmm.node.kubernetes.io`, then it can also delete patch and update

ybettan: Ohh sorry. I missed it says `resourceName` and not just `resource`. My bad.

PR #1104, opened 2025-05-15
Removing the old build/sign packages and the old build-sign controller.

ybettan:
The `build_sign_reconciler` has been replaced with the MBSC controller and the `build`/`sign` packages were replaced by the `buildsign` package.

---

/assign @yevgeny-shnaidman


ybettan:
/hold

Instead of fixing the RBAC, I will rebase on top of https://github.com/kubernetes-sigs/kernel-module-management/pull/1102#issuecomment-2883166988 once merged.

PR #1109, opened 2025-05-18
Deleting module.ImageExist and module.ImageDigest

TomerNewman:
Since we now check if images exists using the new MIC object, those functions are now redundant.
This commit deletes them and their dependencies (registry and auth packages).

---

/cc @yevgeny-shnaidman 
/assign @ybettan


ybettan:
/ok-to-test

@TomerNewman I think we can also remove `ImageDigest` in the same file. It is not used anymore either.

ybettan:
I think we should also remove all the code that isn't being used anymore after removing those functions. The commit message should probably be updated accordingly.

ybettan:
/approve
@yevgeny-shnaidman PTAL.

PR #1111, opened 2025-05-18
Change Preflight VerifactionStatus and VerifactionStage definitions

yevgeny-shnaidman:
Since preflight is moving to MIC/MBSC implementation, the preflight implementation is also changing. MIC is taking over all the handling of build/sign and image verification, so we only need one stage: Image, and the status is now becoming InProgress/Success/Failure

### DIFF
--- a/api/v1beta1/preflightvalidation_types.go
+++ b/api/v1beta1/preflightvalidation_types.go
@@ -25,13 +25,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


ybettan:
After some discussions, although this may be a breaking change, we didn't get any feedback from users using it and in addition to that, if an admin is running preflight, he will most probably do it manually, therefore the risk of breaking is pretty small.

/lgtm

PR #1112, opened 2025-05-19
Adding a preflight package

yevgeny-shnaidman:
This package is going to be used by the Preflight controller to perform manipulation on the in-memory PreflightValidation object. The main functionalities: setting the status of the veirfied module, checking the status of veirfied module, and checking if all the modules in the cluster has already been verified


internal/preflight/preflight.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"github.com/kubernetes-sigs/kernel-module-management/api/v1beta2"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const VerificationStatusReasonVerified = "Verification successful (%s), this Module will not be verified again in this Preflight CR"
+
+//go:generate mockgen -source=preflight.go -package=preflight -destination=mock_preflight_api.go
```
ybettan, line 11:
```suggestion
//go:generate mockgen -source=preflight.go -package=preflight -destination=mock_preflight.go
```

internal/preflight/preflight.go
```diff
@@ -0,0 +1,67 @@
...
+	pv.Status.Modules = append(pv.Status.Modules, newStatus)
+}
+
+func (p *preflight) GetModuleStatus(pv *v1beta2.PreflightValidation, namespace, name string) string {
+	for _, moduleStatus := range pv.Status.Modules {
+		if moduleStatus.Name == name && moduleStatus.Namespace == namespace {
+			return moduleStatus.VerificationStatus
+		}
+	}
+	return ""
```
ybettan, line 57: Should we return an error here instead of an empty value? errors are more often checked than empty values.

internal/preflight/preflight.go
yevgeny-shnaidman: Fixed

internal/preflight/preflight.go
yevgeny-shnaidman: this is not an error, the module is not present, since it might be that we have not yet pick it up. The empty status is ok in this case

PR #1116, opened 2025-05-27
Adding init container for device plugin implementation

yevgeny-shnaidman:
This commit creates an init container definition for the device-plugin daemonset, based on the definitions in the Module CR. In case the ini-container is not defined in the Module CR, then it won't be defined for the device-plugin daemonset

### DIFF
--- a/internal/controllers/device_plugin_reconciler.go
+++ b/internal/controllers/device_plugin_reconciler.go
@@ -364,19 +364,8 @@ func (dsci *daemonSetCreatorImpl) setDevicePluginAsDesired(
 				Finalizers: []string{constants.NodeLabelerFinalizer},
 			},
 			Spec: v1.PodSpec{
-				Containers: []v1.Container{
-					{
-						Args:            mod.Spec.DevicePlugin.Container.Args,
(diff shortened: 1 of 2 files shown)


internal/controllers/device_plugin_reconciler.go
```diff
@@ -364,19 +364,37 @@ func (dsci *daemonSetCreatorImpl) setDevicePluginAsDesired(
...
+							Env:             mod.Spec.DevicePlugin.Container.Env,
+							Name:            "device-plugin",
+							Image:           mod.Spec.DevicePlugin.Container.Image,
+							ImagePullPolicy: mod.Spec.DevicePlugin.Container.ImagePullPolicy,
+							Resources:       mod.Spec.DevicePlugin.Container.Resources,
+							SecurityContext: &v1.SecurityContext{Privileged: ptr.To(true)},
+							VolumeMounts:    append(mod.Spec.DevicePlugin.Container.VolumeMounts, containerVolumeMounts...),
+						},
+					},
+				*/
```
ybettan, line 397: Remove?

PR #1129, opened 2025-06-08
removing duplicate imageSpec from MIC

yevgeny-shnaidman:
CreateOrPatch MIC API receives slice of imageSpecs. Since usually the input is created by going over the the target nodes and adding their images to the slice, we need to remove duplicate images, since for most of the cases all the nodes will have the same target image

### DIFF
--- a/internal/mic/mic.go
+++ b/internal/mic/mic.go
@@ -9,6 +9,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -49,6 +50,8 @@ func (mici *micImpl) CreateOrPatch(ctx context.Context, name, ns string, images 
 			Namespace: ns,
 		},
 	}
+
+	images = filterDuplicateImages(images)
 
 	opRes, err := controllerutil.CreateOrPatch(ctx, mici.client, mic, func() error {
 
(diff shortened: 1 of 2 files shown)


internal/mic/mic.go
```diff
@@ -127,3 +130,16 @@ func (mici *micImpl) DoAllImagesExist(micObj *kmmv1beta1.ModuleImagesConfig) boo
...
+func removeDuplicateImages(images []kmmv1beta1.ModuleImageSpec) []kmmv1beta1.ModuleImageSpec {
+	imagesSet := sets.New[string]()
+	for i := 0; i < len(images); {
+		if !imagesSet.Has(images[i].Image) {
+			imagesSet.Insert(images[i].Image)
+			i++
+		} else {
+			images = append(images[:i], images[i+1:]...)
+		}
+	}
```
ybettan:
```suggestion
	for _, imgSpec := range images {
		if !imagesSet.Has(imgSpec.Image) {
			imagesSet.Insert(imgSpec.Image)
		}
	}
```
WDYT?

**Edit: If you agree to the change suggested later in this PR this whole method can be removed.

internal/mic/mic.go
```diff
@@ -127,3 +130,16 @@ func (mici *micImpl) DoAllImagesExist(micObj *kmmv1beta1.ModuleImagesConfig) boo
...
+	return images
```
ybettan:
```suggestion
	return sets.List(imageSet)
```

internal/mic/mic.go
```diff
@@ -50,6 +51,8 @@ func (mici *micImpl) CreateOrPatch(ctx context.Context, name, ns string, images
 		},
 	}
 
+	images = removeDuplicateImages(images)
```
ybettan: Can we use a Set directly when [creating the imageList](https://github.com/kubernetes-sigs/kernel-module-management/blob/189d136a41e17852ab54f197a9487ee46a9fc591/internal/controllers/module_reconciler.go#L376) instead of filtering it later on?

We can then send a filtered list using `set.List()` directly to `mic.CreateOrPatch`

yevgeny-shnaidman: you cannot use set.List directly, because the comparison function for the struct is all the fields , and not a specific field. So you will basically need to implement the function that i created at every place that calls CreateOrPatch (module/preflight/managedcluster controllers). I made the change in MIC because this is a single point. If we go the way of changing all the places that are calling CreateOrPatch, then i would prefer to change MIC API  and make it receive only input parameters and create the ImageSpec by itself

PR #1130, opened 2025-06-08
[WIP] Ending the `Module` reconciliation loop if the MIC was just created.

ybettan:
The module reconciler will try to add entries to the NMC object for kmods
which their image exist and have a valid node to be loaded to.

If we continue the reconciliation loop directly after creating the MIC
we may encounter a race condition between k8s actually creating the MIC
object and KMM trying to access it to check if an image exist.
Although this is automatically fixed in the next loop iteration, this will
result in some errors in the operator logs which can lead a debugger in
the wrong direction when checking the logs.

By ending the loop if the MIC was just created we can ensure that, in the
next iteration, the MIC is already created and as soon as the MIC status
is updated the module reconciler will get notified.

---

/assign @yevgeny-shnaidman 
This is WIP because I still need to make sure it solves the issue and fix the UT but this is the idea of the change.

### DIFF
--- a/internal/controllers/hub/managedclustermodule_reconciler.go
+++ b/internal/controllers/hub/managedclustermodule_reconciler.go
@@ -115,10 +115,12 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, mcm *hub
 			continue
 		}
 
-		err = r.reconHelper.setMicAsDesired(ctx, mcm, cluster.Name, kernelVersions)
-		if err != nil {
+		if micOpRes, err := r.reconHelper.setMicAsDesired(ctx, mcm, cluster.Name, kernelVersions); err != nil {
 			logger.Info(utils.WarnString(fmt.Sprintf("Failed to set MIC as desired: %v", err)))
 			continue
+		} else if micOpRes == controllerutil.OperationResultCreated {
+			logger.Info("MIC was just created, waiting for its status to get updated")
+			return ctrl.Result{}, nil
 		}
 
 		allImagesReady, err := r.reconHelper.areImagesReady(ctx, mcm.Name, cluster.Name)
@@ -187,7 +189,8 @@ func (r *ManagedClusterModuleReconciler) SetupWithManager(mgr ctrl.Manager) erro
 //go:generate mockgen -source=managedclustermodule_reconciler.go -package=hub -destination=mock_managedclustermodule_reconciler.go managedClusterModuleReconcilerHelperAPI
 
 type managedClusterModuleReconcilerHelperAPI interface {
-	setMicAsDesired(ctx context.Context, mcm *hubv1beta1.ManagedClusterModule, clusterName string, kernelVersions []string) error
+	setMicAsDesired(ctx context.Context, mcm *hubv1beta1.ManagedClusterModule, clusterName string,
+		kernelVersions []string) (controllerutil.OperationResult, error)
 	areImagesReady(ctx context.Context, mcmName, clusterName string) (bool, error)
 }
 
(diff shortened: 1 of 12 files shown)


internal/controllers/preflightvalidation_reconciler.go
```diff
@@ -100,14 +101,16 @@ func (r *preflightValidationReconciler) Reconcile(ctx context.Context, pv *v1bet
 		return ctrl.Result{}, fmt.Errorf("failed to get modules' data: %v", err)
 	}
 
-	err = r.helper.updateStatus(ctx, modsWithMapping, modsWithoutMapping, pv)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to update PreflightValidation's status: %v", err)
+	if micOpRes, err := r.helper.processPreflightValidation(ctx, modsWithMapping, pv); err != nil {
```
yevgeny-shnaidman, line 104: Why do we need to switch places between updateStatus  and prcoessPreflightValidation?

internal/controllers/preflightvalidation_reconciler.go
```diff
@@ -230,10 +239,12 @@ func (p *preflightReconcilerHelperImpl) processPreflightValidation(ctx context.C
 			RegistryTLS:   mod.RegistryTLS,
 		}
 		micName := mod.Name + "-preflight"
-		err := p.micAPI.CreateOrPatch(ctx, micName, mod.Namespace, []kmmv1beta1.ModuleImageSpec{micObjSpec}, mod.ImageRepoSecret, mod.ImagePullPolicy, pv)
+		// We only care about the last opRes as we want to make sure we give enough time for the MIC to be created
+		opRes, err = p.micAPI.CreateOrPatch(ctx, micName, mod.Namespace, []kmmv1beta1.ModuleImageSpec{micObjSpec},
```
yevgeny-shnaidman, line 243: what if the first opRes is "Created" and the last is "Patch"?

internal/controllers/preflightvalidation_reconciler.go
ybettan: Because we need to validate the CreateOrPatch result before to understand if checking the status should be skipped or not.

internal/controllers/preflightvalidation_reconciler.go
ybettan: Then we return "Patch" and ignore it in the caller. By this time, we know for sure that it was already created cause we had time to patch it.

internal/controllers/preflightvalidation_reconciler.go
yevgeny-shnaidman: checking status knows how to handle missing MIC, so that is not supposed to be an issue. It also set the initial statuses of the existing modules in the preflight, so it makes sense for it ot be in first place

internal/controllers/preflightvalidation_reconciler.go
yevgeny-shnaidman: but we create multiple MICs, one per module, so if the first MIC returns Created, but the last returns Patched, you  are missing the whole point of checking the operation status, no?

internal/controllers/hub/managedclustermodule_reconciler.go
```diff
@@ -115,10 +115,12 @@ func (r *ManagedClusterModuleReconciler) Reconcile(ctx context.Context, mcm *hub
 			continue
 		}
 
-		err = r.reconHelper.setMicAsDesired(ctx, mcm, cluster.Name, kernelVersions)
-		if err != nil {
+		if micOpRes, err := r.reconHelper.setMicAsDesired(ctx, mcm, cluster.Name, kernelVersions); err != nil {
 			logger.Info(utils.WarnString(fmt.Sprintf("Failed to set MIC as desired: %v", err)))
 			continue
+		} else if micOpRes == controllerutil.OperationResultCreated {
```
TomerNewman, line 121: question, shouldn't we also address when mic is patched?

PR #1136, opened 2025-06-17
Adding PreflightValidation Webhook

TomerNewman:
Until now users were able to add invalid kernel versions to PreflightValidation, which made KMM controller panic. 
This commit adds a validation webhook for PreflightValidation resource.

---

### DIFF
--- a/cmd/webhook-server/main.go
+++ b/cmd/webhook-server/main.go
@@ -109,8 +109,9 @@ func main() {
 			cmd.FatalError(setupLogger, err, "unable to create conversion webhook", "name", "PreflightValidation/v1beta1")
 		}
 
-		if err = ctrl.NewWebhookManagedBy(mgr).For(&kmmv1beta2.PreflightValidation{}).Complete(); err != nil {
-			cmd.FatalError(setupLogger, err, "unable to create conversion webhook", "name", "PreflightValidation/v1beta2")
+		logger.Info("Enabling PreflightValidation webhook")
+		if err = webhook.NewPreflightValidationValidator(logger).SetupWebhookWithManager(mgr, &kmmv1beta2.PreflightValidation{}); err != nil {
+			cmd.FatalError(setupLogger, err, "unable to create webhook", "webhook", "PreflightValidationValidator")
 		}
 	}
 
--- a/internal/utils/replaceStrings.go
+++ b/internal/utils/replaceStrings.go
@@ -14,10 +14,13 @@ const (
 	kernelVersionPatchIdx = 2
 )
 
-var kernelRegexp = regexp.MustCompile("[.,-]")
+var KernelRegexp = regexp.MustCompile("[.,-]")
 
-func KernelComponentsAsEnvVars(kernel string) []string {
-	osConfigFieldsList := kernelRegexp.Split(kernel, -1)
+func KernelComponentsAsEnvVars(kernel string) ([]string, error) {
+	osConfigFieldsList := KernelRegexp.Split(kernel, -1)
+	if len(osConfigFieldsList) < 3 {
+		return nil, fmt.Errorf("invalid kernel version %s: expected at least three components (major.minor.patch)", kernel)
+	}
 
 	envvars := []string{
 		"KERNEL_FULL_VERSION=" + kernel,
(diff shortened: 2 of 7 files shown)


internal/utils/replaceStrings.go
```diff
@@ -16,8 +16,11 @@ const (
 
 var kernelRegexp = regexp.MustCompile("[.,-]")
 
-func KernelComponentsAsEnvVars(kernel string) []string {
+func KernelComponentsAsEnvVars(kernel string) ([]string, error) {
 	osConfigFieldsList := kernelRegexp.Split(kernel, -1)
+	if len(osConfigFieldsList) < 3 {
+		return nil, fmt.Errorf("invalid kernel version")
```
ybettan: Can we return a more detailed error message? For example, "expecting the %s format but got %s" or something similar.

cmd/webhook-server/main.go
```diff
@@ -112,6 +112,11 @@ func main() {
 		if err = ctrl.NewWebhookManagedBy(mgr).For(&kmmv1beta2.PreflightValidation{}).Complete(); err != nil {
 			cmd.FatalError(setupLogger, err, "unable to create conversion webhook", "name", "PreflightValidation/v1beta2")
 		}
+
+		logger.Info("Enabling PreflightValidation webhook")
+		if err = webhook.NewPreflightValidationValidator(logger).SetupWebhookWithManager(mgr); err != nil {
+			cmd.FatalError(setupLogger, err, "unable to create webhook", "webhook", "PreflightValidationValidator")
+		}
```
ybettan: Why do we need [that](https://github.com/kubernetes-sigs/kernel-module-management/pull/1136/files#diff-82f99234420e5315633bec59f91c966d1ffcdccc1728a57328b8dc4e59c7ff52R112-R114) if we are adding this section?

Also, we need to add the webhook for both preflightv2 and preflightv1 as both of them are still supported in the current KMM version.

TomerNewman: We use conversion webhook between the two version so this is why I can not delete it.
Because of that it can "block" preflightv1 as well so I am not sure how much value will it add.
```yaml
apiVersion: kmm.sigs.x-k8s.io/v1beta1
kind: PreflightValidation
metadata:
  name: preflight
  namespace: preflight
spec:
  pushBuiltImage: false
  kernelVersion: tesdt
```

```
oc apply -f pf.yaml 
-> Warning: kmm.sigs.x-k8s.io/v1beta1 PreflightValidation is deprecated; use kmm.sigs.x-k8s.io/v1beta2 PreflightValidation
Error from server (Forbidden): error when creating "pf.yaml": admission webhook "vpreflightvalidation.kb.io" denied the request: invalid kernelVersion tesdt


oc get preflightvalidations.kmm.sigs.x-k8s.io -A
-> No resources found
```

cmd/webhook-server/main.go
```diff
@@ -109,8 +109,9 @@ func main() {
 			cmd.FatalError(setupLogger, err, "unable to create conversion webhook", "name", "PreflightValidation/v1beta1")
 		}
 
-		if err = ctrl.NewWebhookManagedBy(mgr).For(&kmmv1beta2.PreflightValidation{}).Complete(); err != nil {
-			cmd.FatalError(setupLogger, err, "unable to create conversion webhook", "name", "PreflightValidation/v1beta2")
+		logger.Info("Enabling PreflightValidation webhook")
+		if err = webhook.NewPreflightValidationValidator(logger).SetupWebhookWithManager(mgr); err != nil {
```
ybettan:
```suggestion
		if err = webhook.NewPreflightValidationValidator(logger).SetupWebhookWithManager(mgr, &kmmv1beta2.PreflightValidation{}); err != nil {
```
I would explicitly add the v2 preflight as input so it clear at first sight how v1 is created and how v2 is created in the same place.

internal/webhook/preflight.go
```diff
@@ -0,0 +1,75 @@
...
+	"regexp"
+
+	"github.com/go-logr/logr"
+	kmmv1beta2 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta2"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+var kernelVersionRegexp = regexp.MustCompile(`[.,-]`)
```
ybettan: Can we export and use [this](https://github.com/kubernetes-sigs/kernel-module-management/blob/51d554e947eed86bb86e97915a34f1cd98fb8447/internal/utils/replaceStrings.go#L17) instead of duplicating the definition?

PR #1137, opened 2025-06-19
adding documentation for kmods events that are not shown

TomerNewman:
Kubernetes includes an anti-spam mechanism for events that are created rapidly.
This commit adds documentation about it,
as this feature can cause kernel module events to not be shown.

---


docs/mkdocs/documentation/deploy_kmod.md
```diff
@@ -270,6 +270,10 @@ KMM will then create worker Pods where required to run `modprobe -r` and unload
     KMM ships with a validating admission webhook that rejects the deletion of namespaces that contain at least one
     `Module` resource.
 
+### Kernel modules events on Nodes
```
yevgeny-shnaidman, line 273: lets change it to "some events may not...", otherwise it reads like all the events will be missing

PR #1144, opened 2025-06-22
Adding PushBuiltImage variable to MIC and MBSC objects

yevgeny-shnaidman:
This flag will be used to signal whether the built object must be also push to the destination registry. The main user will be the Preflight. In the regular flow (Module) the image that is built is also always pushed. When running Preflight, user might specify whether to push the built image, or not (he wants to verify that image can be built in time, but does not want to push it right now)


api/v1beta1/moduleimagesconfig_types.go
```diff
@@ -72,6 +72,11 @@ type ModuleImagesConfigSpec struct {
 	// ImagePullPolicy defines the pull policy used for verifying the presence of the image
 	//+optional
 	ImagePullPolicy v1.PullPolicy `json:"imagePullPolicy"`
+
+	// Boolean flag that determines whether images built must also
+	// be pushed to a defined repository
+	// +optional
+	PushBuiltImage bool `json:"pushBuiltImage"`
```
ybettan, line 79: Let's put it in `images[]`? Alongside with `registryTLS` and `skipWaitMissingImage`.

yevgeny-shnaidman: but it is relevant for all images in the Module, not just for a specific image

ybettan: But we are building each image individually, and each one of them can be pushed or not. The same way that we allow different TLS conf for each image separately and decide for each image if we skip waiting or not.

The reason `imageRepoSecret` is global because it is like a single `dockerconfigjson` file on a Linux FS that contains tokens for all registries.

By making `pushBuiltImage` global, we restrict the user while other similar configuration are allowed per image.

ybettan: Since the `Preflight` CRD doesn't have this kind of granularity, we will make this `pushBuildImage` field global.

PR #1145, opened 2025-06-22
Propagate PushBuiltImage flag from different controllers to Build object

yevgeny-shnaidman:
Propagating the PushBuilt image to the MIC object and eventually to the MBSC object. Module/MCM reconcilers are setting it to true, Preflight is setting ti based on value of the PushBuiltImage in Preflight Spec

### DIFF
--- a/internal/mic/mic.go
+++ b/internal/mic/mic.go
@@ -19,7 +19,7 @@ import (
 
 type MIC interface {
 	CreateOrPatch(ctx context.Context, name, ns string, images []kmmv1beta1.ModuleImageSpec,
-		imageRepoSecret *v1.LocalObjectReference, pullPolicy v1.PullPolicy, owner metav1.Object) error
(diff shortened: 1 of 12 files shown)


internal/mic/mic.go
```diff
@@ -19,7 +19,7 @@ import (
 
 type MIC interface {
 	CreateOrPatch(ctx context.Context, name, ns string, images []kmmv1beta1.ModuleImageSpec,
-		imageRepoSecret *v1.LocalObjectReference, pullPolicy v1.PullPolicy, owner metav1.Object) error
+		imageRepoSecret *v1.LocalObjectReference, pullPolicy v1.PullPolicy, pushBuiltFlag bool, owner metav1.Object) error
```
ybettan:
```suggestion
		imageRepoSecret *v1.LocalObjectReference, pullPolicy v1.PullPolicy, pushBuiltImage bool, owner metav1.Object) error
```

nit;
Let's rename to be consistent? It is easier to track with `git grep` when it has the same name everywhere.

yevgeny-shnaidman: Fxied

PR #1148, opened 2025-06-26
fixing HashAnnotationDiffer function potential bug

TomerNewman:
The NMC logic ensure we call this function in such a way we never fall into one of the 2 buggy `if` sections.
That said, if the implementation were to change, we could hit this bug in real time, therefore, this method should be fixed.

### DIFF
--- a/internal/pod/workerpodmanager.go
+++ b/internal/pod/workerpodmanager.go
@@ -304,11 +304,11 @@ func (wpmi *workerPodManagerImpl) GetTolerationsAnnotation(p *v1.Pod) string {
 func (wpmi *workerPodManagerImpl) HashAnnotationDiffer(p1, p2 *v1.Pod) bool {
 
 	if p1 == nil && p2 == nil {
-		return true
+		return false
 	}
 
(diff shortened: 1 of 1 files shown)


internal/pod/workerpodmanager.go
```diff
@@ -304,11 +304,11 @@ func (wpmi *workerPodManagerImpl) GetTolerationsAnnotation(p *v1.Pod) string {
 func (wpmi *workerPodManagerImpl) HashAnnotationDiffer(p1, p2 *v1.Pod) bool {
 
 	if p1 == nil && p2 == nil {
-		return true
+		return false
 	}
 
-	if (p1 == nil && p2 != nil) || (p1 != nil) && (p2 == nil) {
-		return false
+	if p1 == nil || p2 == nil {
```
ybettan: I would keep the old `if` condition. It makes it more clear that only one of them is not `nil` without the need to look at the first condition to extract additional information. This will also not mess up the function if the first condition is removed for some reason in the future.

PR #1152, opened 2025-07-02
Setting the cluster as a hub cluster.

ybettan:
We need this step in order to install the `manifestwork` and `clusterclaim` CRDs in the cluster.

Without it, `operator-sdk` will still run and manage to do the upgrade but the installed operator will constantly fail due to missing CRDs in the cluster.

This is not affecting the job correctness which is only supposed to test the upgrade but it will make thing much easier if it fails and we need to inspect the "real failing logs" and not the one generated because of missing CRDs.

---

/assign @yevgeny-shnaidman


ybettan:
/hold

The job isn't supposed to fail.

PR #1154, opened 2025-07-10
Introducing DockerfileOCIArtifact field to Module CRD.

TomerNewman:
As part of transitioning our builds to Shipwright using OCIArtifacts instead of pods running the Kaniko container, we need to allow users to specify the registry and artifact that contain the Dockerfile used for building the kernel module image.
The `dockerfileOCIArtifact` field will remain optional for now, until the migration to Shipwright is complete.

---

### DIFF
--- a/api/v1beta1/module_types.go
+++ b/api/v1beta1/module_types.go
@@ -53,6 +53,11 @@ type Build struct {
 	DockerfileConfigMap *v1.LocalObjectReference `json:"dockerfileConfigMap"`
 
 	// +optional
+	// DockerfileOCIArtifact specifies an OCI artifact that contains the dockerfile
+	// used to build the kernel module container image.
+	DockerfileOCIArtifact string `json:"dockerfileOCIArtifact"`
+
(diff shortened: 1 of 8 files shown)


api/v1beta1/module_types.go
```diff
@@ -52,6 +52,10 @@ type Build struct {
 	// ConfigMap that holds Dockerfile contents
 	DockerfileConfigMap *v1.LocalObjectReference `json:"dockerfileConfigMap"`
 
+	// +optional
+	// Image that holds Dockerfile contents
+	DockerfileImage string `json:"dockerfileImage"`
```
ybettan: How about we call it `DockerfileOCIArtifcat`? `DockerfileImage` can be confused with the context of a "container image" or "docker image".

We also need to add a validation webhook to ensure that `DockerfileConfigMap` and the new field are mutually exclusive.

TomerNewman: yes I agree I will change it

PR #1155, opened 2025-07-13
Adjusting webhook for shipwright builds

TomerNewman:
As part of KMM transitioning to Shipwright, KMM's webhook should allow only the following scenarios:
1. DockerfileOCIArtifact is defined and not DockerfileConfigMap.
2. DockerfileConfigMap is defined and not DockerfileOCIArtifact.

---

/assign @yevgeny-shnaidman 
---

/hold
lets wait for #1154


internal/webhook/module.go
```diff
@@ -175,6 +175,26 @@ func validateModuleLoaderContainerSpec(container kmmv1beta1.ModuleLoaderContaine
 		}
 	}
 
+	return validateModuleLoaderContainerBuildSpec(container.Build)
```
yevgeny-shnaidman, line 182: We have Build definition both global (in ModuleLoaderContainerSpec) and in KernelMapping.  Currently you are verifying only the global one, and you are missing verification for Build definition in the KernelMappings

TomerNewman: you are right, I've added validation for each Kernel mapping's build

PR #1177, opened 2025-08-28
[WIP] Changing build and sign to use Buildah instead of Kaniko container

TomerNewman:
Because Kaniko is archived and thus not supported, it would be wise to change the build and sign feature to use Buildah container instead of Kaniko.
That way we can also match the build and sign d/s to behave the same.

---

### DIFF
--- a/config/manager-base/manager.yaml
+++ b/config/manager-base/manager.yaml
@@ -48,7 +48,7 @@ spec:
               fieldRef:
                 fieldPath: metadata.namespace
           - name: RELATED_IMAGE_BUILD
-            value: gcr.io/kaniko-project/executor:latest
+            value: quay.io/buildah/stable:latest
           - name: RELATED_IMAGE_SIGN
             value: signer
         securityContext:
--- a/internal/buildsign/resource/common.go
+++ b/internal/buildsign/resource/common.go
@@ -6,8 +6,9 @@ import (
 	"embed"
 	"fmt"
 	"os"
-	"strings"
 	"text/template"
+
+	"k8s.io/utils/ptr"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/api"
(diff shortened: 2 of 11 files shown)


config/manager-base/manager.yaml
```diff
@@ -48,7 +48,7 @@ spec:
               fieldRef:
                 fieldPath: metadata.namespace
           - name: RELATED_IMAGE_BUILD
-            value: gcr.io/kaniko-project/executor:latest
+            value: quay.io/buildah/stable:latest
```
ybettan, line 51: Is this the official image that is being used by the u/s Shipwright project?
(I don't remember if when we tested it it was with the u/s version of Shipwright)

internal/buildsign/resource/common.go
```diff
@@ -34,11 +36,18 @@ var tmpl = template.Must(
 	template.ParseFS(templateFS, "templates/Dockerfile.gotmpl"),
 )
 
+func formatBuildArgs(buildArgs []kmmv1beta1.BuildArg) string {
```
ybettan: Should we put this method in the `BuildArgOverrider` interface?

internal/buildsign/resource/common.go
```diff
@@ -108,33 +114,102 @@ func signSpec(mld *api.ModuleLoaderData, destinationImg string, pushImage bool)
...
 
-func containerArgs(mld *api.ModuleLoaderData, destinationImg string,
-	tlsOptions kmmv1beta1.TLSOptions, pushImage bool) []string {
+type BuildOperation int
 
-	args := []string{}
+const (
+	Build BuildOperation = iota
+	Sign
+)
```
ybettan: Can we use the API's `BuildOrSignAction` instead of defining a new "type"?

internal/buildsign/resource/common.go
```diff
@@ -108,33 +114,102 @@ func signSpec(mld *api.ModuleLoaderData, destinationImg string, pushImage bool)
...
+	}
+
+	actionDescription := "build"
+	pushDescription := "image"
+	if operation == Sign {
+		actionDescription = "build for signing"
+		pushDescription = "signed image"
+	}
+
+	script := fmt.Sprintf(`
```
ybettan: Any reason we are not using [go templates](https://pkg.go.dev/text/template)? Like we did until now for signing?

config/manager-base/manager.yaml
TomerNewman: they use in their samples `quay.io/containers/buildah:v1.40.1`
and bump it each time.
https://github.com/shipwright-io/build/commit/10936eee4f54b78eefebfd9439d99f5f31889dec

TomerNewman: in their official docs (buildah)
```container images will be available at https://quay.io/repository/buildah/stable and https://quay.io/repository/containers/buildah.```

ybettan: Ack. Then let's use the official one as you did. Is there a way to use a specific version instead of `latest`? I think we should use `v1.41.3` instead of `latest` - we don't want things to start breaking under our legs.

TomerNewman:
Closing the PR.
KMM was recently started to use CRIO to pull images, instead of pulling from the code, switching to Buildah would require us to revert to the old approach and handle manually image mirroring and certifications.

PR #1193, opened 2025-11-13
Adding "version.ready" label to the ordered upgrade flow

yevgeny-shnaidman:
Currently we have only one "ready" label to signify that the kernel module is loaded successfuly. In order upgrade scenario, the user needs to know that the loaded module is actually the one with the targeted module version, in order to coordinated further actions. This PR introduces a new label: kmm.node.kubernetes.io/%s.%s.version.ready, whose value will signify the version of the currently loaded kernel module.
The changes in the code are:
1. Adding Version field both to status and spec of NMC (will be used to construct the new label)
2. Polulating the Version field of the NCM's spec during NMC update/create
3. Adding the "version" annotation to the worker pod, so that the version may be extracted for NMC's status
4. Changing the Node package API to receive maps instead of slice for labels to be added/remove. This is done since now we must also update the value of the label, and not just the key
5. Rewrite the UpdateNodeLabel function in the NMC controller to support new flow and new Node API
6. In case "ready" label is removed, we also automaticaly remove "version.ready" label. If it does not exists - nothing happens
7. In case "ready" label needs to be added, we check if the version field exists in the NMC status, and if it does, then we add the "version.ready" label with appropriate value

### DIFF
--- a/internal/controllers/nmc_reconciler.go
+++ b/internal/controllers/nmc_reconciler.go
@@ -102,7 +102,7 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}
 
 	errs := make([]error, 0, len(nmcObj.Spec.Modules)+len(nmcObj.Status.Modules))
-	var readyLabelsToRemove []string
+	readyLabelsToRemove := make(map[string]string)
 	for _, mod := range nmcObj.Spec.Modules {
 		moduleNameKey := mod.Namespace + "/" + mod.Name
 
@@ -110,7 +110,7 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 
 		// skipping handling NMC spec module until node is ready
 		if !r.nodeAPI.IsNodeSchedulable(&node, mod.Tolerations) {
-			readyLabelsToRemove = append(readyLabelsToRemove, utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name))
+			readyLabelsToRemove[utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name)] = ""
 			delete(statusMap, moduleNameKey)
 			continue
 		}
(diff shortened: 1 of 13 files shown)


config/manager/kustomization.yaml
```diff
@@ -10,8 +10,8 @@ patches:
 
 images:
 - name: controller
-  newName: gcr.io/k8s-staging-kmm/kernel-module-management-operator
-  newTag: latest
+  newName: quay.io/yshnaidm/kmmo
+  newTag: minikube-version-label
```
ybettan: This has been forgotten.

internal/controllers/nmc_reconciler.go
```diff
@@ -714,6 +670,75 @@ func (h *nmcReconcilerHelperImpl) RecordEvents(node *v1.Node, loadedModules, unl
...
+	versions := make(map[types.NamespacedName]string)
+
+	for _, module := range nmc.Status.Modules {
+		if module.Version != "" {
+			versions[types.NamespacedName{Namespace: module.Namespace, Name: module.Name}] = module.Version
+		}
+	}
+	return versions
+}
+
```
ybettan, line 740: Why did we move all those functions? It makes it harder to see the actual change on the PR.

internal/controllers/nmc_reconciler.go
```diff
@@ -668,20 +613,31 @@ func (h *nmcReconcilerHelperImpl) UpdateNodeLabels(ctx context.Context, nmc *kmm
...
 
-	var loadedLabels []string
-	unloadedLabels := deprecatedNodeModuleReadyLabels.UnsortedList()
+	loadedLabels := make(map[string]string)
+	unloadedLabels := deprecatedNodeModuleReadyLabels
 
 	for _, label := range nsnLabelsToBeRemoved {
-		unloadedLabels = append(unloadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
+		unloadedLabels[utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name)] = ""
+		//= append(unloadedLabels, utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name))
```
ybettan: Do we need to remove this line?

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: i moved them because they belong to the nmcHelper and not to the labelPreparationHelper. Now all the functions are grouped according to the interface, which actually make the code more readable

config/manager/kustomization.yaml
yevgeny-shnaidman: fixed

internal/controllers/nmc_reconciler.go
yevgeny-shnaidman: fixed

PR #1263, opened 2026-05-05
Add dev skills for local KMM development, e2e testing and order upgrade testing

TomerNewman:
Three AI-assisted skills for local development workflows:
- kmm-dev-setup: guides setting up minikube + deploying KMM from source code.
- kmm-e2e: runs the full build-sign-load pipeline.
- kmm-order-upgrade:  runs a full order upgrade local test on a multi node cluster.

---

### DIFF
--- a/.agents/skills/kmm-dev-setup/SKILL.md
+++ b/.agents/skills/kmm-dev-setup/SKILL.md
@@ -1,0 +1,120 @@
+---
+name: kmm-dev-setup
+description: >-
+  Set up a local KMM (Kernel Module Management) dev environment from scratch.
+  Use when the user asks to create, rebuild, redeploy, or tear down a KMM
+  development environment, or when you need a running KMM instance to test changes.
+---
+# KMM Dev Environment Setup
+
+All commands below assume you are in the repository root.
+
+Follow these steps to get a working KMM instance running your current code.
+
+## Step 1: Create a multi-node Minikube cluster with a local registry
+
+Check if a cluster already exists:
+
+```bash
+minikube status
+```
+
+- If the cluster is running **and** was created with `--driver=docker --nodes=3`, skip to Step 2.
+- If the cluster exists but uses the wrong driver or node count, delete it first with `minikube delete`.
+- If no cluster exists, create one:
+
+Start the local registry (removes any leftover container first):
+
+```bash
+docker rm -f kmm-registry 2>/dev/null || true
+docker run -d -p 5000:5000 --name kmm-registry registry:2
+```
+
+Create a 3-node cluster (1 control-plane + 2 workers) using the **Docker driver**:
+
+```bash
+minikube start --nodes=3 --driver=docker --insecure-registry=host.minikube.internal:5000
+```
+
+> **Important**: Always use `--driver=docker`. The Podman driver has broken pod DNS (pods cannot resolve external names like `gcr.io`), which breaks the Kaniko build/sign pipeline.
+
+Configure CRI-O on all nodes to trust the insecure registry (the `--insecure-registry` flag alone does not configure CRI-O):
+
+```bash
+for node in minikube minikube-m02 minikube-m03; do
+  minikube ssh -n $node -- sudo bash -c '"mkdir -p /etc/containers/registries.conf.d && cat > /etc/containers/registries.conf.d/host-minikube.conf <<EOF
+[[registry]]
+location = \"host.minikube.internal:5000\"
+insecure = true
+EOF"'
+  minikube ssh -n $node -- sudo systemctl restart crio
+done
+```
+
+To destroy the cluster when done:
+
+```bash
+minikube delete
+docker rm -f kmm-registry
+```
+
+## Step 2: Build images from current code
+
+Generate a unique tag and build only the images affected by your changes:
+
+```bash
+TAG=$(date +%Y%m%d%H%M%S)
+```
+
+Build the controller image (if controller code changed):
+
+```bash
+make docker-build IMG=kmm-controller:$TAG
+```
+
+Build the webhook image (if webhook code changed):
+
+```bash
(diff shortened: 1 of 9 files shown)


TomerNewman:
A thought I had, maybe instead of using Minikube we can Kind, thus having multiple nodes, and that way we can have a Skill to test Odered Upgrade flow, wdyt?

.agents/skills/kmm-dev-setup/SKILL.md
```diff
@@ -0,0 +1,113 @@
...
+  Use when the user asks to create, rebuild, redeploy, or tear down a KMM
+  development environment, or when you need a running KMM instance to test changes.
+---
+# KMM Dev Environment Setup
+
+All commands below assume you are in the repository root.
+
+Follow these steps to get a working KMM instance running your current code.
+
+## Step 1: Create a multi-node Minikube cluster with a local registry
```
ybettan, line 14: Why do we need a multi-node?

.cursor/skills/kmm-dev-setup/SKILL.md
```diff
@@ -0,0 +1,113 @@
...
+docker run -d -p 5000:5000 --name kmm-registry registry:2
+```
+
+Create a 3-node cluster (1 control-plane + 2 workers) using the **Docker driver**:
+
+```bash
+minikube start --nodes=3 --driver=docker --insecure-registry=host.minikube.internal:5000
+```
+
+Configure CRI-O on every node to trust the insecure registry:
```
ybettan: This is not related to CRIO. I this doc you are starting minikube with the default container runtime which isn't CRIO.

.agents/skills/kmm-dev-setup/SKILL.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+## Step 2: Build images from current code
+
+Build both images with a unique tag:
+
+```bash
+TAG=$(date +%Y%m%d%H%M%S)
+make docker-build IMG=kmm-controller:$TAG
+make webhookimage-build WEBHOOK_IMG=kmm-webhook:$TAG
```
ybettan, line 78: Let's tell it to only build images that changes (no need to rebuild the webhook if we just changed a controller).

.agents/skills/kmm-e2e/module.yaml
```diff
@@ -0,0 +1,37 @@
...
+            certSecret:
+              name: kmm-kmod-signing-cert
+            keySecret:
+              name: kmm-kmod-signing-key
+            unsignedImageRegistryTLS:
+              insecure: true
+            filesToSign:
+              - /opt/lib/modules/${KERNEL_FULL_VERSION}/kmm_ci_?.ko
+  selector:
+    kubernetes.io/hostname: minikube-m02
```
ybettan, line 37: Instead of duplicating the module, how about we tell the agent to take ci/prow/module.yaml and to only modify the relevant fields?

ybettan:
I think we should have all our skills gathered into a single `.agent` file or at least `.claude` since this is currently the go-to AI tool for coding.
See https://github.com/agentskills/agentskills/issues/15.

Then, each one can just locally link `.claude` or `.agent` to its preferred AI agent to make it compatible.

.agents/skills/kmm-dev-setup/SKILL.md
TomerNewman: In other skills where we need more than one node (order upgrade for example) we use this skill also.
And for the  rest of the skills that use it, its not harmful.

.agents/skills/kmm-e2e/module.yaml
TomerNewman: The downside of that is that we need to maintain a pretty long (and not that clean) "sed" command to replace the CI yaml with our relevant fields

TomerNewman:
> I think we should have all our skills gathered into a single `.agent` file or at least `.claude` since this is currently the go-to AI tool for coding. See [agentskills/agentskills#15](https://github.com/agentskills/agentskills/issues/15).
> 
> Then, each one can just locally link `.claude` or `.agent` to its preferred AI agent to make it compatible.

Its still a proposal on claude code see https://github.com/agentskills/agentskills/issues/15.
Also why to link locally? its pretty conmmon to use cursor nowdays, I would keep both of the .claude and .cursor paths since they are both common.

ybettan:
> Its still a proposal on claude code see https://github.com/agentskills/agentskills/issues/15.
> Also why to link locally? its pretty conmmon to use cursor nowdays, I would keep both of the .claude and .cursor paths since they are both common.

Although it is not adopted by all models yet, we shouldn't endorse any vendor in a CNCF project.

It has been announced [here](https://groups.google.com/u/1/a/kubernetes.io/g/dev/c/7Y9016gdFZw?pli=1).

Kubernetes tries to stay vendor-neutral and explicitly endorsing a specific vendor like cursor/claude can be an issue.
I did suggest having it all under `.claude` in a previous comment but after reading the [k8s ai-guidance](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance) I think it would be wiser to keep it agnostic.

Each contributor can then locally link `.agent` to what ever preferred AI tools they use.

.agents/skills/kmm-dev-setup/SKILL.md
yevgeny-shnaidman: i don't think we have ever tested multinode on upstream, even the ordered upgrade

TomerNewman: so why do we have order upgrade here?

TomerNewman:
When thinking about it, altough those skills benefits are quite good, I am not sure we have to use AI for them.
We can create scripts for local dev setup, local e2e and order upgrade testing, instead of wasting tokens on them with AI.
wdyt? @yevgeny-shnaidman @ybettan

PR #1265, opened 2026-05-07
Fix duplicate build/sign resource race causing infinite reconcile error

TomerNewman:
Replace GenerateName with a deterministic Name for build and sign pods.
The name is computed as {name}-{build|sign}-{KernelVersion}.

This addresses the root cause of a race between the informer cache read (GetResourceByKernel returns 0 results because the cache is stale) and the API server write (Create) that allowed two concurrent reconciles to each create a pod, leaving the operator stuck in an infinite "expected 0 or 1 … resources, got 2" error loop.

With deterministic names, both reconciles attempt to create the exact same pod name.
The first succeeds, the second receives AlreadyExists from the API server and treats it as a successful no-op.

### DIFF
--- a/internal/buildsign/manager.go
+++ b/internal/buildsign/manager.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -40,7 +41,7 @@ func NewManager(client client.Client, resourceManager ResourceManager, scheme *r
 func (m *manager) GetStatus(ctx context.Context, name, namespace, kernelVersion string,
 	action kmmv1beta1.BuildOrSignAction, owner metav1.Object) (kmmv1beta1.BuildOrSignStatus, error) {
 
-	normalizedKernel := kernel.NormalizeVersion(kernelVersion)
+	normalizedKernel := kernel.DNSSafeKernelVersion(kernelVersion)
 	foundResource, err := m.resourceManager.GetResourceByKernel(ctx, name, namespace, normalizedKernel, action, owner)
 	if err != nil {
 		if !errors.Is(err, ErrNoMatchingBuildSignResource) {
@@ -92,6 +93,10 @@ func (m *manager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushImage
 		logger.Info("Creating resource")
 		err = m.resourceManager.CreateResource(ctx, resourceTemplate)
 		if err != nil {
+			if k8serrors.IsAlreadyExists(err) {
+				logger.Info("Resource already exists, skipping creation (likely the client cache didn't update yet)")
+				return nil
+			}
 			return fmt.Errorf("could not create resource: %v", err)
 		}
 
--- a/internal/buildsign/resource/common.go
+++ b/internal/buildsign/resource/common.go
@@ -281,11 +281,11 @@ func (rm *resourceManager) makeBuildTemplate(ctx context.Context, mld *api.Modul
 
 	build := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
-			GenerateName: mld.Name + "-build-",
-			Namespace:    mld.Namespace,
-			Labels:       resourceLabels(mld.Name, mld.KernelNormalizedVersion, kmmv1beta1.BuildImage),
-			Annotations:  map[string]string{constants.ResourceHashAnnotation: fmt.Sprintf("%d", buildSpecHash)},
-			Finalizers:   []string{constants.GCDelayFinalizer, constants.JobEventFinalizer},
+			Name:        mld.Name + "-build-" + mld.KernelNormalizedVersion,
+			Namespace:   mld.Namespace,
+			Labels:      resourceLabels(mld.Name, mld.KernelNormalizedVersion, kmmv1beta1.BuildImage),
+			Annotations: map[string]string{constants.ResourceHashAnnotation: fmt.Sprintf("%d", buildSpecHash)},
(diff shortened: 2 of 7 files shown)


internal/buildsign/resource/common.go
```diff
@@ -245,6 +245,18 @@ func moduleLabels(moduleName string, resourceType kmmv1beta1.BuildOrSignAction)
 	}
 }
 
+func DeterministicPodName(moduleName string, kernelVersion string, action kmmv1beta1.BuildOrSignAction) string {
```
yevgeny-shnaidman: 1. i would change it to: generatePodName
2. we already have a KernelNormalizedVersion field in the mld. Does why do we need to run NewReplacer, if we are using that field

ybettan:
In the commit message I would replace
```
that allowed two concurrent reconciles to each create a pod
```

with 
```
that allowed two fast reconcile events, each creating a pod
```
Since 2 concurent reconcile loops aren't possible when we use the default `MaxConcurrentReconciles=1` value.

internal/buildsign/resource/common.go
ybettan: Let's use `Resource` instead of `Pod` in the function name as we do for other function names since the resource can be a pod/build depending on the project.

internal/buildsign/resource/common.go
```diff
@@ -245,6 +245,18 @@ func moduleLabels(moduleName string, resourceType kmmv1beta1.BuildOrSignAction)
...
+	actionStr := "build"
+	if action == kmmv1beta1.SignImage {
+		actionStr = "sign"
+	}
+
+	dnsKernel := strings.NewReplacer("_", "-", ".", "-").Replace(strings.ToLower(kernelVersion))
+	dnsKernel = strings.Trim(dnsKernel, "-")
```
ybettan: Why don't we use the kernel/normelizedKernel as is?

internal/buildsign/resource/resourcemanager.go
```diff
@@ -86,7 +86,8 @@ func (rm *resourceManager) GetResourceByKernel(ctx context.Context, name, namesp
 	numFoundResources := len(moduleOwnedResources)
 	if numFoundResources == 0 {
 		return nil, buildsign.ErrNoMatchingBuildSignResource
-	} else if numFoundResources > 1 {
+	}
+	if numFoundResources > 1 {
```
ybettan: Is this supposed to be part of the PR?

internal/buildsign/manager.go
```diff
@@ -92,6 +93,10 @@ func (m *manager) Sync(ctx context.Context, mld *api.ModuleLoaderData, pushImage
 		logger.Info("Creating resource")
 		err = m.resourceManager.CreateResource(ctx, resourceTemplate)
 		if err != nil {
+			if k8serrors.IsAlreadyExists(err) {
+				logger.Info("Resource already exists, skipping creation (likely created by a concurrent reconcile)")
```
ybettan:
```suggestion
				logger.Info("Resource already exists, skipping creation (likely the client cache didn't update yet)")
```

internal/buildsign/resource/common.go
TomerNewman: 1. changing to generateResoruceName
2. in pods name you are not allowed to use `.` or `_`  for example, and  `KernelNormalizedVersion` can contains this.

internal/buildsign/resource/common.go
TomerNewman: in pods name you are not allowed to use `.` or `_` for example, and KernelNormalizedVersion can contain this.

internal/buildsign/resource/resourcemanager.go
TomerNewman: will revert it

internal/buildsign/resource/common.go
yevgeny-shnaidman: The KernelNormalizedVersion is created in the https://github.com/kubernetes-sigs/kernel-module-management/blob/7f06f37e40616829f8e2650d87c99b2634aeda48/internal/controllers/mbsc_reconciler.go#L146
It is used only by MBSC controller. WDYT about creating  KernelNormalizedVersion in https://github.com/kubernetes-sigs/kernel-module-management/blob/7f06f37e40616829f8e2650d87c99b2634aeda48/internal/controllers/mbsc_reconciler.go#L187 already in the way that we need, and then the name creation in Build or Sign can be simply:
Name: mld.Name + "-" + "build/sign-" + mld.KernelNormalizedVersion

internal/buildsign/resource/resourcemanager.go
yevgeny-shnaidman: why?It is a more correct version of the code, no?

TomerNewman: correct but not related to the commit, we can do it in a separate commit/pr

internal/buildsign/resource/common.go
TomerNewman: We use in `KernelNormalizedVersion` [here](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/internal/buildsign/resource/common.go#L279).
Wouldn't it be a problem in upgrade scenario?

TomerNewman: And also [here](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/internal/module/kernelmapper.go#L152) which leads to [here](https://github.com/kubernetes-sigs/kernel-module-management/blob/main/internal/utils/replaceStrings.go#L19).

meaning it will change KERNEL_FULL_VERSION and KERNEL_VERSION that are passed as env

yevgeny-shnaidman: the onces used in the builsign are used for labeling, so as long as we are consistent (setting the labels and quiring the labels with the same field) it should be ok. The last one is not used by the build/sign code only by module reconciler (please correct me if i am wrong)

ybettan:
/lgtm
/hold
@yevgeny-shnaidman Can unhold when he is happy with the changes.

PR #1269, opened 2026-05-19
Add DRA support to the Module CRD

TomerNewman:
DRA is the Kubernetes (from 1.35) framework for advertising and claiming hardware devices, replacing the Device Plugin API.

In this commit we firstly introducing support in KMM for DRA.
We extend the Module CRD with a spec.dra section for declaring a DRA kubelet plugin.

Notice that I've added security context to the API since I can see that vendors use privileges=true for their DRA driver pod, for example [here](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/blob/main/deployments/helm/dra-driver-nvidia-gpu/values.yaml#L283) (it does not exist in Device Plugin though).

---


api/v1beta1/module_types.go
```diff
@@ -321,13 +322,101 @@ type DevicePluginSpec struct {
 	AutomountServiceAccountToken *bool `json:"automountServiceAccountToken,omitempty"`
 }
 
+type DRAContainerSpec struct {
```
yevgeny-shnaidman: This is basically a copy-paste of DevicePluginContainerSpec, right? Can't we just  move it into a dedicated struct and let DevicePluginSpec and DRASepc to use it (in the unnamed manner, in order not to break backport compatability

api/v1beta1/module_types.go
```diff
@@ -321,13 +322,101 @@ type DevicePluginSpec struct {
...
+	// +kubebuilder:validation:MinLength=1
+	Name string `json:"name"`
+
+	// Spec is the upstream DeviceClassSpec applied to the created DeviceClass.
+	Spec resourcev1.DeviceClassSpec `json:"spec"`
+}
+
+type DRASpec struct {
+	// Container holds the properties for the DRA kubelet plugin container.
+	Container DRAContainerSpec `json:"container"`
```
yevgeny-shnaidman: The same comment as above for the following fields:
Container, InitContainer,ServiceaccountName,Volumes, AutomountServiceAccountToken

api/v1beta1/module_types.go
TomerNewman: Right I forgot we can use inline, I will update

api/v1beta1/module_types.go
TomerNewman: `Container` and `InitContainer` can not be shared though

TomerNewman: `DevicePluginSpec` uses `DevicePluginContainerSpec` but `DRASpec` uses `DRAContainerSpec` (which adds `SecurityContext`)

TomerNewman: I can move `SecurityContext` to the shared struct but this will cause DevicePlugin to have access to SecurityContext field, and it did not had this access before

ybettan:
I assume that DRA and devicePlugin are mutually exclusive? If this is the case WDTY about adding a webhook validation that only one of them is defined in the module as part of this CRD PR? 

*EDIT*
Ignore it, I just saw https://github.com/kubernetes-sigs/kernel-module-management/pull/1269/changes#diff-02968acb7754c416e4513c83e659a7477fd4cda8ec3dfab413fe4da23cf0196cR408

ybettan:
/lgtm
/hold
@yevgeny-shnaidman Can unhold when he is happy with the changes.

PR #1274, opened 2026-05-25
Fix device plugin pod not evicted during node drain

TomerNewman:
## Summary

- Fixes a deadlock during node drain where the device plugin DaemonSet pod remains running (because it auto-tolerates `NoSchedule` taints and its `nodeSelector` only required `kmm-ready`), blocking the module unloader from proceeding.
- Introduces a new `device-plugin-target` node label managed by the `DevicePluginReconciler`. The DaemonSet `nodeSelector` now requires both `kmm-ready` AND `device-plugin-target`. When a node becomes unschedulable, the reconciler removes `device-plugin-target`, causing the DaemonSet controller to evict the device plugin pod and break the deadlock.
- The NMC reconciler defers unloader creation while the `device-plugin-ready` label is present on an unschedulable node, ensuring the unloader only runs after the device plugin pod terminates.

## Design

The `DevicePluginReconciler` now watches node events (taint and label changes) and:
1. **Adds** `device-plugin-target` to schedulable nodes that have `kmm-ready`
2. **Removes** `device-plugin-target` from unschedulable nodes (triggering pod eviction)
3. **Cleans up** `device-plugin-target` labels on module deletion

This keeps `kmm-ready` semantics intact (it still means "kmod is loaded") while giving the system a lever to evict device plugin pods during drain.

## Test plan

- [x] Unit tests updated for `DevicePluginReconciler` (new `handleDevicePluginTargetLabels` and `removeDevicePluginTargetLabels` coverage)
- [x] Unit tests updated for `NMCReconciler` (orphan status deferral logic)
- [x] Verified on minikube: `NoSchedule` taint causes `device-plugin-target` removal → pod eviction
- [x] Verified recovery: removing taint re-adds `device-plugin-target` → pod rescheduled
- [ ] E2E test with real kernel module load + drain cycle

### DIFF
--- a/internal/controllers/device_plugin_reconciler.go
+++ b/internal/controllers/device_plugin_reconciler.go
@@ -20,13 +20,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 	"strings"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/metrics"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
@@ -34,8 +34,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
@@ -73,6 +75,11 @@ func (r *DevicePluginReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&kmmv1beta1.Module{}).
 		Owns(&appsv1.DaemonSet{}).
+		Watches(
+			&v1.Node{},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindModulesForNode),
+			builder.WithPredicates(filter.DevicePluginReconcilerNodePredicate()),
+		).
 		Named(DevicePluginReconcilerName).
 		Complete(
 			reconcile.AsReconciler[*kmmv1beta1.Module](r.client, r),
@@ -90,11 +97,18 @@ func (r *DevicePluginReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.
 	}
 
 	if mod.GetDeletionTimestamp() != nil {
+		if err = r.reconHelperAPI.removeDevicePluginTargetLabels(ctx, mod); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not remove device-plugin-target labels on deletion: %v", err)
+		}
 		err = r.reconHelperAPI.handleModuleDeletion(ctx, existingDevicePluginDS)
 		return ctrl.Result{}, err
 	}
 
 	r.reconHelperAPI.setKMMOMetrics(ctx)
+
+	if err = r.reconHelperAPI.handleDevicePluginTargetLabels(ctx, mod); err != nil {
+		return res, fmt.Errorf("could not reconcile device-plugin-target labels: %v", err)
+	}
 
 	err = r.reconHelperAPI.handleDevicePlugin(ctx, mod, existingDevicePluginDS)
 	if err != nil {
@@ -122,6 +136,8 @@ func (r *DevicePluginReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.
 type devicePluginReconcilerHelperAPI interface {
 	setKMMOMetrics(ctx context.Context)
 	handleDevicePlugin(ctx context.Context, mod *kmmv1beta1.Module, existingDevicePluginDS []appsv1.DaemonSet) error
+	handleDevicePluginTargetLabels(ctx context.Context, mod *kmmv1beta1.Module) error
+	removeDevicePluginTargetLabels(ctx context.Context, mod *kmmv1beta1.Module) error
 	garbageCollect(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet) error
 	handleModuleDeletion(ctx context.Context, existingDevicePluginDS []appsv1.DaemonSet) error
 	moduleUpdateDevicePluginStatus(ctx context.Context, mod *kmmv1beta1.Module, existingDevicePluginDS []appsv1.DaemonSet) error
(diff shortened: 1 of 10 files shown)


ybettan:
In the commit message, there is `The PR was made by Bugs Buddy - AI workflow` which is not allowed based on the [ai-guidance](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance) k8s policy.

internal/controllers/nmc_reconciler.go
```diff
@@ -133,6 +133,14 @@ func (r *NMCReconciler) Reconcile(ctx context.Context, req reconcile.Request) (r
 	for statusNameKey, status := range statusMap {
 		logger := logger.WithValues("status", statusNameKey)
 
+		if !r.nodeAPI.IsNodeSchedulable(&node, status.Tolerations) {
```
yevgeny-shnaidman: This code is not needed at all: NMC controller should not deal with anything related to device plugin. DevicePlugin reconciler will take care of the device plugin pod, and that will eventually lead to worker pod to succeed

internal/controllers/nmc_reconciler.go
```diff
@@ -628,6 +636,7 @@ func (h *nmcReconcilerHelperImpl) UpdateNodeLabels(ctx context.Context, nmc *kmm
 
 	for _, label := range nsnLabelsToBeRemoved {
 		unloadedLabels[utils.GetKernelModuleReadyNodeLabel(label.Namespace, label.Name)] = ""
+		unloadedLabels[utils.GetDevicePluginTargetNodeLabel(label.Namespace, label.Name)] = ""
```
yevgeny-shnaidman: the same as above

internal/controllers/nmc_reconciler_test.go
```diff
@@ -303,6 +303,7 @@ var _ = Describe("NodeModulesConfigReconciler_Reconcile", func() {
 			wh.EXPECT().ProcessModuleSpec(contextWithValueMatch, nmc, &spec0, &status0, &node),
 			nm.EXPECT().IsNodeSchedulable(&node, nil).Return(true),
 			wh.EXPECT().ProcessModuleSpec(contextWithValueMatch, nmc, &spec1, nil, &node),
+			nm.EXPECT().IsNodeSchedulable(&node, nil).Return(true),
```
yevgeny-shnaidman: no need for any changes in NMC reconciler or unit-tests

internal/filter/filter.go
```diff
@@ -187,6 +187,13 @@ func ModuleReconcilerNodePredicate() predicate.Predicate {
 	)
 }
 
+func DevicePluginReconcilerNodePredicate() predicate.Predicate {
+	return predicate.And(
+		skipDeletions,
+		predicate.Or(nodeTaintsChanged, predicate.LabelChangedPredicate{}),
```
yevgeny-shnaidman: does it cover node creation? why do we need label change predicate?

pkg/labels/labels.go
```diff
@@ -12,6 +12,10 @@ func GetDevicePluginNodeLabel(namespace, moduleName string) string {
 	return utils.GetDevicePluginNodeLabel(namespace, moduleName)
 }
 
+func GetDevicePluginTargetNodeLabel(namespace, moduleName string) string {
```
yevgeny-shnaidman: why is it needed for external package? Should KMM user be aware of this label?

internal/controllers/device_plugin_reconciler.go
```diff
@@ -195,6 +211,65 @@ func (dprh *devicePluginReconcilerHelper) handleDevicePlugin(ctx context.Context
...
+	readyLabel := utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name)
+	targetLabel := utils.GetDevicePluginTargetNodeLabel(mod.Namespace, mod.Name)
+
+	nodeList := v1.NodeList{}
+	if err := dprh.client.List(ctx, &nodeList, client.MatchingLabels{readyLabel: ""}); err != nil {
+		return fmt.Errorf("could not list nodes with ready label: %v", err)
+	}
+
+	var errs []error
+	for i := range nodeList.Items {
```
yevgeny-shnaidman: We can significantly reduce this function. The pseudo-code can be:
1. Get list of node targeted by Module's node selector.
2. Per each node:
    a. if node schedulable  - add device plugin label
    b. if node no-nschedulable - remove device plugin label

UpdateNode API uses patch, which means that if we do't actually change labels, then there is no actual request to API Server

yevgeny-shnaidman: additional note: we are looking at nodes that are targeted by the KMM Module. In case KMM Module is changed or the node selector's label is removed from the node, this code won't remove the label from the node. Not sure if it is significant or not. The label will eventually be cleanoff when the Module is deleted

internal/controllers/device_plugin_reconciler.go
```diff
@@ -195,6 +211,57 @@ func (dprh *devicePluginReconcilerHelper) handleDevicePlugin(ctx context.Context
...
+// This enables the device plugin DaemonSet to be evicted from draining nodes.
+func (dprh *devicePluginReconcilerHelper) handleDevicePluginTargetLabels(ctx context.Context, mod *kmmv1beta1.Module) error {
+	if mod.Spec.DevicePlugin == nil {
+		return nil
+	}
+
+	targetLabel := utils.GetDevicePluginTargetNodeLabel(mod.Namespace, mod.Name)
+
+	nodeList := v1.NodeList{}
+	if err := dprh.client.List(ctx, &nodeList, client.MatchingLabels(mod.Spec.Selector)); err != nil {
```
yevgeny-shnaidman: We have a dedicated API for that: nodeAPI.GetNodesListBySelector

TomerNewman: `GetNodesListBySelector` filters to only schedulable nodes (it calls `IsNodeSchedulable` internally and excludes unschedulable ones). Here we need both schedulable and unschedulable nodes in order to add the label to schedulable nodes and remove it from unschedulable ones. Using `GetNodesListBySelector` would only give us the "add" path — we would lose visibility into unschedulable nodes that need the label removed.

yevgeny-shnaidman: if you pass the tolerations parameter as nil, then this function will return all the nodes targeted by the NodeSelector field, regardless if the node is schedulable or not

TomerNewman: Good point. I extracted the first part of `GetNodesListBySelector` into a new `GetAllNodesBySelector` method on the `Node` interface (returns all nodes matching a selector without filtering by schedulability). `GetNodesListBySelector` now calls it internally. The device plugin reconciler uses `GetAllNodesBySelector` to get all targeted nodes and then decides per-node whether to add or remove the label based on schedulability.

internal/node/node.go
```diff
@@ -14,6 +14,7 @@ import (
 
 type Node interface {
 	IsNodeSchedulable(node *v1.Node, tolerations []v1.Toleration) bool
+	GetAllNodesBySelector(ctx context.Context, selector map[string]string) ([]v1.Node, error)
 	GetNodesListBySelector(ctx context.Context, selector map[string]string, tolerations []v1.Toleration) ([]v1.Node, error)
```
yevgeny-shnaidman: GetAllNodesBySelector and GetNodesListBySelector are confusing names. Lets change the GetNodesListBySelector to GetSchedulableNodes. WDYT?

ybettan:
I don't understand how this PR fixes the issue. It might mitigate it but the race condition still exists.

Current race (module reconciler runs before NMC reconciler):
1. Module reconciler is removing the NMC spec entry
2. NMC reconciler first unload the kmod and fail since the devicePlugin is still running (only later in the flow NMC is updating the ready label)

New race (module reconciler --> NMC reconciler --> DevicePluging reconciler)
Same race is happening here.

Why don't we split the update of labels by NMC into 2:
1. Removing ready labels happens before processing the spec (unloading the kmod)
2. Updating the labels (not sure if we still need label removal here or just label addition) after processing the NMC spec.

yevgeny-shnaidman:
> I don't understand how this PR fixes the issue. It might mitigate it but the race condition still exists.
> 
> Current race (module reconciler runs before NMC reconciler):
> 
> 1. Module reconciler is removing the NMC spec entry
> 2. NMC reconciler first unload the kmod and fail since the devicePlugin is still running (only later in the flow NMC is updating the ready label)
> 
> New race (module reconciler --> NMC reconciler --> DevicePluging reconciler) Same race is happening here.
> 
> Why don't we split the update of labels by NMC into 2:
> 
> 1. Removing ready labels happens before processing the spec (unloading the kmod)
> 2. Updating the labels (not sure if we still need label removal here or just label addition) after processing the NMC spec.

This fixes the race condition since the labels are removed, the device plugin pod stops running, so the worker pod will eventually succeed without any external intervention. 
Removing the "ready" label prior to actually successfully removing the kernel module can lead to a situation where the module is still loaded , while the label is missing, which breaks are API contract

internal/node/node.go
TomerNewman: changed to `GetSchedulableNodesBySelector`

PR #1276, opened 2026-05-27
Rename DevicePluginContainerSpec to CommonContainerSpec

TomerNewman:
### Summary

Introduces `CommonContainerSpec` as the canonical container configuration type, with `DevicePluginContainerSpec` kept as a backward-compatible type alias. This prepares for DRA support where `DRASpec` will reuse `CommonContainerSpec` for its driver container configuration.

### Changes

- **`api/v1beta1/module_types.go`:**
  - Renamed `DevicePluginContainerSpec` struct to `CommonContainerSpec`
  - Added `type DevicePluginContainerSpec = CommonContainerSpec` (backward-compatible alias)
  - `DevicePluginSpec` unchanged — keeps its fields as-is
- **`api/v1beta1/zz_generated.deepcopy.go`:** Regenerated — deepcopy methods now on `CommonContainerSpec`
- **`internal/controllers/device_plugin_reconciler_test.go`:** Added backward compatibility test verifying JSON serialization is unchanged

### Testing

- **Unit tests:** All existing tests pass without any changes to test logic
- **Backward compatibility:** New test verifies `DevicePluginSpec` JSON field names are unchanged
- **CRD diff:** Zero — `kmm.sigs.x-k8s.io_modules.yaml` is identical before and after
- **Validation:** `make generate`, `make manifests`, `make fmt`, `make vet`, `make lint`, `make unit-test` all pass

### Acceptance Criteria

- [x] `CommonContainerSpec` is defined with the same fields as the original `DevicePluginContainerSpec`
- [x] `DevicePluginContainerSpec` is a backward-compatible type alias for `CommonContainerSpec`
- [x] `DevicePluginSpec` struct is unchanged — no embedding, no field changes
- [x] All existing references compile without modification
- [x] Existing unit tests pass without behavioral change
- [x] CRD schema is identical before and after


api/v1beta1/module_types.go
```diff
@@ -300,12 +300,15 @@ type DevicePluginContainerSpec struct {
 	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
 }
 
-type DevicePluginSpec struct {
-	Container DevicePluginContainerSpec `json:"container"`
+// DevicePluginContainerSpec is a backward-compatible alias for PluginContainerSpec.
+type DevicePluginContainerSpec = PluginContainerSpec
```
yevgeny-shnaidman: why do we need this?

api/v1beta1/module_types.go
```diff
@@ -300,12 +300,15 @@ type DevicePluginContainerSpec struct {
...
+
+type PluginSpec struct {
```
yevgeny-shnaidman: Why do we need PluginSpec and DevicePLuginSpec? We can keep DevicePluginSpec, which will contain the Container and InitContainer and PluginContainerSpec field

api/v1beta1/module_types.go
TomerNewman: The type alias keeps all existing code that references `DevicePluginContainerSpec` compiling without changes — controllers, tests, and the webhook all use this name. In the next PR (Story 2), the new `DRASpec` will reference `PluginContainerSpec` directly, so the shared type gets a generic name while the alias preserves backward compatibility.

api/v1beta1/module_types.go
TomerNewman: `PluginSpec` is the shared base that both `DevicePluginSpec` and the upcoming `DRASpec` (Story 2) will use. DRA drivers need the same pod-level fields — container, init container, service account, volumes. Extracting them into `PluginSpec` now avoids duplicating those fields across both specs. `DevicePluginSpec` wraps it with `json:",inline"` so the existing CRD schema is unchanged (verified via zero CRD diff).

ybettan:
Can we remove `[MGMT-24467]` from the commit message? (It is not relevant to this project).

ybettan:
Something isn't clear to me.

1. This PR is just renaming struct without any other change
2. How this PR is related to https://github.com/kubernetes-sigs/kernel-module-management/pull/1269?
3. If this PR is just a "preparation" to [1269](https://github.com/kubernetes-sigs/kernel-module-management/pull/1269) I think it would be clearer to have a single PR since this PR has no purpose on its own and the commit message is confusing since nothing is really extracted.

internal/controllers/device_plugin_reconciler_test.go
```diff
@@ -606,8 +608,10 @@ var _ = Describe("DevicePluginReconciler_setDevicePluginAsDesired", func() {
 					},
 				},
 				DevicePlugin: &kmmv1beta1.DevicePluginSpec{
-					Container: kmmv1beta1.DevicePluginContainerSpec{Image: devicePluginImage},
-					Volumes:   []v1.Volume{vol},
+					CommonSpec: kmmv1beta1.CommonSpec{
```
yevgeny-shnaidman: Lets add a unit test that verifies the backword compatibility also

TomerNewman: added "DevicePluginSpec backward compatibility" unit test

yevgeny-shnaidman:
/lgtm
/hold
/assign @ybettan 
@ybettan please unhold once you are satisfied

ybettan:
@yevgeny-shnaidman @TomerNewman I don't understand what do we do in this PR besides renaming `DevicePluginContainerSpec` --> `CommonContainerSpec`.

I don't see any thing that is being "extracted" in order to get "reused".

yevgeny-shnaidman:
> @yevgeny-shnaidman @TomerNewman I don't understand what do we do in this PR besides renaming `DevicePluginContainerSpec` --> `CommonContainerSpec`.
> 
> I don't see any thing that is being "extracted" in order to get "reused".

The next PR will provide the definition for the DRA Spec which will reuse the CommonContainerSpec

TomerNewman:
> > @yevgeny-shnaidman @TomerNewman I don't understand what do we do in this PR besides renaming `DevicePluginContainerSpec` --> `CommonContainerSpec`.
> > I don't see any thing that is being "extracted" in order to get "reused".
> 
> The next PR will provide the definition for the DRA Spec which will reuse the CommonContainerSpec

correct but the commit message was not accurate since it included the previous implementation, I've amended it now.

PR #1282, opened 2026-06-09
Add DRA webhook validation with Kubernetes version gating

TomerNewman:
## Summary
- Implement admission webhook validation for Modules with `spec.dra`, gating DRA usage to Kubernetes 1.34+ using a major-version-aware comparison (so future k8s 2.x+ is automatically accepted)
- Detect the cluster Kubernetes version at webhook startup via the discovery client and pass it through to validators
- Validate `driverName` format/presence and `deviceClasses` name uniqueness/format

Note: mutual exclusivity of `spec.dra` and `spec.devicePlugin` is already enforced by the CEL `XValidation` rule on `ModuleSpec` and is not duplicated in the webhook.

## Changes
- `internal/webhook/module.go`: Add `KubeVersion` struct, `ParseKubeVersion`, `DiscoverKubeVersion`, and `validateDRA` functions; thread `KubeVersion` through `ModuleValidator` and `validateModule`
- `internal/webhook/hub/managedclustermodule_webhook.go`: Pass `KubeVersion` through to `NewModuleValidator`
- `internal/constants/constants.go`: Add `MinKubeMajorForDRA` and `MinKubeMinorForDRA` constants
- `cmd/webhook-server/main.go`: Discover Kubernetes version at startup and pass to webhook validators
- `internal/webhook/module_test.go`: Add tests for `ParseKubeVersion` (including k8s 2.x), `validateDRA` (version gate boundaries, field validation)

## Test plan
- [x] Unit tests for `ParseKubeVersion` cover standard versions, build metadata, pre-release, no `v` prefix, hypothetical k8s 2.x
- [x] Unit tests for `validateDRA` cover version gate (k8s 0.99 reject, 1.33 reject, 1.34 accept, 1.35 accept, 2.0 accept), driverName validation, deviceClasses validation
- [x] All webhook tests pass
- [x] `go build ./...` compiles cleanly
- [x] Verified `DiscoverKubeVersion` against a live minikube cluster (k8s v1.35.1 → major=1, minor=35)

### DIFF
--- a/internal/webhook/module.go
+++ b/internal/webhook/module.go
@@ -20,15 +20,20 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/validation"
-	"regexp"
-	"strings"
 
 	"github.com/go-logr/logr"
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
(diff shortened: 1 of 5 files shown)


internal/webhook/module.go
```diff
@@ -127,6 +178,50 @@ func validateModule(mod *kmmv1beta1.Module) (admission.Warnings, error) {
...
+	}
+
+	if kubeVersion.Major < constants.MinKubeMajorForDRA || (kubeVersion.Major == constants.MinKubeMajorForDRA && kubeVersion.Minor < constants.MinKubeMinorForDRA) {
+		return fmt.Errorf(
+			"spec.dra requires Kubernetes %d.%d or later; current cluster version is %d.%d",
+			constants.MinKubeMajorForDRA, constants.MinKubeMinorForDRA, kubeVersion.Major, kubeVersion.Minor,
+		)
+	}
+
+	if mod.Spec.DevicePlugin != nil {
```
yevgeny-shnaidman: do we still need it, if we have defined mutual exclusion via kubebuilder

internal/webhook/module.go
```diff
@@ -20,28 +20,75 @@ import (
...
+// DiscoverKubeVersion queries the Kubernetes API server and returns its version.
+func DiscoverKubeVersion(cfg *rest.Config) (KubeVersion, error) {
+	discClient, err := discovery.NewDiscoveryClientForConfig(cfg)
+	if err != nil {
+		return KubeVersion{}, fmt.Errorf("creating discovery client: %v", err)
+	}
+
+	serverVersion, err := discClient.ServerVersion()
+	if err != nil {
+		return KubeVersion{}, fmt.Errorf("querying Kubernetes server version: %v", err)
```
ybettan: "failed to create discovery client..."

internal/webhook/module.go
```diff
@@ -127,6 +178,46 @@ func validateModule(mod *kmmv1beta1.Module) (admission.Warnings, error) {
...
+	if driverName == "" {
+		return errors.New("spec.dra.driverName is required")
+	}
+
+	if errs := validation.IsDNS1123Subdomain(driverName); len(errs) > 0 {
+		return fmt.Errorf("spec.dra.driverName %q is not a valid DNS subdomain: %s", driverName, strings.Join(errs, "; "))
+	}
+
+	seen := sets.New[string]()
+	for i, dc := range mod.Spec.DRA.DeviceClasses {
```
ybettan, line 203: Can the deviceClasses be an empty list? In the current webhook this is allowed.

internal/webhook/module.go
TomerNewman: fixing

internal/webhook/module.go
TomerNewman: good question, is it possible that a user will make his dra driver create the device classes by himself?
if not, then it should be required as you suggested.
wdyt @yevgeny-shnaidman ?

yevgeny-shnaidman: deviceClasses can be an empty list. Users might decide that they are going to deploy deviceClasses by themselves, or a maybe at a later stage

TomerNewman: Then I will keep it as it is

internal/webhook/module.go
```diff
@@ -20,28 +20,75 @@ import (
...
+	if err != nil {
+		return KubeVersion{}, fmt.Errorf("failed to query Kubernetes server version: %v", err)
+	}
+
+	return ParseKubeVersion(serverVersion.GitVersion)
+}
+
+// ParseKubeVersion extracts the major and minor version from a Kubernetes
+// GitVersion string such as "v1.34.0" or "v1.34.0+k3s1".
+func ParseKubeVersion(gitVersion string) (KubeVersion, error) {
```
yevgeny-shnaidman: This function is not used outside of the module package, lets make it internal

PR #1283, opened 2026-06-09
Add DRA DaemonSet reconciler

TomerNewman:
### Summary
Implements a `DRAReconciler` that manages DRA driver DaemonSets from `Module.spec.dra`, mirroring the `DevicePluginReconciler` pattern. Also improves the `DevicePluginReconciler` with role labeling, cleanup on `spec.devicePlugin` removal, and clearer naming.

### Changes
- **`internal/constants/constants.go`:** Add `DRARoleLabelValue = "dra"` and `DevicePluginRoleLabelValue = "device-plugin"` constants
- **`internal/controllers/dra_reconciler.go`:** New DRA reconciler with helper and DaemonSet creator — creates DaemonSets with mandatory kubelet-plugins host-path volumes, DRA-specific container names, `DaemonSetRole` label, and status patching
- **`internal/controllers/dra_reconciler_test.go`:** Comprehensive unit tests covering all reconcile paths, helper methods, and DaemonSet spec generation
- **`internal/controllers/device_plugin_reconciler.go`:**
  - Fix `getModuleDevicePluginDaemonSets` to exclude DRA-labeled DaemonSets
  - Add `DevicePluginRoleLabelValue` label to device plugin DaemonSets (exclusion-based query kept for backward compat, with `TODO(3.0)` to switch)
  - Add cleanup logic when `spec.devicePlugin` is removed (delete DaemonSets + clear status)
  - Rename `handleModuleDeletion` → `deleteDevicePluginDaemonSets` for clarity
- **`internal/controllers/device_plugin_reconciler_test.go`:** Tests for DRA exclusion, `spec.devicePlugin` nil cleanup, `clearDevicePluginStatus`, and updated label expectations

### Testing
- **Unit tests:** Full coverage of `Reconcile` (good flow, error flows, deletion, spec removal), `handleDRA` (create/update), `deleteDRADaemonSets`, `moduleUpdateDRAStatus`, `clearDRAStatus`, `setDRAAsDesired` (with/without init container, volumes, labels), `getModuleDRADaemonSets`, `clearDevicePluginStatus`, and device plugin nil cleanup paths
- **Integration tests:** N/A
- **Coverage:** 100% on all new behavioral functions

### Acceptance Criteria

- [x] AC-1: `DRAReconciler` watches `Module` and owns `DaemonSet` resources
- [x] AC-2: Reconcile is a no-op when `spec.dra == nil`
- [x] AC-3: DaemonSet is created with `GenerateName: mod.Name + "-dra-"`
- [x] AC-4: DaemonSet auto-mounts `kubelet-plugins` and `kubelet-plugins-registry` host paths; user volumes appended after
- [x] AC-5: Container names are `dra-init` (init) and `dra` (main)
- [x] AC-6: DaemonSet pod labels include `ModuleNameLabel`, `DaemonSetRole: DRARoleLabelValue`
- [x] AC-7: DaemonSet `nodeSelector` includes `GetKernelModuleReadyNodeLabel(ns, name)`
- [x] AC-8: DaemonSet is updated when Module spec changes and deleted when `spec.dra` is removed or Module is deleted
- [x] AC-9: `status.dra` is patched mirroring `moduleUpdateDevicePluginStatus`
- [x] AC-10: Unit tests verify DaemonSet spec generation, early exit, create/update/delete paths, and status patching


internal/controllers/device_plugin_reconciler.go
```diff
@@ -178,7 +178,8 @@ func (dprh *devicePluginReconcilerHelper) getModuleDevicePluginDaemonSets(ctx co
 	devicePluginsList := make([]appsv1.DaemonSet, 0, len(dsList.Items))
 	// remove the older version module loader daemonsets
 	for _, ds := range dsList.Items {
-		if ds.GetLabels()[constants.DaemonSetRole] != constants.ModuleLoaderRoleLabelValue {
+		role := ds.GetLabels()[constants.DaemonSetRole]
```
yevgeny-shnaidman, line 192: what about adding Role label to the device plugin daemonset? This way you can add the Role label to opts := []client.ListOption{....}
and get the needed DS without afterwards going over the label.

internal/controllers/dra_reconciler.go
```diff
@@ -0,0 +1,326 @@
...
+
+	return res, nil
+}
+
+//go:generate mockgen -source=dra_reconciler.go -package=controllers -destination=mock_dra_reconciler.go draReconcilerHelperAPI,draDaemonSetCreator
+
+type draReconcilerHelperAPI interface {
+	getModuleDRADaemonSets(ctx context.Context, name, namespace string) ([]appsv1.DaemonSet, error)
+	handleDRA(ctx context.Context, mod *kmmv1beta1.Module, existingDRADS []appsv1.DaemonSet) error
+	handleModuleDeletion(ctx context.Context, existingDRADS []appsv1.DaemonSet) error
```
yevgeny-shnaidman: This function is used to delete DRA DS both when the Module is deleted, and when the Module is updated and its DRA field is set to nil. WDYT about renaming it to: deleteDRADaemonset, or something like that?

yevgeny-shnaidman:
we should also add a code to the DevicePluginReconciler, that deletes the DevicePlugin DS in case device plugin is not set in Module (the same way as you do in DRA reconciler). This should cover the case where user switches between DRA and DevicePlugin

yevgeny-shnaidman:
what about DeviceResource? Where will it be reconciled?

internal/controllers/device_plugin_reconciler.go
TomerNewman: its not backward compatible, old DS will not be found after an upgrade

internal/controllers/dra_reconciler.go
TomerNewman: I named it in puepose in order to be the same as we do it in DevicePlugin, but since you are right I renamed it both in DRA and in DevicePlugin controllers.

TomerNewman:
> what about DeviceResource? Where will it be reconciled?

DeviceClass do you mean?

internal/controllers/device_plugin_reconciler.go
yevgeny-shnaidman: why does it matter? The code will run CreateOrPatch with the same name and namespace, and the DS will be updated accordingly, and will be re-found in the next loop. Is there something catch that i am missing?

TomerNewman: The DS is created with `GenerateName`, not a fixed name. So `CreateOrPatch` would create a new DS instead of updating the old one. The old DS would remain orphaned — never found by the query, never garbage collected.

yevgeny-shnaidman:
> > what about DeviceResource? Where will it be reconciled?
> 
> DeviceClass do you mean?

yes :)

internal/controllers/device_plugin_reconciler.go
TomerNewman: updating - we will add the label now but quarry like it was implemented before, when 3.0 release will come, we will quarry with this new label.

internal/controllers/dra_reconciler.go
```diff
@@ -0,0 +1,326 @@
...
+) draReconcilerHelperAPI {
+	daemonSetHelper := newDRADaemonSetCreator(scheme)
+	return &draReconcilerHelper{
+		client:          client,
+		daemonSetHelper: daemonSetHelper,
+		nodeAPI:         nodeAPI,
+	}
+}
+
+func (drh *draReconcilerHelper) getModuleDRADaemonSets(ctx context.Context, name, namespace string) ([]appsv1.DaemonSet, error) {
```
yevgeny-shnaidman, line 142: i don't think we should see more then 1 DRA daemoset here. lets return the pointer to the DS, instead of slice

yevgeny-shnaidman: The same can implemented also for DevicePlugin

TomerNewman: Currently, DRA only creates a single DaemonSet per Module — you're right about that. However, the next phase of DRA work will add ordered upgrade support, mirroring DevicePlugin's existing mechanism. During ordered upgrades, two DRA DaemonSets will coexist (old version kept until `DesiredNumberScheduled == 0`, then garbage collected) — same as DevicePlugin does today via `getExistingDSFromVersion` + `garbageCollect`.

Keeping the slice now avoids refactoring it back when ordered upgrades are implemented. For the same reason, DevicePlugin should also stay as a slice — it already uses multiple DaemonSets during version transitions.

PR #1285, opened 2026-06-11
Add DeviceClass lifecycle management to DRA reconciler

TomerNewman:
### Summary

Extends the DRAReconciler to manage cluster-scoped Kubernetes DeviceClass resources declared in `spec.dra.deviceClasses`. Since DeviceClasses are cluster-scoped, ownership is tracked via labels (`ModuleNameLabel` + `ModuleNamespaceLabel`) rather than ownerReferences. The reconciler performs declarative convergence — creating missing, patching drifted, and deleting stale DeviceClasses each cycle.

### Changes

**`internal/constants/constants.go`**
- Added `ModuleNamespaceLabel` constant for cluster-scoped ownership tracking

**`internal/controllers/dra_reconciler.go`**
- Added `getModuleDeviceClasses` — lists DeviceClasses by ownership labels
- Added `handleDeviceClasses` — orchestrates create/patch/delete convergence
- Added `createOrPatchDeviceClass` — handles single DeviceClass create or patch
- Added `deleteDRAResources` — unified deletion of DaemonSets and DeviceClasses with `errors.Join` aggregation and `IsNotFound` tolerance
- Integrated DeviceClass handling into `Reconcile` loop (deletion, nil-DRA, and happy paths)

**`internal/controllers/dra_reconciler_test.go`**
- Updated existing `Reconcile` tests for new DeviceClass expectations
- Added `DRAReconciler_getModuleDeviceClasses` test block
- Added `DRAReconciler_handleDeviceClasses` test block (create, patch, delete, errors, no-op)
- Added `DRAReconciler_deleteDRAResources` test block (empty, success, partial failure, aggregated errors)

### Testing
- **Unit tests:** Comprehensive Ginkgo/Gomega tests covering all behavioral paths
- **Coverage:** All new functions at 100%; `Reconcile` at 96.2%
- **Checks:** `make generate`, `make fmt`, `make vet`, `make unit-test` all pass

### Acceptance Criteria

- [x] DRAReconciler lists DeviceClasses with `ModuleNameLabel` + `ModuleNamespaceLabel`
- [x] Missing DeviceClasses are created with correct `spec.selectors` and `spec.config`
- [x] Existing DeviceClasses are patched when spec differs from desired state
- [x] DeviceClasses present in cluster but absent from desired list are deleted
- [x] Ownership uses labels only (no ownerReferences — cluster-scoped)
- [x] Module deletion or `spec.dra` removal triggers cleanup of all matching DeviceClasses
- [x] Partial delete failures return aggregated error (controller-runtime requeues)
- [x] Unit tests cover create, patch, delete, drift correction, and cleanup paths

### DIFF
--- a/internal/controllers/dra_reconciler.go
+++ b/internal/controllers/dra_reconciler.go
@@ -27,6 +27,8 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -82,16 +84,18 @@ func (r *DRAReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Module) (
 		return res, fmt.Errorf("could not get DRA DaemonSets for module %s, namespace %s: %v", mod.Name, mod.Namespace, err)
 	}
 
+	existingDCs, err := r.reconHelperAPI.getModuleDeviceClasses(ctx, mod.Name, mod.Namespace)
+	if err != nil {
+		return res, fmt.Errorf("could not get DeviceClasses for module %s, namespace %s: %v", mod.Name, mod.Namespace, err)
+	}
+
 	if mod.GetDeletionTimestamp() != nil {
-		err = r.reconHelperAPI.deleteDRADaemonSets(ctx, existingDRADS)
-		return ctrl.Result{}, err
+		return ctrl.Result{}, r.reconHelperAPI.deleteDRAResources(ctx, mod.Name, mod.Namespace)
 	}
 
 	if mod.Spec.DRA == nil {
-		if len(existingDRADS) > 0 {
-			if err = r.reconHelperAPI.deleteDRADaemonSets(ctx, existingDRADS); err != nil {
-				return ctrl.Result{}, err
-			}
+		if err = r.reconHelperAPI.deleteDRAResources(ctx, mod.Name, mod.Namespace); err != nil {
+			return ctrl.Result{}, err
 		}
 		return ctrl.Result{}, r.reconHelperAPI.clearDRAStatus(ctx, mod)
 	}
@@ -99,6 +103,11 @@ func (r *DRAReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Module) (
 	err = r.reconHelperAPI.handleDRA(ctx, mod, existingDRADS)
 	if err != nil {
 		return res, fmt.Errorf("could not handle DRA: %v", err)
+	}
+
+	err = r.reconHelperAPI.handleDeviceClasses(ctx, mod, existingDCs)
+	if err != nil {
+		return res, fmt.Errorf("could not handle DeviceClasses: %v", err)
 	}
 
 	err = r.reconHelperAPI.moduleUpdateDRAStatus(ctx, mod, existingDRADS)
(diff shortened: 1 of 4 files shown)


internal/controllers/dra_reconciler.go
```diff
@@ -82,14 +84,18 @@ func (r *DRAReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Module) (
...
 	if mod.GetDeletionTimestamp() != nil {
-		err = r.reconHelperAPI.deleteDRADaemonSets(ctx, existingDRADS)
-		return ctrl.Result{}, err
+		return ctrl.Result{}, r.reconHelperAPI.deleteDRAResources(ctx, existingDRADS, existingDCs)
 	}
 
 	if mod.Spec.DRA == nil {
-		if len(existingDRADS) > 0 {
-			if err = r.reconHelperAPI.deleteDRADaemonSets(ctx, existingDRADS); err != nil {
+		if len(existingDRADS) > 0 || len(existingDCs) > 0 {
```
yevgeny-shnaidman: no need to check if they exist: deleteDRAResource support receiving zero-length slices

internal/controllers/dra_reconciler.go
```diff
@@ -233,6 +253,110 @@ func (drh *draReconcilerHelper) clearDRAStatus(ctx context.Context, mod *kmmv1be
...
+func (drh *draReconcilerHelper) createOrPatchDeviceClass(
+	ctx context.Context,
+	mod *kmmv1beta1.Module,
+	desired kmmv1beta1.DeviceClassSpec,
+	existingDC resourcev1.DeviceClass,
+	exists bool,
+) error {
+	logger := log.FromContext(ctx)
+
+	if !exists {
```
yevgeny-shnaidman: Why not use CreateOrPatch function here? That way you won't need to pass the exist and existingDC parameters, and is implified the code

internal/controllers/dra_reconciler.go
```diff
@@ -233,6 +253,110 @@ func (drh *draReconcilerHelper) clearDRAStatus(ctx context.Context, mod *kmmv1be
...
+	desiredByName := make(map[string]kmmv1beta1.DeviceClassSpec, len(mod.Spec.DRA.DeviceClasses))
+	for _, dc := range mod.Spec.DRA.DeviceClasses {
+		desiredByName[dc.Name] = dc
+	}
+
+	var errs []error
+
+	// Create missing or patch existing DeviceClasses to match the desired spec.
+	for _, desired := range mod.Spec.DRA.DeviceClasses {
+		existingDC, exists := existingByName[desired.Name]
```
yevgeny-shnaidman: you can remove the desired from existingDC, and then in 304 you can just go over DC that are left in the existingDC

internal/controllers/dra_reconciler.go
TomerNewman: Done, removed the len() guard.

internal/controllers/dra_reconciler.go
TomerNewman: Done, switched to controllerutil.CreateOrPatch.

internal/controllers/dra_reconciler.go
TomerNewman: Done, now deleting from existingByName during the loop and iterating over what remains.

ybettan:
/hold
While I am reviewing

internal/controllers/dra_reconciler.go
```diff
@@ -183,14 +194,21 @@ func (drh *draReconcilerHelper) handleDRA(ctx context.Context, mod *kmmv1beta1.M
...
-			return fmt.Errorf("failed to delete DRA DaemonSet %s/%s: %v", ds.Namespace, ds.Name, err)
+// deleteDRAResources deletes all DRA-owned DaemonSets and DeviceClasses.
+// It attempts every deletion and returns a joined error so the controller requeues on partial failure.
+func (drh *draReconcilerHelper) deleteDRAResources(ctx context.Context, daemonSets []appsv1.DaemonSet, deviceClasses []resourcev1.DeviceClass) error {
+	var errs []error
+	for _, ds := range daemonSets {
+		if err := drh.client.Delete(ctx, &ds); err != nil && !apierrors.IsNotFound(err) {
+			errs = append(errs, fmt.Errorf("failed to delete DRA DaemonSet %s/%s: %v", ds.Namespace, ds.Name, err))
+		}
+	}
```
ybettan: Can we use `drh.client.DeleteAllOf(...)` instead of looping?

internal/controllers/dra_reconciler.go
```diff
@@ -183,14 +194,21 @@ func (drh *draReconcilerHelper) handleDRA(ctx context.Context, mod *kmmv1beta1.M
...
+	for _, dc := range deviceClasses {
+		if err := drh.client.Delete(ctx, &dc); err != nil && !apierrors.IsNotFound(err) {
+			errs = append(errs, fmt.Errorf("failed to delete DeviceClass %s: %v", dc.Name, err))
```
ybettan: Can we use `drh.client.DeleteAllOf(...)` instead of looping?

internal/controllers/dra_reconciler.go
```diff
@@ -233,6 +251,75 @@ func (drh *draReconcilerHelper) clearDRAStatus(ctx context.Context, mod *kmmv1be
...
+		} else {
+			logger.Info("Reconciled DeviceClass", "name", desired.Name, "result", opRes)
+		}
+		delete(existingByName, desired.Name)
+	}
+
+	// Delete DeviceClasses that exist in the cluster but are no longer in the desired spec.
+	for name, dc := range existingByName {
+		if deleteErr := drh.client.Delete(ctx, &dc); deleteErr != nil && !apierrors.IsNotFound(deleteErr) {
+			errs = append(errs, fmt.Errorf("failed to delete DeviceClass %s: %v", name, deleteErr))
```
ybettan, line 325: Can we construct a list of DC to be deleted and delete all of them in a single API call?

TomerNewman: Unlike the full cleanup path (where DeleteAllOf works since all targets share the same labels), here we only want to delete the stale subset — not all module-owned DeviceClasses. DeleteAllOf with module labels would also remove the ones we just created/patched. Since there's no Kubernetes API to bulk-delete a specific set of named resources, individual Delete calls on the remaining map entries is the correct approach for selective deletion.

PR #1286, opened 2026-06-12
Fix hub webhook DRA version gating to skip cluster version check

TomerNewman:
The ManagedClusterModule webhook on the hub was incorrectly using the hub cluster's Kubernetes version to gate DRA usage. Since the hub cannot know the target managed cluster's version, this could wrongly reject valid specs or accept invalid ones.

Fix by passing an empty KubeVersion to the hub validator and skipping the version gate in validateDRA when the version is zero. Field format validation (driverName, deviceClasses) still runs unconditionally. The spoke cluster's own Module webhook enforces the version requirement.

Also moves DiscoverKubeVersion inside the enableModule branch so that webhook modes that don't need cluster version (namespace, preflight) no longer require ClusterVersion API access.

---

/cc @yevgeny-shnaidman 
/assign @yevgeny-shnaidman

### DIFF
--- a/cmd/webhook-server/main.go
+++ b/cmd/webhook-server/main.go
@@ -82,17 +82,17 @@ func main() {
 		cmd.FatalError(setupLogger, err, "unable to create manager")
 	}
 
-	kubeVersion, err := webhook.DiscoverKubeVersion(restCfg)
-	if err != nil {
-		cmd.FatalError(setupLogger, err, "unable to discover Kubernetes version")
-	}
-
-	setupLogger.Info("Detected Kubernetes version", "major", kubeVersion.Major, "minor", kubeVersion.Minor)
-
 	if enableModule {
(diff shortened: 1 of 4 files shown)


cmd/webhook-server/main.go
```diff
@@ -100,7 +100,7 @@ func main() {
 	if enableManagedClusterModule {
 		logger.Info("Enabling ManagedClusterModule webhook")
 
-		if err = hub.NewManagedClusterModuleValidator(logger, kubeVersion).SetupWebhookWithManager(mgr); err != nil {
+		if err = hub.NewManagedClusterModuleValidator(logger, webhook.KubeVersion{}).SetupWebhookWithManager(mgr); err != nil {
```
yevgeny-shnaidman: since KubeVersion is struct, wdyt about passing it as pointer? This way it will be much easier and clear to understand that we are passing nil pointer

PR #1288, opened 2026-06-17
Add DeviceClass watch to DRA reconciler

TomerNewman:
Watch DeviceClass resources so the controller reacts to external drift (e.g. manual deletion). Uses filter.HasLabel predicate to only process KMM-managed DeviceClasses. Also adds list, watch, and deletecollection RBAC verbs for the deviceclasses resource.

---

/cc @yevgeny-shnaidman 
/assign @yevgeny-shnaidman

### DIFF
--- a/internal/controllers/dra_reconciler.go
+++ b/internal/controllers/dra_reconciler.go
@@ -23,6 +23,7 @@ import (
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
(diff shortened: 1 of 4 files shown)


internal/controllers/dra_reconciler.go
```diff
@@ -64,16 +68,35 @@ func NewDRAReconciler(
...
+			handler.EnqueueRequestsFromMapFunc(deviceClassToModuleReconcileRequest),
+			builder.WithPredicates(filter.HasLabel(constants.ModuleNameLabel)),
+		).
 		Named(DRAReconcilerName).
 		Complete(
 			reconcile.AsReconciler[*kmmv1beta1.Module](r.client, r),
 		)
 }
 
+func deviceClassToModuleReconcileRequest(_ context.Context, obj client.Object) []reconcile.Request {
```
yevgeny-shnaidman: lets move it to filter also

PR #1289, opened 2026-06-17
DRA pod reconciler and node labeling

TomerNewman:
### Summary

Merge `DevicePluginPodReconciler` and the new DRA pod labeling logic into a single `PodNodeLabelReconciler`. The reconciler watches DaemonSet pods with a `ModuleNameLabel`, determines the role from the `DaemonSetRole` label, and manages the corresponding node label (`device-plugin-ready` or `dra-ready`) based on pod readiness. This provides a node-level readiness signal indicating which nodes have a functioning device-plugin or DRA driver for a Module.

### Changes

**Unified controller:**
- `PodNodeLabelReconciler` in `internal/controllers/pod_node_label_reconciler.go` — replaces `DevicePluginPodReconciler`
- Single registration in `cmd/manager/main.go`
- Determines label name based on `DaemonSetRole` label: `dra` → `dra-ready`, otherwise → `device-plugin-ready`

**Label utilities:**
- `GetDRANodeLabel` in `internal/utils/kmmlabels.go` — generates the `dra-ready` node label
- `GetDRANodeLabel` exposed in public API via `pkg/labels/labels.go`

**Constants:**
- `DRARoleLabelValue` and `DRAReadySuffix` in `internal/constants/constants.go`

### Testing
- **Unit tests:** Tests organized in two Context blocks (device-plugin pods, DRA pods) covering label add/remove, finalizer, replacement check, error paths, empty-nodeName, and DRA exclusion
- **Live cluster validation:** Deployed to minikube — verified labeling on ready, unlabeling on delete, switching between roles, and independent operation of both roles

### Acceptance Criteria

- [x] AC-1: `PodNodeLabelReconciler` watches pods and sets `dra-ready` or `device-plugin-ready` label on node when ready
- [x] AC-2: Label removed when pod not ready or deleting
- [x] AC-3: Pod event filter requires `ModuleNameLabel` + DaemonSet owner
- [x] AC-4: Reconciler uses `NodeLabelerFinalizer`
- [x] AC-5: Device-plugin replacement check excludes DRA pods
- [x] AC-6: Single `PodNodeLabelReconciler` registered in `cmd/manager/main.go`
- [x] AC-7: Unit tests cover both roles — label add/remove, predicate filtering, replacement check, error paths


cmd/manager/main.go
```diff
@@ -170,6 +170,10 @@ func main() {
 		cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.DevicePluginPodReconcilerName)
 	}
 
+	if err = controllers.NewDRAPodReconciler(client).SetupWithManager(mgr); err != nil {
```
ybettan: How do we handle DRA objects in hub-spoke?

ybettan:
> providing downstream consumers a signal (dra-ready node label) indicating which nodes have a functioning DRA driver for a Module.

Why is d/s related here?

internal/controllers/dra_pod_reconciler.go
```diff
@@ -0,0 +1,146 @@
...
+	logger := ctrl.LoggerFrom(ctx)
+
+	nodeName := pod.Spec.NodeName
+
+	moduleName, ok := pod.Labels[constants.ModuleNameLabel]
+	if !ok {
+		return ctrl.Result{}, fmt.Errorf("pod %s/%s has no %q label", pod.Namespace, pod.Name, constants.ModuleNameLabel)
+	}
+
+	labelName := utils.GetDRANodeLabel(pod.Namespace, moduleName)
```
ybettan: `GetDRANodeLabel` is a it confusing here since we are constructing/building the label pattern, not getting it from any object.

Can we rename this function while still staying consistent with how we construct version-ready/ready labels?

ybettan: Also, the `labelName` receiver have a too generic name IMO.

internal/controllers/pod_node_label_reconciler.go
```diff
@@ -0,0 +1,146 @@
...
+	labelName := utils.GetDRANodeLabel(pod.Namespace, moduleName)
+
+	logger = logger.WithValues(
+		"node name", nodeName,
+		"module name", moduleName,
+		"label name", labelName,
+	)
+
+	if !podutils.IsPodReady(pod) || !pod.DeletionTimestamp.IsZero() {
+		if nodeName != "" {
```
ybettan, line 66: I would add a funcion here or anything else to make it clearer we are targeting pods that weren't scheduled yet.

internal/controllers/dra_pod_reconciler.go
```diff
@@ -0,0 +1,146 @@
...
+		"node name", nodeName,
+		"module name", moduleName,
+		"label name", labelName,
+	)
+
+	if !podutils.IsPodReady(pod) || !pod.DeletionTimestamp.IsZero() {
+		if nodeName != "" {
+			logger.Info("Unlabeling node")
+
+			labelSelector := client.MatchingLabels{
```
ybettan:
```suggestion
			// Making sure there is no other DRA pod already running
			labelSelector := client.MatchingLabels{
```

cmd/manager/main.go
TomerNewman: Hub-and-spoke support for DRA objects will be addressed in a later PR. This reconciler only operates at the spoke level (watching DRA pods and labeling nodes) — same as `DevicePluginPodReconciler`, which also doesn't handle h&s itself.

internal/controllers/dra_pod_reconciler.go
TomerNewman: This follows the same naming convention as the existing `GetDevicePluginNodeLabel`, `GetKernelModuleReadyNodeLabel`, `GetDevicePluginTargetNodeLabel`, etc. in `kmmlabels.go`. Renaming just this one would be inconsistent

TomerNewman: `labelName` is the same variable name used in `DevicePluginPodReconciler` for the same concept. Changing it only here would be inconsistent.

internal/controllers/pod_node_label_reconciler.go
TomerNewman: Done — added the same comment as in `DevicePluginPodReconciler`.

internal/controllers/dra_pod_reconciler.go
TomerNewman: Done.

TomerNewman:
> > providing downstream consumers a signal (dra-ready node label) indicating which nodes have a functioning DRA driver for a Module.
> 
> Why is d/s related here?

"Downstream consumers" here doesn't refer to hub-spoke — it means anything that reads node labels to make scheduling or readiness decisions (other controllers, custom schedulers, or user workloads with node selectors). Poor wording on my part — I've updated the PR description to say "provides a node-level readiness signal" instead.

internal/constants/constants.go
```diff
@@ -13,6 +13,7 @@ const (
 
 	WorkerPodVersionLabelPrefix    = "beta.kmm.node.kubernetes.io/version-worker-pod"
 	DevicePluginVersionLabelPrefix = "beta.kmm.node.kubernetes.io/version-device-plugin"
+	DRAVersionLabelPrefix          = "beta.kmm.node.kubernetes.io/version-dra"
```
yevgeny-shnaidman: I don't think this should be in this PR. This is related to ordered uprade.

yevgeny-shnaidman:
General question: dra_pod_reconciler.go is a copy of device_plugin_pod_reconciler.go, except for the label it uses. Since device plugin and dra are mutually exclusive, can't we just use the device-pod-reconcile (and rename it) for both device plugin and dra

internal/controllers/pod_node_label_reconciler.go
```diff
@@ -38,18 +35,26 @@ func (dppr *DevicePluginPodReconciler) Reconcile(ctx context.Context, pod *v1.Po
 		return ctrl.Result{}, fmt.Errorf("pod %s/%s has no %q label", pod.Namespace, pod.Name, constants.ModuleNameLabel)
 	}
 
-	labelName := utils.GetDevicePluginNodeLabel(pod.Namespace, moduleName)
+	isDRA := pod.Labels[constants.DaemonSetRole] == constants.DRARoleLabelValue
+
+	var labelName string
```
yevgeny-shnaidman:
```suggestion
	labelName := utils.GetDevicePluginNodeLabel(pod.Namespace, moduleName)
	if isDRA {
	    labelName = utils.GetDRANodeLabel(pod.Namespace, moduleName)
	}
```

internal/controllers/pod_node_label_reconciler.go
```diff
@@ -59,23 +64,31 @@ func (dppr *DevicePluginPodReconciler) Reconcile(ctx context.Context, pod *v1.Po
 		if nodeName != "" {
 			logger.Info("Unlabeling node")
 
-			// Make sure we don't already have a new running pod before unlabeling the node
+			// Making sure there is no other pod of the same role already running
 			labelSelector := client.MatchingLabels{constants.ModuleNameLabel: moduleName}
+			if isDRA {
```
yevgeny-shnaidman: have not we also added a role label for devicePlugin?

internal/controllers/pod_node_label_reconciler.go
```diff
@@ -59,23 +64,31 @@ func (dppr *DevicePluginPodReconciler) Reconcile(ctx context.Context, pod *v1.Po
...
+
+			var podsList v1.PodList
+			if err := r.client.List(ctx, &podsList, labelSelector, fieldSelector); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to get list of pods for module %s on node %s: %v", moduleName, nodeName, err)
 			}
+
 			var foundRunningPod bool
-			for _, p := range modulePodsList.Items {
+			for _, p := range podsList.Items {
+				if !isDRA && p.Labels[constants.DaemonSetRole] == constants.DRARoleLabelValue {
```
yevgeny-shnaidman: don't we need to make the same check for DRA pod?

internal/controllers/pod_node_label_reconciler.go
```diff
@@ -139,50 +151,10 @@ func (dppr *DevicePluginPodReconciler) SetupWithManager(mgr ctrl.Manager) error
...
+		Named(PodNodeLabelReconcilerName).
 		For(&v1.Pod{}).
 		WithEventFilter(p).
 		Complete(
-			reconcile.AsReconciler[*v1.Pod](dppr.client, dppr),
+			reconcile.AsReconciler[*v1.Pod](r.client, r),
 		)
 }
-
-func (dppr *DevicePluginPodReconciler) addLabel(ctx context.Context, nodeName string, labelName string) error {
```
yevgeny-shnaidman, line 150: why not keep this function here? this is the only code that is using it

internal/controllers/pod_node_label_reconciler.go
TomerNewman: Existing device-plugin pods from older KMM versions may not have the role label. Adding it to the selector would miss those pods during the replacement check and incorrectly unlabel the node. That's why we only narrow the selector for DRA (which is new — all DRA pods will always have the label) and use a post-filter skip for device-plugin instead.

internal/controllers/pod_node_label_reconciler.go
TomerNewman: No — when `isDRA` is true, lines 69-71 already add `DaemonSetRole=dra` to the label selector, so the List only returns DRA pods. Device-plugin pods won't appear in the results. This skip (line 81) is only needed for the device-plugin path, where the query doesn't filter by role (for backwards compatibility), so DRA pods could appear in the list and need to be excluded.

internal/constants/constants.go
TomerNewman: Done — removed.

internal/controllers/pod_node_label_reconciler.go
TomerNewman: Done.

internal/controllers/pod_node_label_reconciler.go
TomerNewman: Done — moved back as private methods on the reconciler.

internal/controllers/dra_pod_reconciler.go
ybettan: > labelName is the same variable name used in DevicePluginPodReconciler for the same concept. Changing it only here would be inconsistent.

Seems like this controller doesn't exist anymore. So can we change it only in one place in the converged controller?

ybettan:
I am happy with this PR. @yevgeny-shnaidman can lgtm when he is happy too.

internal/controllers/pod_node_label_reconciler.go
```diff
@@ -59,23 +64,30 @@ func (dppr *DevicePluginPodReconciler) Reconcile(ctx context.Context, pod *v1.Po
 		if nodeName != "" {
 			logger.Info("Unlabeling node")
 
-			// Make sure we don't already have a new running pod before unlabeling the node
+			// Making sure there is no other pod of the same role already running
 			labelSelector := client.MatchingLabels{constants.ModuleNameLabel: moduleName}
+			if isDRA {
```
yevgeny-shnaidman: we already have a code in line 42-43 that check if this DRA or not. Lets define a variable there that will hold either constants.DRARoleLabelValue or  constants.DevicePluginRoleLabelValue and here we just set it , without needing to check again if it is DRA

PR #1291, opened 2026-06-22
Add DRA support to ordered-upgrade state machine

TomerNewman:
## Summary

- Extend the `NodeLabelModuleVersionReconciler` to handle DRA modules through the same ordered-upgrade label action table used by device plugin modules, adding a new `version-dra` node label prefix
- Parameterize `getLabelAndAction` with a role-specific label function so the existing state machine works for both device plugin and DRA without duplication
- Make the DRA reconciler version-aware: include `version-dra` in DaemonSet node selectors, create per-version DaemonSets (to avoid immutable selector errors on upgrade), and garbage collect old-version DaemonSets

## Changes

- `internal/constants/constants.go` — new `DRAVersionLabelPrefix` constant
- `internal/utils/kmmlabels.go` — new `GetDRAVersionLabelName`, `IsDRAVersionLabel`; refactored `GetNodesVersionLabels` to use `IsVersionLabel`
- `internal/controllers/node_label_module_version_reconciler.go` — added `isDRA`/`draVersionLabel` fields, `setModulesDRAStatus` (with graceful fallback when Module CR is deleted), `getDRAPods` via shared `getDaemonSetPods` helper, `resolveLabel` for role dispatch
- `internal/controllers/dra_reconciler.go` — version label in DaemonSet labels + node selector, `getExistingDRADSFromVersion`, `handleDRA` version-aware creation, `garbageCollectDRADaemonSets`
- Tests updated across all changed files with full DRA path coverage

## Test plan

- [x] Unit tests: 385 controller specs, 44 utils specs — all passing
- [x] `go vet` and `go build ./...` clean
- [x] E2E on Minikube: deployed v1 DRA module (kmm_ci_a), upgraded to v2 (kmm_ci_b) via version-module label change, verified full label transition sequence, DRA DaemonSet lifecycle (new DS created, old GC'd), and kernel module swap via lsmod — zero controller errors

### DIFF
--- a/internal/controllers/node_label_module_version_reconciler.go
+++ b/internal/controllers/node_label_module_version_reconciler.go
@@ -9,6 +9,8 @@ import (
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -16,13 +18,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-// this struct contains all the version labels related to a specific Module
 type modulesVersionLabels struct {
-	name                     string
-	namespace                string
-	moduleVersionLabel       string
-	workerPodVersionLabel    string
-	devicePluginVersionLabel string
+	name                       string
+	namespace                  string
+	moduleVersionLabel         string
+	workerPodVersionLabel      string
+	schedulePluginVersionLabel string
 }
 
 const (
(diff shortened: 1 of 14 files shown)


internal/controllers/node_label_module_version_reconciler.go
```diff
@@ -119,24 +136,46 @@ func (nlmvha *nodeLabelModuleVersionHelper) getLabelsPerModules(ctx context.Cont
 }
 
 func (nlmvha *nodeLabelModuleVersionHelper) getDevicePluginPods(ctx context.Context, nodeName string) ([]v1.Pod, error) {
-	var kmmPodsList v1.PodList
-	fieldSelector := client.MatchingFields{"spec.nodeName": nodeName}
-	labelSelector := client.HasLabels{constants.ModuleNameLabel}
-	err := nlmvha.client.List(ctx, &kmmPodsList, labelSelector, fieldSelector)
+	return nlmvha.getDaemonSetPods(ctx, nodeName, client.HasLabels{constants.ModuleNameLabel})
```
yevgeny-shnaidman: i think that we will have a role label in DevicePlugin too

yevgeny-shnaidman:
Suggestion:
from [node_label_module_version reconciler POV, it does not matter what is running: DRA or DevicePLugin.
Since they are mutually exlusive, i suggest we unify the labels:
1. one label for both: beta.kmm.node.kubernetes.io/version-dra-device-plugin (or something else)
2. in the reconciler, only gettting Pods needs to be changed (it should now check DevicePlugin or DRA pods).
3. state machine should use the new label and should change the labelActionKey devicePLugin to a new name

PR #1292, opened 2026-06-22
Restrict device plugin hostPath volumes to allowed prefixes

TomerNewman:
## Summary

Add webhook validation that rejects hostPath volumes in `spec.devicePlugin.volumes` unless the path resolves under `/dev`, `/sys`, `/var` or `/opt`. This prevents device plugins from mounting arbitrary host directories such as `/` or `/etc`.

## Details

- New `validateDevicePluginVolumes` function in the webhook with `filepath.Clean` to prevent path traversal bypasses (e.g. `/dev/../etc`)
- Called from `validateModule` on both create and update
- Non-hostPath volumes (ConfigMap, Secret, EmptyDir, etc.) are allowed through since they don't expose host filesystem

## Test plan

- [x] Unit tests added (13 cases covering nil, empty, allowed paths, rejected paths, prefix tricks, path traversal)
- [x] `make fmt` passes
- [x] `make lint` passes
- [x] `make unit-test` passes

---

/assign @ybettan @yevgeny-shnaidman

### DIFF
--- a/internal/webhook/module.go
+++ b/internal/webhook/module.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -162,6 +163,10 @@ func validateModule(mod *kmmv1beta1.Module, kubeVersion *KubeVersion) (admission
 		return nil, fmt.Errorf("failed to validate DRA: %v", err)
 	}
 
+	if err := validateDevicePluginVolumes(mod.Spec.DevicePlugin); err != nil {
+		return nil, fmt.Errorf("failed to validate device plugin volumes: %v", err)
+	}
+
 	if mod.Spec.ModuleLoader == nil {
 		// If ModuleLoader is nil, there is no need to validate related fields
 		return nil, nil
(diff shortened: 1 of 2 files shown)


internal/webhook/module.go
```diff
@@ -222,6 +227,35 @@ func validateDRA(mod *kmmv1beta1.Module, kubeVersion *KubeVersion) error {
 	return nil
 }
 
+var allowedHostPathPrefixes = []string{"/dev", "/sys", "/var", "/opt"}
+
+func isAllowedHostPath(hostPath string) bool {
+	p := filepath.Clean(hostPath)
+	for _, prefix := range allowedHostPathPrefixes {
+		if p == prefix || strings.HasPrefix(p, prefix+"/") {
```
yevgeny-shnaidman, line 235: why do we need `prefix + "/"`? What happens if the volume is "/var"?

yevgeny-shnaidman:
please update the title and the commit message

TomerNewman:
> please update the title and the commit message

Done

internal/webhook/module.go
TomerNewman: The check has two parts:

```go
if cleanPath == prefix || strings.HasPrefix(cleanPath, prefix+"/") {
```

- `cleanPath == prefix` handles the exact match — so `/var` by itself is allowed.
- `strings.HasPrefix(cleanPath, prefix+"/")` handles subdirectories like `/var/lib/kubelet`.

The `+ "/"` is specifically to prevent false positives: without it, `strings.HasPrefix("/variable", "/var")` would return `true`, incorrectly allowing a path like `/variable`. The trailing slash ensures we only match actual subdirectories.

PR #1297, opened 2026-06-29
Adding dra-kubelet communication volumes

yevgeny-shnaidman:
This commit add mapping of a number of volumes and envs that are necessary for DRA driver <--> kubelet communication

### DIFF
--- a/internal/controllers/dra_reconciler.go
+++ b/internal/controllers/dra_reconciler.go
@@ -32,6 +32,7 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -48,6 +49,10 @@ const (
 	kubeletPluginsPath               = "/var/lib/kubelet/plugins/"
 	kubeletPluginsRegistryVolumeName = "kubelet-plugins-registry"
 	kubeletPluginsRegistryPath       = "/var/lib/kubelet/plugins_registry/"
+	cdiVolumeName                    = "cdi"
+	cdiPath                          = "/var/run/cdi"
+
+	draHealthcheckPort = 51515
 )
 
 type DRAReconciler struct {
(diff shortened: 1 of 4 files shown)


internal/controllers/dra_reconciler.go
```diff
@@ -438,9 +443,76 @@ func (dsci *draDaemonSetCreatorImpl) setDRAAsDesired(
...
+			Name:  "KUBELET_REGISTRAR_DIRECTORY_PATH",
+			Value: kubeletPluginsRegistryPath,
+		},
+		{
+			Name:  "KUBELET_PLUGINS_DIRECTORY_PATH",
+			Value: kubeletPluginsPath,
+		},
+		{
+			Name:  "HEALTHCHECK_PORT",
+			Value: fmt.Sprintf("%d", draHealthcheckPort),
```
ybettan, line 489: What happens if 2 DRA pods lands on the same node and both bind the 51515 port? One of them will probably fail isn't it?

internal/controllers/dra_reconciler.go
```diff
@@ -438,9 +443,76 @@ func (dsci *draDaemonSetCreatorImpl) setDRAAsDesired(
...
+				Path: cdiPath,
+				Type: &hostPathDirOrCreate,
+			},
+		},
+	}
+
 	containerVolumeMounts := []v1.VolumeMount{
 		{Name: kubeletPluginsVolumeName, MountPath: kubeletPluginsPath},
 		{Name: kubeletPluginsRegistryVolumeName, MountPath: kubeletPluginsRegistryPath},
+		{Name: cdiVolumeName, MountPath: cdiPath},
```
ybettan, line 459: Here the CDI volume mount override the user mounts unlike the ENV variables that allows the user to override the pre-set variables. Is this by design? Shouldn't we have a consistent behavior?

internal/controllers/dra_reconciler.go
```diff
@@ -471,12 +543,13 @@ func (dsci *draDaemonSetCreatorImpl) setDRAAsDesired(
 			},
 			Spec: v1.PodSpec{
 				InitContainers:               generatePodContainerSpec(mod.Spec.DRA.InitContainer, "dra-init", nil),
-				Containers:                   generatePodContainerSpec(&mod.Spec.DRA.Container, "dra", containerVolumeMounts),
+				Containers:                   []v1.Container{draContainer},
```
ybettan: Why don't we keep using `generatePodContainerSpec` and append additional fields? This mean that what ever is added to the output of `generatePodContainerSpec` will not be inherited by this container.

internal/controllers/dra_reconciler.go
yevgeny-shnaidman: they don't override user mount. See line 503

internal/controllers/dra_reconciler.go
yevgeny-shnaidman: dra pods are deployed by daemonset, so 2 pods cannot land on the same node

internal/controllers/dra_reconciler.go
yevgeny-shnaidman: fixed

internal/controllers/dra_reconciler.go
yevgeny-shnaidman: changed the order, so that user's volume's definition will take precedence

PR #1298, opened 2026-06-29
Fix DRA nodeSelector for in-tree driver mode

abyrne55:
When `spec.moduleLoader` is nil (in-tree driver mode), the DRA reconciler's `nodeSelector` includes the KMM ready label, which is never set because no module loading occurs. This prevents the DRA DaemonSet from scheduling on any node.

Fall back to `spec.selector` instead, matching the existing behavior in the device plugin reconciler.

**Suggested test plan**
- [ ] Verify the DRA DaemonSet uses `spec.selector` as its nodeSelector
- [ ] Deploy KMM with an in-tree Module (`moduleLoader: nil`) and DRA enabled
- [ ] Verify the DRA DaemonSet pods schedule correctly on matching nodes

### DIFF
--- a/internal/controllers/dra_reconciler.go
+++ b/internal/controllers/dra_reconciler.go
@@ -512,10 +512,14 @@ func (dsci *draDaemonSetCreatorImpl) setDRAAsDesired(
 		utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name): "",
 	}
 
-	if mod.Spec.ModuleLoader != nil && mod.Spec.ModuleLoader.Container.Version != "" {
-		versionLabel := utils.GetSchedulePodVersionLabelName(mod.Namespace, mod.Name)
-		standardLabels[versionLabel] = mod.Spec.ModuleLoader.Container.Version
-		nodeSelector[versionLabel] = mod.Spec.ModuleLoader.Container.Version
+	if mod.Spec.ModuleLoader != nil {
+		if mod.Spec.ModuleLoader.Container.Version != "" {
+			versionLabel := utils.GetSchedulePodVersionLabelName(mod.Namespace, mod.Name)
+			standardLabels[versionLabel] = mod.Spec.ModuleLoader.Container.Version
+			nodeSelector[versionLabel] = mod.Spec.ModuleLoader.Container.Version
+		}
+	} else {
+		nodeSelector = mod.Spec.Selector
 	}
 
 	ds.SetLabels(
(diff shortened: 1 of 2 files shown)


internal/controllers/dra_reconciler.go
```diff
@@ -516,6 +516,8 @@ func (dsci *draDaemonSetCreatorImpl) setDRAAsDesired(
 		versionLabel := utils.GetSchedulePluginVersionLabelName(mod.Namespace, mod.Name)
 		standardLabels[versionLabel] = mod.Spec.ModuleLoader.Container.Version
 		nodeSelector[versionLabel] = mod.Spec.ModuleLoader.Container.Version
+	} else if mod.Spec.ModuleLoader == nil {
```
yevgeny-shnaidman: @abyrne55 great find, we missed that use-case!
one note:
in case **mod.Spec.ModuleLoader != nil** but **mod.Spec.ModuleLoader.Container.Version == ""** , your fix will set the **nodeSelector = mod.Spec.Selector**, while it should still be **utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name): ""**

abyrne55: Thanks for taking a look @yevgeny-shnaidman! I'm a bit confused: IIUC, when `ModuleLoader != nil && Version == ""`, the first condition (`!= nil && Version != ""`) is false, and the `else if (ModuleLoader == nil)` is also false since `ModuleLoader` is non-nil. As a result, neither branch executes, so the nodeSelector retains the default ready label `{GetKernelModuleReadyNodeLabel: ""}`. Am I missing something?

yevgeny-shnaidman: Sorry, my bad! blacked out on "if else",  thought it was "esle". You are correct. Adding a suggestion to  make it a little bit more clear

internal/controllers/dra_reconciler.go
yevgeny-shnaidman:
```suggestion
	if mod.Spec.ModuleLoader != nil {
	    if mod.Spec.ModuleLoader.Container.Version != "" {
	        versionLabel := utils.GetSchedulePluginVersionLabelName(mod.Namespace, mod.Name)
		    standardLabels[versionLabel] = mod.Spec.ModuleLoader.Container.Version
		    nodeSelector[versionLabel] = mod.Spec.ModuleLoader.Container.Version
		}
    } else {
         nodeSelector = mod.Spec.Selector
    }
```

yevgeny-shnaidman:
@abyrne55 you also need to fix the unit-tests. You can run them locally using "make unit-test"

PR #1301, opened 2026-07-01
Fix collision between sofdep unloading and hard symbol dependency

yevgeny-shnaidman:
scenario:
1.KMM Modules needs to load module_a, module_b, module_c
2. module_a has a hard symbol dependency on module_b.
3. modulesLoadingOrder defined as: [module_a, module_b, module_c]

There is no problems when loading. When unloading using softdep, the following happens
1. module_a is unloaded first
2. module_b is unloaded automatically, due to hard symbol dependency
3. modprobe looks at softdep, sees that module_b is unloaded, and does not continue to module_c

This commit changes the Unload flow:
1. no softdep is created.
2. modprobe -r ocmmand is provided with the list of modules defined in the modulesLoadingOrder

This way it will go over every module in the command and try to unload it

### DIFF
--- a/internal/worker/worker.go
+++ b/internal/worker/worker.go
@@ -146,14 +146,19 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, f
 		if cfg.Modprobe.Args != nil {
 			args = append(args, cfg.Modprobe.Args.Unload...)
 		}
-
-		args = append(args, moduleName)
+		if len(cfg.Modprobe.ModulesLoadingOrder) > 0 {
+			w.logger.Info("Preparing to unload multiple modules", "modules", cfg.Modprobe.ModulesLoadingOrder)
+			args = append(args, cfg.Modprobe.ModulesLoadingOrder...)
+		} else {
+			w.logger.Info("Preparing to unload single module", "moduleName", moduleName)
+			args = append(args, moduleName)
+		}
 	}
 
-	w.logger.Info("Unloading module", "name", moduleName)
+	w.logger.Info("Starting unloading", "args", args)
 
 	if err := w.mr.Run(ctx, args...); err != nil {
-		return fmt.Errorf("could not unload module %s: %v", moduleName, err)
+		return fmt.Errorf("failed to unload module %q: %v", args, err)
 	}
 
 	//remove firmware files only (no directories)
--- a/internal/pod/workerpodmanager.go
+++ b/internal/pod/workerpodmanager.go
@@ -166,7 +166,7 @@ func (wpmi *workerPodManagerImpl) LoaderPodTemplate(ctx context.Context, nmc cli
 		return nil, fmt.Errorf("could not create the base Pod: %v", err)
 	}
 
-	if nms.Config.Modprobe.ModulesLoadingOrder != nil {
+	if len(nms.Config.Modprobe.ModulesLoadingOrder) > 0 {
 		if err = setWorkerSofdepConfig(pod, nms.Config.Modprobe.ModulesLoadingOrder); err != nil {
 			return nil, fmt.Errorf("could not set software dependency for mulitple modules: %v", err)
 		}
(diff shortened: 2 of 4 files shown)


TomerNewman (commented):
> This way it will go over every module in the command and try to unload it

but if modprobe will do now `modprobe -r a b c` and when it tries to remove a, b will be removed too, no? so when modprobe will try to remove b it is already gone and because of that fail.

internal/worker/worker.go
```diff
@@ -146,8 +146,13 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, f
...
+		if len(cfg.Modprobe.ModulesLoadingOrder) > 0 {
+			w.logger.Info("Unloading multiple modules", "modules", cfg.Modprobe.ModulesLoadingOrder)
+			args = append(args, cfg.Modprobe.ModulesLoadingOrder...)
+		} else {
+			w.logger.Info("Unloading single module", "moduleName", moduleName)
+			args = append(args, moduleName)
+		}
 	}
 
 	w.logger.Info("Unloading module", "name", moduleName)
```
TomerNewman: if we already remove multiple modules, the logs will still show that we remove only a single module

yevgeny-shnaidman:
> > This way it will go over every module in the command and try to unload it
> 
> but if modprobe will do now `modprobe -r a b c` and when it tries to remove a, b will be removed too, no? so when modprobe will try to remove b it is already gone and because of that fail.

in this case modprobe just goes over every module it received in the command line and tries to remove, no matter if it is present or not. In case of softdep the flow is different

internal/worker/worker.go
yevgeny-shnaidman: fixed

internal/worker/worker.go
```diff
@@ -146,11 +146,16 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, f
...
+			w.logger.Info("Preparing to unload single module", "moduleName", moduleName)
+			args = append(args, moduleName)
+		}
 	}
 
-	w.logger.Info("Unloading module", "name", moduleName)
+	w.logger.Info("Starting unloading", "args", args)
 
 	if err := w.mr.Run(ctx, args...); err != nil {
 		return fmt.Errorf("could not unload module %s: %v", moduleName, err)
```
TomerNewman: I would also change the log message here since we can unload multiple modules

internal/worker/worker.go
```diff
@@ -146,11 +146,16 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, f
 		if cfg.Modprobe.Args != nil {
 			args = append(args, cfg.Modprobe.Args.Unload...)
 		}
-
-		args = append(args, moduleName)
+		if len(cfg.Modprobe.ModulesLoadingOrder) > 0 {
```
TomerNewman, line 149: in [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/1301/changes#diff-3d5ae4251fdd0cba3654cdd507d2a1a1419c888cb42e1f3d28b9fe08d2b26bacL169)
we check with ModulesLoadingOrder != nil, in order to be consistent with the code I would change to 
`cfg.Modprobe.ModulesLoadingOrder != nil`

internal/worker/worker.go
yevgeny-shnaidman: fixed

TomerNewman:
/unhold

I lgtmed accidently

internal/worker/worker.go
yevgeny-shnaidman: using len is safer

TomerNewman: so lets change it [here](https://github.com/kubernetes-sigs/kernel-module-management/pull/1301/changes#diff-3d5ae4251fdd0cba3654cdd507d2a1a1419c888cb42e1f3d28b9fe08d2b26bacL169) to be consistent

yevgeny-shnaidman: fixed

PR #1304, opened 2026-07-05
Add DRA e2e CI test

TomerNewman:
Add an end-to-end test that validates the full DRA lifecycle: DaemonSet creation, DeviceClass management, node labeling, status reporting, and cleanup on Module deletion.

---


yshnaidman:
General note:
I think that the flow needs to verify that the DRA isd actually working,,meaning:
apply a pod that will get scheduled based on DRA. Can that be done?

TomerNewman:
@yevgeny-shnaidman you can re-review

PR #1309, opened 2026-07-13
Adding helm deployment

yevgeny-shnaidman:
1. adding deployment directory that contains the all the helm charts
2. adding helm deploy/undeploy commands to the Makefile
3. adjusting skipper image to include helm utility

### DIFF
--- a/Dockerfile.kmm-operator-build
+++ b/Dockerfile.kmm-operator-build
@@ -11,3 +11,8 @@ RUN go install go.uber.org/mock/mockgen@v0.3.0
 RUN curl -L "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" -o /tmp/kubectl
 RUN install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl
 COPY --from=quay.io/openshift/origin-cli:latest /usr/bin/oc /usr/bin
+RUN curl -fsSL https://get.helm.sh/helm-v3.17.3-linux-amd64.tar.gz -o /tmp/helm.tar.gz \
+    && echo "ee88b3c851ae6466a3de507f7be73fe94d54cbf2987cbaa3d1a3832ea331f2cd  /tmp/helm.tar.gz" | sha256sum -c - \
+    && tar -xzf /tmp/helm.tar.gz -C /tmp \
+    && install -o root -g root -m 0755 /tmp/linux-amd64/helm /usr/local/bin/helm \
+    && rm -rf /tmp/helm.tar.gz /tmp/linux-amd64
--- a/Makefile
+++ b/Makefile
@@ -95,6 +95,9 @@ all: generate manager manager-hub manifests
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
+HELM_CHART_KMM ?= deployment/helm/kmm
+HELM_CHART_KMM_HUB ?= deployment/helm/kmm-hub
+
 ##@ Development
 
 .PHONY: manifests
@@ -107,6 +110,16 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 	$(CONTROLLER_GEN) crd paths="./api-hub/..." output:crd:artifacts:config=config/crd-hub/bases
 	$(CONTROLLER_GEN) webhook paths="./internal/webhook/hub" output:webhook:artifacts:config=config/webhook-hub
 	$(CONTROLLER_GEN) rbac:roleName=manager-role paths="./internal/controllers/hub" output:rbac:artifacts:config=config/rbac-hub
+	# Sync CRDs to Helm charts
+	@mkdir -p $(HELM_CHART_KMM)/crds
+	@cp config/crd/bases/kmm.sigs.x-k8s.io_modules.yaml $(HELM_CHART_KMM)/crds/
+	@cp config/crd/bases/kmm.sigs.x-k8s.io_nodemodulesconfigs.yaml $(HELM_CHART_KMM)/crds/
+	@cp config/crd/bases/kmm.sigs.x-k8s.io_moduleimagesconfigs.yaml $(HELM_CHART_KMM)/crds/
+	@cp config/crd/bases/kmm.sigs.x-k8s.io_modulebuildsignconfigs.yaml $(HELM_CHART_KMM)/crds/
+	@mkdir -p $(HELM_CHART_KMM_HUB)/crds
+	@cp config/crd-hub/bases/hub.kmm.sigs.x-k8s.io_managedclustermodules.yaml $(HELM_CHART_KMM_HUB)/crds/
+	@cp config/crd-hub/bases/kmm.sigs.x-k8s.io_modulebuildsignconfigs.yaml $(HELM_CHART_KMM_HUB)/crds/
+	@cp config/crd-hub/bases/kmm.sigs.x-k8s.io_moduleimagesconfigs.yaml $(HELM_CHART_KMM_HUB)/crds/
 
 .PHONY: generate
 generate: controller-gen mockgen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
@@ -359,6 +372,43 @@ workerimage-build: ## Build docker image for worker app.
 
 operatorhub-release:
 	IMG=$(IMG) HUB_IMG=$(HUB_IMG) WORKER_IMG=$(WORKER_IMG) SIGNER_IMG=$(SIGNER_IMG) VERSION=$(VERSION) ./hack/release-operatorhub
+
+##@ Helm Deployment
+
+HELM_RELEASE_KMM ?= kmm
+HELM_RELEASE_KMM_HUB ?= kmm-hub
+HELM_NAMESPACE ?= kmm-operator-system
+
+.PHONY: helm-deploy
+helm-deploy: deploy-cert-manager manifests ## Deploy KMM controller using Helm to the K8s cluster specified in ~/.kube/config.
+	helm upgrade --install $(HELM_RELEASE_KMM) $(HELM_CHART_KMM) \
+		--namespace $(HELM_NAMESPACE) --create-namespace \
+		--set controller.image=$(IMG) \
+		--set controller.workerImage=$(WORKER_IMG) \
+		--set controller.signerImage=$(SIGNER_IMG) \
+		--set webhookServer.image=$(WEBHOOK_IMG)
+
+.PHONY: helm-deploy-hub
+helm-deploy-hub: deploy-cert-manager manifests ## Deploy KMM-Hub controller using Helm to the K8s cluster specified in ~/.kube/config.
+	helm upgrade --install $(HELM_RELEASE_KMM_HUB) $(HELM_CHART_KMM_HUB) \
+		--namespace $(HELM_NAMESPACE) --create-namespace \
+		--set controller.image=$(HUB_IMG) \
+		--set controller.signerImage=$(SIGNER_IMG) \
+		--set webhookServer.image=$(WEBHOOK_IMG)
+
+.PHONY: helm-undeploy-kmm
+helm-undeploy-kmm: ## Uninstall KMM Helm release from the K8s cluster.
+	helm uninstall $(HELM_RELEASE_KMM) --namespace $(HELM_NAMESPACE)
+
+.PHONY: helm-undeploy-kmm-hub
+helm-undeploy-kmm-hub: ## Uninstall KMM-Hub Helm release from the K8s cluster.
+	helm uninstall $(HELM_RELEASE_KMM_HUB) --namespace $(HELM_NAMESPACE)
+
+.PHONY: helm-undeploy
+helm-undeploy: helm-undeploy-kmm undeploy-cert-manager ## Undeploy KMM controller and cert-manager from the K8s cluster. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+
+.PHONY: helm-undeploy-hub
+helm-undeploy-hub: helm-undeploy-kmm-hub undeploy-cert-manager ## Undeploy KMM-Hub controller and cert-manager from the K8s cluster. Call with ignore-not-found=true to ignore resource not found errors during deletion.
 
 include docs.mk
 
--- a/deployment/helm/kmm/templates/metrics-service.yaml
+++ b/deployment/helm/kmm/templates/metrics-service.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "kmm.namePrefix" . }}controller-metrics-service
+  namespace: {{ .Release.Namespace }}
+  labels:
+    control-plane: controller
+    {{- include "kmm.labels" . | nindent 4 }}
+spec:
+  ports:
+  - name: https
+    port: 8443
+    protocol: TCP
+    targetPort: https
+  selector:
+    control-plane: controller
+    {{- include "kmm.selectorLabels" . | nindent 4 }}
--- a/deployment/helm/kmm/templates/preflightvalidations-crd.yaml
+++ b/deployment/helm/kmm/templates/preflightvalidations-crd.yaml
@@ -1,0 +1,232 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.20.1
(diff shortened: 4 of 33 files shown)


Makefile
```diff
@@ -360,5 +360,53 @@ workerimage-build: ## Build docker image for worker app.
...
+	helm upgrade --install $(HELM_RELEASE_KMM) $(HELM_CHART_KMM) \
+		--namespace $(HELM_NAMESPACE) --create-namespace \
+		--set controller.image=$(IMG) \
+		--set controller.workerImage=$(WORKER_IMG) \
+		--set controller.signerImage=$(SIGNER_IMG) \
+		--set webhookServer.image=$(WEBHOOK_IMG)
+
+.PHONY: helm-deploy-hub
+helm-deploy-hub: deploy-cert-manager manifests ## Deploy KMM-Hub controller using Helm to the K8s cluster specified in ~/.kube/config.
+	@mkdir -p $(HELM_CHART_KMM_HUB)/crds
```
TomerNewman: The KMM chart has its CRDs committed in `deployment/helm/kmm/crds/`, but the KMM-Hub chart does not — the `crds/` directory is created dynamically here via `mkdir -p` at deploy time and CRDs are copied from `config/crd-hub/bases/`.

This means installing the `kmm-hub` chart directly (e.g., via `helm install` without the Makefile) would result in **no CRDs** being applied, and the operator would fail. The KMM chart doesn't have this problem because its CRDs are committed in the repo.

Consider either:
1. Committing the Hub CRDs into `deployment/helm/kmm-hub/crds/` (matching the KMM chart approach), or
2. Removing the committed CRDs from the KMM chart and relying on the Makefile `mkdir`/`cp` pattern for both (making both charts consistent)

deployment/helm/kmm/templates/webhook-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+kind: Service
+metadata:
+  name: {{ include "kmm.namePrefix" . }}webhook-service
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/instance: webhook-service
+    app.kubernetes.io/component: webhook
+    app.kubernetes.io/created-by: kernel-module-management
+    app.kubernetes.io/part-of: kernel-module-management
+    {{- include "kmm.labels" . | nindent 4 }}
```
TomerNewman, line 9: The inline labels here (`app.kubernetes.io/component: webhook`, `app.kubernetes.io/part-of: kernel-module-management`) conflict with the labels emitted by `include "kmm.labels"`, which expands to include `app.kubernetes.io/component: kmm` and `app.kubernetes.io/part-of: kmm` (from `values.yaml`).

This produces **duplicate YAML keys** in the rendered manifest. The YAML spec treats duplicate keys as undefined behavior — Kubernetes happens to accept it (last value wins), but `helm lint --strict` would flag it and some YAML parsers may reject it.

The same issue affects `cert-manager.yaml` (both the Issuer and Certificate resources) and the equivalent files in the `kmm-hub` chart.

Consider removing the conflicting inline labels from these templates since the `include` helper already provides the correct values.

Makefile
```diff
@@ -360,5 +360,53 @@ workerimage-build: ## Build docker image for worker app.
...
+HELM_CHART_KMM_HUB ?= deployment/helm/kmm-hub
+HELM_RELEASE_KMM ?= kmm
+HELM_RELEASE_KMM_HUB ?= kmm-hub
+HELM_NAMESPACE ?= kmm-operator-system
+
+.PHONY: helm-deploy
+helm-deploy: deploy-cert-manager manifests ## Deploy KMM controller using Helm to the K8s cluster specified in ~/.kube/config.
+	@mkdir -p $(HELM_CHART_KMM)/crds
+	@cp config/crd/bases/kmm.sigs.x-k8s.io_modules.yaml $(HELM_CHART_KMM)/crds/
+	@cp config/crd/bases/kmm.sigs.x-k8s.io_nodemodulesconfigs.yaml $(HELM_CHART_KMM)/crds/
```
ybettan: These CRD files are copied from `config/crd/bases/` into `deployment/helm/kmm/crds/` at deploy time. But the chart already has the same 4 CRD files committed in `deployment/helm/kmm/crds/`. This means:

1. The committed CRDs will drift from the source-of-truth in `config/crd/bases/` over time unless someone remembers to re-copy them after every CRD change
2. The Makefile copy silently overwrites them at deploy time, so the committed copies are misleading — they look authoritative but get replaced

Suggestion: either (a) don't commit the CRDs and add `deployment/helm/kmm/crds/` to `.gitignore`, letting this Makefile copy be the only source, or (b) commit them and remove this copy step, relying on a CI check or pre-commit hook to keep them in sync with `config/crd/bases/`.

deployment/helm/kmm/templates/preflightvalidations-crd.yaml
```diff
@@ -0,0 +1,232 @@
+---
```
ybettan, line 1: This CRD is in `templates/` while the other 4 CRDs are in `crds/`. This gives them different Helm lifecycle behavior:

- `crds/` CRDs: installed before templates, **never deleted** by `helm uninstall`, **never upgraded** by `helm upgrade`
- This CRD (in `templates/`): **deleted** on `helm uninstall` (potentially orphaning existing PreflightValidation CRs), **upgraded** on `helm upgrade`

I'm guessing this is intentional — the file has a `cert-manager.io/inject-ca-from` annotation that needs Helm templating, which doesn't work in `crds/`. But the inconsistent lifecycle between CRDs could surprise users. Worth a comment in the file or in a chart README explaining why this one is different.

Dockerfile.kmm-operator-build
```diff
@@ -11,3 +11,7 @@ RUN go install go.uber.org/mock/mockgen@v0.3.0
 RUN curl -L "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl" -o /tmp/kubectl
 RUN install -o root -g root -m 0755 /tmp/kubectl /usr/local/bin/kubectl
 COPY --from=quay.io/openshift/origin-cli:latest /usr/bin/oc /usr/bin
+RUN curl -fsSL https://get.helm.sh/helm-v3.17.3-linux-amd64.tar.gz -o /tmp/helm.tar.gz \
```
ybettan, line 14: The Helm binary is downloaded by version tag but not verified with a SHA256 checksum. For a CI/build image this is a supply chain risk — a compromised CDN or MITM could inject a malicious binary.

Helm publishes checksums alongside each release. Consider adding verification:
```dockerfile
RUN curl -fsSL https://get.helm.sh/helm-v3.17.3-linux-amd64.tar.gz -o /tmp/helm.tar.gz \
    && echo "<expected-sha256>  /tmp/helm.tar.gz" | sha256sum -c - \
    && tar -xzf /tmp/helm.tar.gz -C /tmp \
    && install -o root -g root -m 0755 /tmp/linux-amd64/helm /usr/local/bin/helm \
    && rm -rf /tmp/helm.tar.gz /tmp/linux-amd64
```

(Note: the existing `kubectl` download above has the same issue, but no reason to propagate it.)

Makefile
yevgeny-shnaidman: @ybettan please review the latest changes, i think all those comments are already fixed

deployment/helm/kmm/templates/metrics-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    control-plane: controller
+    {{- include "kmm.labels" . | nindent 4 }}
+spec:
+  ports:
+  - name: https
+    port: 8443
+    protocol: TCP
+    targetPort: https
```
ybettan, line 14: Bug: `targetPort: https` resolves by port **name** on the target pod, but the controller Deployment (`manager-deployment.yaml`) names port 8443 as `metrics`, not `https`:

```yaml
ports:
  - name: metrics
    containerPort: 8443
```

This mismatch means the Service won't find a matching named port on the pod. Should be `targetPort: metrics` (or rename the port in the Deployment to `https`).

Same issue likely applies to `deployment/helm/kmm-hub/templates/metrics-service.yaml`.

deployment/helm/kmm/templates/preflightvalidations-crd.yaml
yevgeny-shnaidman: same comment as above

Dockerfile.kmm-operator-build
yevgeny-shnaidman: This is a skipper Dockerfile. Is skipper used by CI/build

ybettan: I know. But still, if there is an easy checksum check isn't it worth it to validate it? Ignore the "ci part" in the previous message.

Makefile
yevgeny-shnaidman: fixed

ybettan:
**Concern: dual maintenance of kustomize + hand-written Helm charts**

The kustomize manifests under `config/` are the existing source of truth for deployments. This PR adds a second, hand-maintained copy as Helm charts. Over time these will drift — any change to RBAC, CRDs, deployments, or webhooks in `config/` must be manually replicated in `deployment/helm/`, and there's nothing enforcing that today.

**Suggestion:** use [helmify](https://github.com/arttor/helmify) to **generate** the Helm charts from kustomize output, keeping `config/` as the single source of truth. The generated charts live in their own committed directory (`deployment/helm/`), but are always produced from kustomize — never hand-edited:

```makefile
.PHONY: generate-helm
generate-helm: manifests ## Regenerate Helm charts from kustomize manifests.
	$(KUSTOMIZE) build config/default | $(HELMIFY) deployment/helm/kmm
	$(KUSTOMIZE) build config/default-hub | $(HELMIFY) deployment/helm/kmm-hub
```

Then add a CI script `ci/prow/check-helm-charts` following the same pattern as `ci/prow/check-api-changes`:

```bash
#!/usr/bin/env bash
set -euxo pipefail

make generate-helm

if [[ $(git diff) ]]; then
   echo 'Please run `make generate-helm` and repush'
   echo 'The following differences were found:'
   echo $(git diff)
   exit 1
fi
```

This also requires a companion PR to [`kubernetes/test-infra`](https://github.com/kubernetes/test-infra) to register the presubmit job in `config/jobs/kubernetes-sigs/kernel-module-management/kernel-module-management-presubmits.yaml`, mirroring the existing `pull-kernel-module-management-check-api-changes` entry:

```yaml
  - name: pull-kernel-module-management-check-helm-charts
    cluster: eks-prow-build-cluster
    always_run: true
    decorate: true
    annotations:
      testgrid-dashboards: sig-node-kernel-module-management
      testgrid-tab-name: check-helm-charts
      description: "make sure Helm charts are up-to-date with kustomize manifests."
    spec:
      containers:
      - image: public.ecr.aws/docker/library/golang:1.25-alpine
        command: [sh]
        args:
        - -c
        - |
          apk add bash curl gcc git make

          # Install kubectl
          curl -Lo /usr/local/bin/kubectl "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
          chmod +x /usr/local/bin/kubectl

          # Install helm
          curl -fsSL https://get.helm.sh/helm-v3.17.3-linux-amd64.tar.gz | tar -xz -C /tmp
          install /tmp/linux-amd64/helm /usr/local/bin/helm

          # Include the destination directory of `go install` in $PATH
          export PATH=$(go env GOPATH)/bin:${PATH}

          ./ci/prow/check-helm-charts
        env:
        - name: CGO_ENABLED
          value: '0'
        resources:
          limits:
            cpu: 1
            memory: 4Gi
          requests:
            cpu: 1
            memory: 4Gi
```

This way:
1. `deployment/helm/` is always committed and ready to use — no generation step at deploy time
2. `config/` remains the single source of truth — chart changes go through kustomize
3. CI catches any drift on every PR, just like API changes are validated today

This would also resolve the CRD drift issue (inline comment above) — the generated chart would always match `config/crd/bases/`.

One caveat: helmify overwrites templates on every run, so hand-tuned Helm logic would be lost. But that's the tradeoff — single source of truth means edits happen in kustomize, not in the chart.

deployment/helm/kmm/templates/metrics-service.yaml
yevgeny-shnaidman: fixed

yevgeny-shnaidman:
> **Concern: dual maintenance of kustomize + hand-written Helm charts**
> 
> The kustomize manifests under `config/` are the existing source of truth for deployments. This PR adds a second, hand-maintained copy as Helm charts. Over time these will drift — any change to RBAC, CRDs, deployments, or webhooks in `config/` must be manually replicated in `deployment/helm/`, and there's nothing enforcing that today.
> 
> **Suggestion:** use [helmify](https://github.com/arttor/helmify) to **generate** the Helm charts from kustomize output, keeping `config/` as the single source of truth. The generated charts live in their own committed directory (`deployment/helm/`), but are always produced from kustomize — never hand-edited:
> 
> ```makefile
> .PHONY: generate-helm
> generate-helm: manifests ## Regenerate Helm charts from kustomize manifests.
> 	$(KUSTOMIZE) build config/default | $(HELMIFY) deployment/helm/kmm
> 	$(KUSTOMIZE) build config/default-hub | $(HELMIFY) deployment/helm/kmm-hub
> ```
> 
> Then add a CI script `ci/prow/check-helm-charts` following the same pattern as `ci/prow/check-api-changes`:
> 
> ```shell
> #!/usr/bin/env bash
> set -euxo pipefail
> 
> make generate-helm
> 
> if [[ $(git diff) ]]; then
>    echo 'Please run `make generate-helm` and repush'
>    echo 'The following differences were found:'
>    echo $(git diff)
>    exit 1
> fi
> ```
> 
> This also requires a companion PR to [`kubernetes/test-infra`](https://github.com/kubernetes/test-infra) to register the presubmit job in `config/jobs/kubernetes-sigs/kernel-module-management/kernel-module-management-presubmits.yaml`, mirroring the existing `pull-kernel-module-management-check-api-changes` entry:
> 
> ```yaml
>   - name: pull-kernel-module-management-check-helm-charts
>     cluster: eks-prow-build-cluster
>     always_run: true
>     decorate: true
>     annotations:
>       testgrid-dashboards: sig-node-kernel-module-management
>       testgrid-tab-name: check-helm-charts
>       description: "make sure Helm charts are up-to-date with kustomize manifests."
>     spec:
>       containers:
>       - image: public.ecr.aws/docker/library/golang:1.25-alpine
>         command: [sh]
>         args:
>         - -c
>         - |
>           apk add bash curl gcc git make
> 
>           # Install kubectl
>           curl -Lo /usr/local/bin/kubectl "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
>           chmod +x /usr/local/bin/kubectl
> 
>           # Install helm
>           curl -fsSL https://get.helm.sh/helm-v3.17.3-linux-amd64.tar.gz | tar -xz -C /tmp
>           install /tmp/linux-amd64/helm /usr/local/bin/helm
> 
>           # Include the destination directory of `go install` in $PATH
>           export PATH=$(go env GOPATH)/bin:${PATH}
> 
>           ./ci/prow/check-helm-charts
>         env:
>         - name: CGO_ENABLED
>           value: '0'
>         resources:
>           limits:
>             cpu: 1
>             memory: 4Gi
>           requests:
>             cpu: 1
>             memory: 4Gi
> ```
> 
> This way:
> 
> 1. `deployment/helm/` is always committed and ready to use — no generation step at deploy time
> 2. `config/` remains the single source of truth — chart changes go through kustomize
> 3. CI catches any drift on every PR, just like API changes are validated today
> 
> This would also resolve the CRD drift issue (inline comment above) — the generated chart would always match `config/crd/bases/`.
> 
> One caveat: helmify overwrites templates on every run, so hand-tuned Helm logic would be lost. But that's the tradeoff — single source of truth means edits happen in kustomize, not in the chart.

Helmify does not produce a production ready helm charts, we will probably need some adjustments after running helmify, which again brings us to the same place where we are now. The better solution is to use the current commit, and then add the prow test for helm deployment. This will discover any "drifts" that might occur. Additional consideration - at this stage the config hardly ever changes, besides the CRDs, and this is taken care of in the manifests Makefile target

Dockerfile.kmm-operator-build
yevgeny-shnaidman: fixed

deployment/helm/kmm/templates/webhook-service.yaml
yevgeny-shnaidman: fixed

ybettan:
/lgtm
/hold
@TomerNewman Can unhold when he is happy with the changes too.

deployment/helm/kmm/templates/webhook-server-deployment.yaml
```diff
@@ -0,0 +1,91 @@
...
+      containers:
+        - image: {{ .Values.webhookServer.image }}
+          name: webhook-server
+          env:
+            - name: OPERATOR_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          args:
+            - --config={{ .Values.webhookServer.configName }}
```
TomerNewman: on one hand this in unqouted, but on the other hand in `deployment/helm/kmm/templates/manager-deployment.yaml` line 42 it is, I would make it consistent

yevgeny-shnaidman: fixed

PR #1311, opened 2026-07-15
Remove stale device-plugin-target node labels when spec.devicePlugin removed

TomerNewman:
When spec.devicePlugin is removed, the DevicePluginReconciler deletes DaemonSets and clears status but does not remove device-plugin-target node labels. Call removeDevicePluginTargetLabels in the nil branch, matching the cleanup already done in the deletion path.

---


ybettan (commented):
/lgtm

PR #1313, opened 2026-07-19
Adding Helm-based e2e tests for spoke and hub-spoke

yevgeny-shnaidman:
Add two new e2e test scripts that mirror the existing kustomize-based e2e tests but deploy KMM using Helm charts instead:

- ci/prow/e2e-helm-incluster-build: deploys KMM via helm-deploy, then runs the full module build/sign/load/unload lifecycle test.
- ci/prow/e2e-helm-hub-spoke-incluster-build: deploys both KMM and KMM-Hub via helm-deploy/helm-deploy-hub with OCM, then validates ManagedClusterModule propagation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -379,24 +379,21 @@ HELM_RELEASE_KMM ?= kmm
 HELM_RELEASE_KMM_HUB ?= kmm-hub
 HELM_NAMESPACE ?= kmm-operator-system
 HELM_EXTRA_ARGS ?=
+HELM_VALUES_FILE ?=
+HELM_HUB_VALUES_FILE ?=
 
 .PHONY: helm-deploy
 helm-deploy: deploy-cert-manager manifests ## Deploy KMM controller using Helm to the K8s cluster specified in ~/.kube/config.
 	helm upgrade --install $(HELM_RELEASE_KMM) $(HELM_CHART_KMM) \
 		--namespace $(HELM_NAMESPACE) --create-namespace \
-		--set controller.image=$(IMG) \
-		--set controller.workerImage=$(WORKER_IMG) \
-		--set controller.signerImage=$(SIGNER_IMG) \
-		--set webhookServer.image=$(WEBHOOK_IMG) \
+		$(if $(HELM_VALUES_FILE),-f $(HELM_VALUES_FILE)) \
 		$(HELM_EXTRA_ARGS)
 
 .PHONY: helm-deploy-hub
 helm-deploy-hub: deploy-cert-manager manifests ## Deploy KMM-Hub controller using Helm to the K8s cluster specified in ~/.kube/config.
 	helm upgrade --install $(HELM_RELEASE_KMM_HUB) $(HELM_CHART_KMM_HUB) \
 		--namespace $(HELM_NAMESPACE) --create-namespace \
-		--set controller.image=$(HUB_IMG) \
-		--set controller.signerImage=$(SIGNER_IMG) \
-		--set webhookServer.image=$(WEBHOOK_IMG) \
+		$(if $(HELM_HUB_VALUES_FILE),-f $(HELM_HUB_VALUES_FILE)) \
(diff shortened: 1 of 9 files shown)


ci/prow/e2e-helm-incluster-build
```diff
@@ -0,0 +1,62 @@
...
+echo "Add an kmm-ci Module that contains a valid mapping..."
+timeout 1m bash -c 'until kubectl apply -f ci/module-kmm-ci-build-sign.yaml; do sleep 3; done'
+
+echo "Check that the module gets loaded on the node..."
+timeout 10m bash -c 'until minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; done'
+
+echo "Check that the dependent module is also loaded on the node..."
+if ! minikube ssh -- lsmod | grep kmm_ci_b; then
+  echo "Unexpected lsmod output - kmm_ci_b is not loaded on the node"
+  return 1
```
TomerNewman:
```suggestion
  exit 1
```

Makefile
```diff
@@ -379,24 +379,21 @@ HELM_RELEASE_KMM ?= kmm
...
 
 .PHONY: helm-deploy
 helm-deploy: deploy-cert-manager manifests ## Deploy KMM controller using Helm to the K8s cluster specified in ~/.kube/config.
 	helm upgrade --install $(HELM_RELEASE_KMM) $(HELM_CHART_KMM) \
 		--namespace $(HELM_NAMESPACE) --create-namespace \
-		--set controller.image=$(IMG) \
-		--set controller.workerImage=$(WORKER_IMG) \
-		--set controller.signerImage=$(SIGNER_IMG) \
-		--set webhookServer.image=$(WEBHOOK_IMG) \
+		$(if $(HELM_VALUES_FILE),-f $(HELM_VALUES_FILE)) \
```
TomerNewman, line 389: meaning now if u do `make helm-deploy IMG=llocalhost/test-image:latest` (without `HELM_VALUES_FILE` set) the new IMG env will be ignored?

ci/helm-values-ci-spoke.yaml
```diff
@@ -0,0 +1,10 @@
+controller:
```
TomerNewman, line 1: since it is almost the exact same file as `ci/helm-values-ci.yaml` maybe you can use `ci/helm-values-ci.yaml` as a base values file and this way there will not be code duplication? `ci/helm-values-ci-spoke.yaml` only adds `managed: true`

ci/prow/e2e-helm-incluster-build
yevgeny-shnaidman: This is basically the same file and logic as ci/prow/e2e-incluster-build, so lets keep it as is. "exit 1" and "return 1" in this context is the same

Makefile
yevgeny-shnaidman: helm is different then kustomize. "--set" will always take preference over helmfile,so using both when "-set" is always defined  does not make sense. "-set" can be passed in "HELM_EXTRA_ARGS" or partial values can be defined in the temp values file using HELM_VALUES_FILE

ci/helm-values-ci-spoke.yaml
yevgeny-shnaidman: This way it seems more clear to the use how everything is defined. Since it is not a production code, i thought this would be more appropriate

ci/prow/e2e-helm-incluster-build
TomerNewman: so maybe lets fix this `return 1` in `ci/prow/e2e-incluster-build` as well, this is a problem since we can not do return 1 not inside a function.

try for example to run locally 
```
bash -c '#!/usr/bin/env bash
set -euxo pipefail
echo "before return"
return 1
echo "after return"
'
```

yevgeny-shnaidman: Fixed

PR #1318, opened 2026-07-29
Propagate Module tolerations to image pull pods

TomerNewman:
Pull pods created by CreatePullPod did not inherit the tolerations from Module.spec.tolerations. On clusters where all worker nodes have custom taints, the pull pod could not be scheduled and blocked the entire module workflow.

---

/cc @yevgeny-shnaidman

### DIFF
--- a/internal/controllers/mic_reconciler.go
+++ b/internal/controllers/mic_reconciler.go
@@ -286,6 +286,7 @@ func (mrhi *micReconcilerHelperImpl) processImagesSpecs(ctx context.Context, mic
 					oneTimePod,
 					micObj.Spec.ImageRepoSecret,
 					micObj.Spec.ImagePullPolicy,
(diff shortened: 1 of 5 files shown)


internal/controllers/mic_reconciler.go
```diff
@@ -286,7 +286,8 @@ func (mrhi *micReconcilerHelperImpl) processImagesSpecs(ctx context.Context, mic
 					oneTimePod,
 					micObj.Spec.ImageRepoSecret,
 					micObj.Spec.ImagePullPolicy,
-					micObj)
+					micObj,
+					micObj.Spec.Tolerations)
```
yevgeny-shnaidman: lets replace the ordering of the argument - the owner should be last

TomerNewman: done

PR #1319, opened 2026-07-30
Add PRD for modprobe.d config file support

NataliShemtov:
Adds an enhancement doc describing the requirements for letting users supply modprobe.d configuration files in their driver container image, so KMM can apply them during kernel module load/unload.

### DIFF
--- a/docs/enhancements/0005-modprobed-config.md
+++ b/docs/enhancements/0005-modprobed-config.md
@@ -1,0 +1,169 @@
+# Modprobe.d Configuration Files for Driver Container Images
+
+| Field       | Value   |
+|-------------|---------|
+| Author(s)   | Natali Shemtov |
+| Date        | 2026-07-29 |
+
+## 1. Problem Statement
+
+Kernel modules can subscribe to multiple kernel subsystems during
+initialization (for example, PCI). Each subscription increments the
+module's reference count. When the module needs to be unloaded, its exit
+function isn't called until that reference count reaches zero — which
+typically requires running a user-space script to unwind those
+subscriptions first. Today, KMM users have no way to run such a script
+before or after `modprobe` loads or unloads their kernel module, so
+modules that rely on this pattern cannot be reliably unloaded through KMM.
+
+## 2. Goals and Non-Goals
+
+### 2.1 Goals
+
+- Users can supply a set of modprobe.d configuration files in their driver
+  container image, and have KMM apply them so that the corresponding
+  init/de-init scripts run automatically when the kernel module is loaded
+  and unloaded.
+- The modprobe.d capability works correctly together with the existing
+  Firmware loading capability, with no degradation to either.
+- Users are always informed when their modprobe.d configuration could not
+  be applied, rather than experiencing a silent failure.
+- The modprobe.d capability introduces no degradation to the existing
+  `ModulesLoadingOrder` capability — achieved by preventing the two from
+  being enabled together on the same module (see Non-Goals), rather than
+  allowing them to silently conflict.
+- Users can specify where in their driver container image their
+  modprobe.d files live, via a new field on the Module API. KMM always
+  copies those files to the same fixed location on the worker pod
+  (`/etc/modprobe.d/`), regardless of the configured source path.
+  Modules that don't set the field are unaffected (see NFR-2).
+
+### 2.2 Non-Goals
+
+- Live reconfiguration: applying a new or changed set of modprobe.d files
+  to an already-running module requires the user to restart/recreate the
+  module's worker pods.
+- Validation of modprobe.d file contents: KMM does not check or validate
+  the syntax of the modprobe.d files a user provides — malformed files
+  fail at `modprobe` execution time, not as a KMM-reported error.
+- Nested directory structures: only modprobe.d files placed directly in
+  the configured directory are supported. Files placed in sub-directories
+  are not picked up.
+- Combined use with `ModulesLoadingOrder`: `ModulesLoadingOrder` is
+  implemented today by mounting a read-only volume at `/etc/modprobe.d/`
+  in the worker Pod, which collides with this capability's use of the
+  same path. Merging both into a single input (for example, generating
+  and mounting the combined configuration from an init container) is a
+  long-term solution that is out of scope for this enhancement. Users who
+  need both softdep-style module ordering and other modprobe.d directives
+  can express the ordering directly in their own modprobe.d files.
+
+## 3. Requirements
+
+### 3.1 Functional Requirements
+
+- **FR-1:** Users must be able to enable the modprobe.d capability on a
+  module by setting a new field on the Module API that specifies the
+  directory in their driver container image containing their modprobe.d
+  configuration files. The field accepts any well-formed absolute path;
+  KMM's validating webhook must reject the Module if the value isn't a
+  well-formed absolute path, but otherwise does not restrict which
+  directory the user chooses.
+- **FR-2:** When a module has the modprobe.d capability enabled, the
+  configuration files in `/etc/modprobe.d/` must take effect before
+  `modprobe` is invoked to load the kernel module, so that any load-time
+  init sequence defined in those files runs automatically.
+- **FR-3:** When a module has the modprobe.d capability enabled, the
+  configuration files in `/etc/modprobe.d/` must take effect before
+  `modprobe` is invoked to unload the kernel module, so that any de-init
+  sequence defined in those files runs automatically (including
+  sequences needed to release kernel subsystem reference counts before
+  unload can proceed).
+- **FR-4:** If the directory configured for the modprobe.d capability
+  does not exist in the driver container image, or exists but contains no
+  files, the module must fail to load rather than silently proceeding
+  without the modprobe.d configuration.
+- **FR-5:** If the directory configured for the modprobe.d capability
+  contains a sub-directory, the module must fail to load.
+- **FR-6:** Applying an updated set of modprobe.d files requires the user
+  to restart or recreate the module's worker pods; a running pod does not
+  pick up changes to the configured directory automatically.
+- **FR-7:** The modprobe.d capability must continue to work correctly for
+  modules that also use Firmware loading.
+- **FR-8:** KMM must reject, via validating webhook, any Module that
+  enables the modprobe.d capability while also setting
+  `ModulesLoadingOrder`, since the two currently cannot coexist (see
+  Non-Goals). The rejection must clearly state why the combination isn't
+  supported and point the user to expressing module ordering directly in
+  their own modprobe.d files instead.
+
+### 3.2 Non-Functional Requirements
+
+- **NFR-1:** Enabling the modprobe.d capability on a module requires that
+  module's worker pods to run with a Privileged security context, applied
+  automatically by KMM. Users must be able to determine, before deploying,
+  that enabling the modprobe.d capability implies this elevated privilege
+  level for the module's worker pods.
+- **NFR-2:** Modules that do not enable the modprobe.d capability continue
+  to behave exactly as they do today — this capability introduces no
+  change for existing modules.
+- **NFR-3:** When the modprobe.d configuration cannot be applied (per
+  FR-4 or FR-5), the user must be able to observe the failure and its
+  cause through the Module's status or a Kubernetes event, without having
+  to inspect pod logs to discover it.
+
+## 4. Acceptance Criteria
+
+- [ ] A user can enable the modprobe.d capability on a module and place
+      configuration files in the directory they configured in their
+      driver container image, and after deploying the Module, the
+      load-time init sequence defined in those files runs as part of
+      loading the kernel module.
+- [ ] The de-init sequence defined in the user's modprobe.d files runs as
+      part of unloading the kernel module, including for modules that
+      require this sequence to release subsystem reference counts before
+      `modprobe -r` can succeed.
+- [ ] A user who enables the modprobe.d capability but whose image has no
+      directory at the configured path, or an empty one, sees the module
+      fail to load and can observe the failure via the Module's status or
+      a Kubernetes event.
+- [ ] A user who enables the modprobe.d capability and whose configured
+      directory contains a sub-directory sees the module fail to load and
+      can observe the failure via the Module's status or a Kubernetes
+      event.
+- [ ] A user can use the modprobe.d capability on a module that also uses
+      Firmware loading, and both capabilities work correctly together.
+- [ ] A user who tries to enable the modprobe.d capability on a module
+      that also sets `ModulesLoadingOrder` is rejected by the validating
+      webhook, with a message explaining the conflict and the
+      recommended workaround.
+- [ ] A user who sets the modprobe.d field to a value that isn't a
+      well-formed absolute path is rejected by the validating webhook.
+
+## 5. Assumptions
+
+- Clusters that adopt this feature permit Privileged pods for modules
+  that use it — clusters that uniformly block Privileged pods (e.g., via
+  a "restricted" PodSecurity standard applied without exception) would be
+  unable to use this capability at all. This mirrors an existing
+  precondition for modules that use Firmware loading today, which is
+  already documented.
+- Users authoring modprobe.d configuration files are already familiar with
+  the modprobe.d file format; KMM provides no authoring assistance or
+  content validation.
+
+## 6. Future Work
+
+- **Long-term merge:** revisit combining KMM-generated configuration
+  (for example, the softdep config produced for `ModulesLoadingOrder`)
+  and user-supplied modprobe.d files into a single input, so the two
+  capabilities can be used together. This likely requires moving away
+  from the current read-only DownwardAPI-based mount for
+  `ModulesLoadingOrder` toward generating configuration into a shared,
+  writable volume (for example, from an init container).
+- **Remove `ModulesLoadingOrder`:** once the modprobe.d capability is
+  available, users can express module load ordering directly in their
+  own modprobe.d files, making the `ModulesLoadingOrder` field redundant.
+  Track a follow-up task to remove it from the API in KMM 3.0 (the next
+  version where breaking API changes for existing users are permitted),
+  once this enhancement's work is complete.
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -33,3 +33,7 @@ testbin/*
 *~
 
 .artifacts/
+
+# ai-workflows (github.com/flightctl/ai-workflows) generated/symlinked files
+.cursor/commands/
+.cursor/skills/
```
ybettan: Those changes should stay local. No need to push them..

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+- The modprobe.d capability works correctly together with the existing
+  `ModulesLoadingOrder` capability, with no degradation to either.
+- Users are always informed when their modprobe.d configuration could not
+  be applied, rather than experiencing a silent failure.
+
+### 2.2 Non-Goals
+
+- Live reconfiguration: applying a new or changed set of modprobe.d files
+  to an already-running module requires the user to restart/recreate the
+  module's pods.
```
ybettan: What are `module's pods`? Can we mention what pods exactly? worker-pods? puller-pods? build-pods?

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+- Validation of modprobe.d file contents: KMM does not check or validate
+  the syntax of the modprobe.d files a user provides — malformed files
+  fail at `modprobe` execution time, not as a KMM-reported error.
+- Nested directory structures: only modprobe.d files placed directly in
+  the designated directory are supported. Files placed in sub-directories
+  are not picked up.
```
ybettan, line 51: This is something that can be moved to goals since it will be easy to validate it in the module's webhook.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+
+## 3. Requirements
+
+### 3.1 Functional Requirements
+
+- **FR-1:** Users must be able to designate a directory in their driver
+  container image that contains their modprobe.d configuration files.
```
ybettan: Do we want to allow any directory on the driver-container? This will require an API change. It is worth considering always having it in `/etc/modprobe.d/` always and document it as users aren't familiar with KMM's implementation details (their files are being copied into a worker file) so have it it `/etc/modprobe.d/` may be what they expect and it will be simpler to implement.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+  in those files runs automatically.
+- **FR-3:** When a user has designated a modprobe.d directory, the
+  configuration files in it must take effect before `modprobe` is invoked
+  to unload the kernel module, so that any de-init sequence defined in
+  those files runs automatically (including sequences needed to release
+  kernel subsystem reference counts before unload can proceed).
+- **FR-4:** If the designated directory does not exist in the driver
+  container image, or exists but contains no files, the module must fail
+  to load rather than silently proceeding without the modprobe.d
+  configuration.
```
ybettan: I would say that the module should be rejected by k8s instead of fail.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+- **FR-5:** If the designated directory contains a sub-directory, the
```
ybettan: I would say that the module should be rejected by k8s instead of fail.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+  module must fail to load.
+- **FR-6:** Applying an updated set of modprobe.d files requires the user
+  to restart or recreate the module's pods; a running pod does not pick
```
ybettan: `module's pods`? Same comment as before.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+  up changes to the directory automatically.
+- **FR-7:** The modprobe.d capability must continue to work correctly for
+  modules that also use Firmware loading.
+- **FR-8:** The modprobe.d capability must continue to work correctly for
+  modules that also use `ModulesLoadingOrder`.
+
+### 3.2 Non-Functional Requirements
+
+- **NFR-1:** Enabling the modprobe.d capability on a module requires that
+  module's pods to run with a Privileged security context, applied
```
ybettan: `module's pods`?

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+  automatically by KMM. Users must be able to determine, before deploying,
+  that designating a modprobe.d directory implies this elevated privilege
+  level for the module's pods.
+- **NFR-2:** Modules that do not designate a modprobe.d directory continue
+  to behave exactly as they do today — this capability introduces no
+  change for existing modules.
+- **NFR-3:** When the modprobe.d configuration cannot be applied (per
+  FR-4 or FR-5), the user must be able to observe the failure and its
+  cause through the Module's status or a Kubernetes event, without having
+  to inspect pod logs to discover it.
```
ybettan, line 113: k8s events is the way to go if it is rejected by the webhook.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+- [ ] The de-init sequence defined in the user's modprobe.d files runs as
+      part of unloading the kernel module, including for modules that
+      require this sequence to release subsystem reference counts before
+      `modprobe -r` can succeed.
+- [ ] A user who designates a modprobe.d directory that doesn't exist, or
+      is empty, sees the module fail to load and can observe the failure
+      via the Module's status or a Kubernetes event.
+- [ ] A user who designates a modprobe.d directory containing a
+      sub-directory sees the module fail to load and can observe the
+      failure via the Module's status or a Kubernetes event.
```
ybettan: Let's use webhook instead.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,145 @@
...
+
+Since KMM automatically applies a Privileged security context whenever a
+modprobe.d directory is designated, users on clusters that enforce
+restrictive PodSecurity standards may be unable to use this capability, or
+may need to make cluster-level policy exceptions to adopt it. This is not
+a new category of risk for KMM: the same trade-off already exists today
+for modules that use Firmware loading, which also triggers automatic
+Privileged elevation, and is already covered by KMM's documented "Pod
+Security standards" guidance instructing administrators to label the
+Module's namespace accordingly.
```
ybettan: Is this a real degradation? Such clusters will probably reject KMM as a whole since it messes with the node's kernel.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: I clarified it now - worker-pods.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Yes - clarified.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Yes - clarified.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Good point - I actually lean toward your suggestion. 
I originally made it configurable simply because that's what I understood from the Jira ticket.
Should I switch to your solution?

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Checked this - your point stands.
Any cluster already running Firmware modules already accepts the same Privileged requirement, so modprobe.d isn't a new "risk". I will remove that section.

docs/enhancements/0005-modprobed-config.md
TomerNewman: @ybettan how would you validate it in the webhook? the files are sitting in the container image, and the webhook validates before KMM does the build

TomerNewman (commented):
I would also mention here the implementation mechanism, meaning that the modeprobe.d files are being copied from the container image into the worker pod's filesystem.

docs/enhancements/0005-modprobed-config.md
ybettan: If making it configurable mean changing the API, although it is not a breaking change, I would try to avoid it. We can always introduce it in a later version if requested.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Confirmed that webhook validation isn't possible here (image contents aren't visible to the webhook), so keeping the runtime-failure criteria as written.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: The webhook can't validate this since these files live inside the container image, not the Module spec. Keeping this as a runtime failure.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Webhook can't inspect the image, so keeping "fail" here.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Same reasoning as previous comment.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Webhook rejection isn't feasible, so this event fires on runtime failure instead.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Agreed - avoiding the API change is the stronger argument.
I'll switch to a fixed /etc/modprobe.d/ path instead of a configurable field.
And as you said, We can always add that later if it's actually requested.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,130 @@
...
+  image, or exists but contains no files, the module must fail to load
+  rather than silently proceeding without the modprobe.d configuration.
+- **FR-5:** If `/etc/modprobe.d/` contains a sub-directory, the module
+  must fail to load.
+- **FR-6:** Applying an updated set of modprobe.d files requires the user
+  to restart or recreate the module's worker pods; a running pod does not
+  pick up changes to `/etc/modprobe.d/` automatically.
+- **FR-7:** The modprobe.d capability must continue to work correctly for
+  modules that also use Firmware loading.
+- **FR-8:** The modprobe.d capability must continue to work correctly for
```
TomerNewman: The current modulesLoadingOrder implementation mounts a read-only DownwardAPI volume at `/etc/modprobe.d` in the worker container (see setWorkerSofdepConfig in internal/pod/workerpodmanager.go).
Since this feature also needs to place files at `/etc/modprobe.d/`, there's a mount-path collision - Kubernetes won't allow two different volume sources at the same path.
This requirement needs to describe how the two coexist.

docs/enhancements/0005-modprobed-config.md
ybettan: We may need to add a boolean to the API anyway to conditionally copy those files to the worker pods but we can wait with that for later implementation.

ybettan:
/approve

@TomerNewman Can lgtm once he is happy with the enhancement too.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: True, I suggest the following : 

modprobe has a `-C / --config <dir>` flag that overrides its default config directory entirely. KMM already uses this same trick for firmware ` (--firmware-path) ` and kmods `(--dirname)` - pointing modprobe at its own private folder instead of touching real system paths.

Do you think we should do the same here ?  merge the generated softdep.conf and the user's modprobe.d files into one KMM-managed folder, then pass it via -C.
Nothing touches /etc/modprobe.d at all, so there's no collision.

**One implication**: ModulesLoadingOrder's existing mount would also need to move into this same private folder (no user-facing change, just an internal one) so -C picks it up consistently.

@ybettan @TomerNewman  do you think its the right way to go ?

TomerNewman: 1. What if the user ships softdep.conf file in his modeprobe.d dir while also using ModulesLoadingOrder?
2. DownwardAPI can not co-exist with writeable files since its a read-only, so it means that the worker code will need to generate the softdep.conf itself.

ybettan: > One implication: ModulesLoadingOrder's existing mount would also need to move into this same private folder (no user-facing change, just an internal one) so -C picks it up consistently.

How do we do that? We still have 2 different volumes that tries to map to the same location isn't it? It is not some copy code that we control, those are k8s mounts- I don't see an easy way to "combine" them.

I think we could have a short-term and long-term solution here:

*Short term*
We restrict the usage of both `moduleLoadingOrder` and setting `modprobe.d` in the kmod container (this may require an API change to mention that the user request us to use its modprobe.d from the container - then it can be validated in the webhook). If a user needs both, it just put both in its `modprobe.d` file.

*Long tem* 
We should combine all modprobe.d user's configuration in a single input.

This requires some thinking to see if it is possible to achieve without introducing breaking changes. We may need to pass on the API new field and have a more complex validation later in the chain just to not break the API. Or we can have the API change but use the API we will use for the long term solution.

NataliShemtov: Following up our sync meeting -  we decided to go with the short-term solution for now. 
KMM will enforce that ModulesLoadingOrder and the modprobe.d-from-container field are mutually exclusive on a given module via webhook validation, since they'd otherwise both need /etc/modprobe.d. 

Anyone who needs both today can just put their softdep directives directly in their own modprobe.d files instead.

We're holding off on designing the long-term solution for now. Instead, we'll open a follow-up task, to be picked up after this epic, to remove the **ModulesLoadingOrder** field entirely in KMM 3.0 - at that point users would express the same ordering behavior purely through their own modprobe.d files, so the field becomes redundant.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,160 @@
...
+  to restart or recreate the module's worker pods; a running pod does not
+  pick up changes to `/etc/modprobe.d/` automatically.
+- **FR-7:** The modprobe.d capability must continue to work correctly for
+  modules that also use Firmware loading.
+- **FR-8:** KMM must reject, via validating webhook, any Module that
+  enables the modprobe.d capability while also setting
+  `ModulesLoadingOrder`, since the two currently cannot coexist (see
+  Non-Goals). The rejection must clearly state why the combination isn't
+  supported and point the user to expressing module ordering directly in
+  their own modprobe.d files instead.
```
ybettan, line 98: If we use webhook validtation, it means that we have added a field to the API asking to using custom modprobe.d files. 

So:
1. This enhancement should mention the API change
2. If we already modifying the API, we should consider making it a configurable path on the driver-container (but always mounted to /etc/modprobe.d on the worker pod) instead of just a boolean.

docs/enhancements/0005-modprobed-config.md
```diff
@@ -0,0 +1,160 @@
...
+  container image, and have KMM apply them so that the corresponding
+  init/de-init scripts run automatically when the kernel module is loaded
+  and unloaded.
+- The modprobe.d capability works correctly together with the existing
+  Firmware loading capability, with no degradation to either.
+- Users are always informed when their modprobe.d configuration could not
+  be applied, rather than experiencing a silent failure.
+- Users are prevented, via validation, from enabling the modprobe.d
+  capability together with `ModulesLoadingOrder` on the same module, since
+  the two capabilities cannot currently coexist (see Non-Goals).
```
ybettan: The goals mention Firmware compatibility but don't mention the mutual exclusion with `ModulesLoadingOrder`. FR-8 covers it in the requirements, but the goals section only calls out Firmware — readers might wonder why `ModulesLoadingOrder` isn't mentioned here too. Consider adding a goal bullet about the mutual exclusion to keep the goals and requirements in sync.

NataliShemtov: This is already there - do you think it should be more clear?

docs/enhancements/0005-modprobed-config.md
NataliShemtov: Agreed on point 1 -  will update the doc to explicitly call out that this adds a new field to the API.

On point 2, I'd lean toward keeping it a simple boolean with the path still fixed at /etc/modprobe.d/, rather than a configurable path. 
You mentioned before that users aren't familiar with KMM's internals, so /etc/modprobe.d/ is what they'd expect, and it keeps things simpler to implement and document. Unless there's a concrete case you have in mind for needing a custom path?

ybettan: I don't have any corner case in mind, I just though that if we already add the API then we can make it a path instead of a boolean just in case users can't put files in that specific directory on the driver-contaier.

It is just a bit more safe so we don't need to fix it later on if user's request it to be configurable (and then the fix introduce a breaking change boolean --> path).

docs/enhancements/0005-modprobed-config.md
ybettan: Is it? I would expect another bullet before the last one saying something like:
```
The modprobe.d capability works correctly together with the existing
  `ModulesLoadinOrder` capability, with no degradation to either.
```

docs/enhancements/0005-modprobed-config.md
NataliShemtov: That's a fair point , lets go with that then.

docs/enhancements/0005-modprobed-config.md
NataliShemtov: I can reword the existing bullet to sound more parallel to the Firmware one while staying accurate, e.g.: "The modprobe.d capability introduces no degradation to the existing ModulesLoadingOrder capability - achieved by preventing the two from being enabled together on the same module (see Non-Goals), rather than allowing them to silently conflict."

PR #1322, opened 2026-07-30
Add configurable StartupProbe and LivenessProbe to Module CR

### LINKED ISSUE #1320: Support configurable probes for DRA and DevicePlugin containers
Neither CommonContainerSpec nor DRASpec/DevicePluginSpec expose probe fields, so there is no way to configure container probes through the Module CR.

Current behavior:

DRA containers get a hardcoded LivenessProbe (dra_reconciler.go:493-504):
  - gRPC on port 51515, service "liveness"
  - InitialDelaySeconds: 30, PeriodSeconds: 10, TimeoutSeconds: 5,
    FailureThreshold: 3
  - No StartupProbe

DevicePlugin containers get no probes at all (device_plugin_reconciler.go:461 passes nil).

Both use CommonContainerSpec, which has no probe fields.

This is a problem for DRA drivers (and potentially device plugins) with slow initialization. Without a StartupProbe, the DRA liveness probe can kill the container before the driver finishes starting — the current defaults give roughly 60s before the first liveness failure (30s initial delay + 3 failures * 10s). A StartupProbe with a higher failure threshold (e.g. 60 retries * 10s = 10 min) would provide a startup grace period without weakening steady-state liveness detection.

Proposed change: add optional StartupProbe and LivenessProbe fields to CommonContainerSpec. When nil, fall back to the current defaults (hardcoded probe for DRA, no probe for DevicePlugin). This would let operators customize probe timing for both container types through the Module CR.

abyrne55:
This PR adds optional `startupProbe` and `livenessProbe` fields to `CommonContainerSpec`, inherited by both DRA and DevicePlugin containers
- When nil, preserves current behavior (hardcoded GRPC liveness probe for DRA, no probes for DevicePlugin)
- When set, the CR-specified probe overrides the default
- The webhook will reject any attempts to set probes on InitContainers

`readinessProbe` is intentionally omitted for now due to its marginal benefit. DaemonSet pods don't sit behind a Service, so readiness has no effect on traffic routing. We can easily add it later if need be.

Fixes #1320

### Testing

Tested on a single-node OCP 4.22 cluster (kernel 5.14.0-687.15.1.el9_8.x86_64) by porting these changes to the [OCP-compatible fork](https://github.com/rh-ecosystem-edge/kernel-module-management), building/deploying, and running manual e2e tests against both DRA and DevicePlugin Modules serially.

**DRA mode** (Module with `spec.dra` only, no `moduleLoader`, using `spec.selector` as DaemonSet nodeSelector):

| Test | Description | Result |
|------|-------------|--------|
| DRA-1 | Default — hardcoded GRPC liveness probe present, no startupProbe | PASS |
| DRA-2 | Patch custom exec livenessProbe — replaces GRPC default | PASS |
| DRA-3 | Patch startupProbe — appears alongside custom liveness | PASS |
| DRA-4 | JSON-patch remove both — reverts to GRPC default, no startup | PASS |

**DevicePlugin mode** (Module with `spec.devicePlugin` only):

| Test | Description | Result |
|------|-------------|--------|
| DP-1 | Default — no liveness or startup probes | PASS |
| DP-2 | Patch custom exec livenessProbe — appears on DaemonSet | PASS |
| DP-3 | Patch startupProbe — appears alongside liveness | PASS |
| DP-4 | JSON-patch remove both — back to no probes | PASS |

Unit tests also pass with the new test entries for both reconcilers (custom liveness, custom startup, and combined probes).


internal/controllers/dra_reconciler.go
```diff
@@ -534,8 +539,8 @@ func (dsci *draDaemonSetCreatorImpl) setDRAAsDesired(
 				Finalizers: []string{constants.NodeLabelerFinalizer},
 			},
 			Spec: v1.PodSpec{
-				InitContainers:               generatePodContainerSpec(mod.Spec.DRA.InitContainer, "dra-init", nil, nil, nil),
-				Containers:                   generatePodContainerSpec(&mod.Spec.DRA.Container, "dra", containerVolumeMounts, presetEnv, draLivenessProbe),
+				InitContainers:               generatePodContainerSpec(mod.Spec.DRA.InitContainer, "dra-init", nil, nil, nil, nil),
```
TomerNewman, line 542: Right now a user theoretically add `mod.Spec.DRA.InitContainer.StartupProbe` but it will be silently ignored, I think we should reject it in the webhook

abyrne55: Good catch! Done (ea5a330)

ybettan (commented):
/lgtm

PR #1323, opened 2026-07-31
Support configurable SecurityContext for DevicePlugin and DRA containers

### LINKED ISSUE #1321: Support configurable SecurityContext for DevicePlugin and DRA containers
Neither `CommonContainerSpec` nor `PluginContainerSpec` expose a `SecurityContext` field, so there is no way to override the container security context through the Module CR.

## Current behavior

DevicePlugin and DRA containers get a hardcoded `privileged: true` security context (`device_plugin_reconciler.go` `generatePodContainerSpec`), with no override via the Module CR.

Module loader pods are more nuanced: they only run `privileged: true` when firmware loading is needed, otherwise they use just `SYS_MODULE` (PR #537). DevicePlugin and DRA containers never received the same treatment.

This is a problem for operators deploying device plugins that don't need full privilege. For example, the Intel GPU device plugin only needs hostPath access to `/dev/dri` and `/sys/class/drm` and can run with a hardened security context:

```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
  readOnlyRootFilesystem: true
  seccompProfile:
    type: RuntimeDefault
  seLinuxOptions:
    type: container_device_plugin_t
```

Under KMM, all of that is lost: DP pods run fully privileged regardless.

## Proposed change

Add an optional `SecurityContext` field to `CommonContainerSpec` (or `PluginContainerSpec`). When nil, fall back to the current hardcoded privileged context. This would let operators customize the security posture for both container types through the Module CR.

This follows the same pattern proposed in #1320 for configurable probes.

abyrne55:
Add an optional SecurityContext field to CommonContainerSpec. When set, it overrides the default privileged security context for DevicePlugin and DRA containers. When nil, the existing privileged: true default is preserved.

Fixes #1321

### Testing

Tested on a single-node OCP 4.22 cluster (kernel 5.14.0-687.15.1.el9_8.x86_64) by porting these changes to the [OCP-compatible fork](https://github.com/rh-ecosystem-edge/kernel-module-management), building/deploying, and running manual e2e tests against both DRA and DevicePlugin Modules serially.

| Test                                | Result                                                                                                   |
| ----------------------------------- | -------------------------------------------------------------------------------------------------------- |
| DP-1: Default (no securityContext)  | `{"privileged": true}`                                                                                   |
| DP-2: Custom securityContext        | `{"allowPrivilegeEscalation": false, "capabilities": {"drop": ["ALL"]}, "readOnlyRootFilesystem": true}` |
| DP-3: Remove securityContext        | Reverts to `{"privileged": true}`                                                                        |
| DRA-1: Default (no securityContext) | `{"privileged": true}`                                                                                   |
| DRA-2: Custom securityContext       | Same custom SC applied                                                                                   |
| DRA-3: Remove securityContext       | Reverts to `{"privileged": true}`                                                                        |

Unit tests also pass with the new test entries for both reconcilers.


internal/controllers/device_plugin_reconciler_test.go
```diff
@@ -975,6 +975,48 @@ var _ = Describe("DevicePluginReconciler_setDevicePluginAsDesired", func() {
 			true,
 		),
 	)
+
+	It("should use a custom SecurityContext when set in the container spec", func() {
```
TomerNewman, line 1063: I would add here a unit test that tests that privileged: true when SecurityContext is nil

internal/controllers/dra_reconciler_test.go
```diff
@@ -800,6 +800,49 @@ var _ = Describe("DRAReconciler_setDRAAsDesired", func() {
 			utils.GetKernelModuleReadyNodeLabel(namespace, draModuleName), "",
 		))
 	})
+
+	It("should use a custom SecurityContext when set in the container spec", func() {
```
TomerNewman, line 878: same here

api/v1beta1/module_types.go
```diff
@@ -299,6 +299,11 @@ type CommonContainerSpec struct {
 	// VolumeMounts is a list of volume mounts that are appended to the default ones.
 	// +optional
 	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
+
+	// SecurityContext overrides the default security context for this container.
+	// When nil, the container runs with privileged: true.
+	// +optional
```
TomerNewman, line 318:
```suggestion
	// SecurityContext overrides the default security context for this container.
	// When nil, the default privileged security context is applied.
	// Ensure the specified context grants sufficient permissions for the
	// container's operation (e.g., device access); an overly restrictive
	// context will cause the pod to fail at runtime.
	// +optional
```

abyrne55: Thanks for the review! All suggestions implemented with 0af6333

TomerNewman:
/hold

#1335 blocks this PR, we need to wait until it is merged

TomerNewman:
@abyrne55 Can you please rebase your PR? the CI should pass fine now

PR #1329, opened 2026-08-15
Fix DRA driver Pod not evicted during node drain

### LINKED ISSUE #1328: DRA DaemonSet Pod stays on a cordoned node while KMM unloads the kernel module
#1274 fixed a drain deadlock for the device plugin by adding a `device-plugin-target` node label. The DRA reconciler never got the same treatment, so for a Module that has both `spec.moduleLoader` and `spec.dra`, the DRA driver Pod can still be running on a cordoned node at the moment KMM creates the worker Pod that unloads the kernel module the driver sits on top of.

I went looking for this after reading #1274 and noticing that `DRAReconciler.SetupWithManager` (`internal/controllers/dra_reconciler.go:75-88`) watches `Module`, owned `DaemonSet`s and `DeviceClass`, but not `Node`.

## Why kmm-ready is not enough on its own

DaemonSet Pods get an automatic toleration for `node.kubernetes.io/unschedulable:NoSchedule`, and `kubectl drain` will not evict them (it refuses to run without `--ignore-daemonsets`, and skips them with it). Cordoning a node therefore does nothing to the DRA Pod by itself. The only thing that can remove it is its nodeSelector no longer matching.

For a Module with a ModuleLoader, `setDRAAsDesired` (`dra_reconciler.go:511-523`) builds that nodeSelector out of `kmm-ready` plus, during an ordered upgrade, the schedule-pod version label. The device plugin's equivalent (`device_plugin_reconciler.go:505-524`) asks for `kmm-ready` and `device-plugin-target`, and `handleDevicePluginTargetLabels` (`device_plugin_reconciler.go:232-259`) strips that second label as soon as the node stops being schedulable.

Meanwhile the unload path is already moving:

1. `ModuleReconciler` only targets schedulable nodes (`module_reconciler.go:125`), so the cordoned node drops out.
2. That leaves its NMC as a leftover in `prepareSchedulingData`, which becomes `disableModuleOnNode` and removes the Module from the NMC `.spec` (`module_reconciler.go:143-151`, `441-467`).
3. The status entry is now an orphan, so `ProcessUnconfiguredModuleStatus` creates the unloader Pod (`nmc_reconciler.go:133-142`, `409-437`). That function checks `IsNodeRebooted` but not schedulability.
4. `kmm-ready` survives all of it, because `removeOrphanedLabels` (`nmc_reconciler.go:742-755`) only drops a label when the module appears in neither spec nor status, and the status entry stays until the unloader succeeds.

So the DRA Pod keeps matching, and the unloader runs anyway.

## There is a race here, which is probably why it went unnoticed

`nmc_reconciler.go:106-117` does remove `kmm-ready` when a node becomes unschedulable, but only for modules that are still in the NMC `.spec`. `ModuleReconciler` and `NMCReconciler` are separate controllers with separate workqueues and both are woken by the same node spec change, so which one gets there first is not defined:

* If the NMC reconciler wins, the module is still in `.spec`, `kmm-ready` comes off, and the DRA Pod stops matching.
* If the module reconciler wins, the module has already left `.spec`, that branch is skipped, and the unloader is created while `kmm-ready` and the DRA Pod are both still in place.

Removal of `device-plugin-target` does not depend on NMC spec state at all, which is why the device plugin is covered either way and DRA only in the first case.

## Reproducing it

I don't have a DRA-capable cluster with a driver that actually pins the module, so I drove the real reconciler code with a fake client and a mocked worker-pod manager, in the ordering above. Everything below comes from production code paths; only Pod creation is mocked.

```
=== RUN   TestRepro1_NodeSelectors
  DRA           DaemonSet nodeSelector: map[kmm.node.kubernetes.io/default.kmm-ci-dra.ready:]
  DevicePlugin  DaemonSet nodeSelector: map[kmm.node.kubernetes.io/default.kmm-ci-dra.device-plugin-target: kmm.node.kubernetes.io/default.kmm-ci-dra.ready:]
--- PASS
=== RUN   TestRepro2_AfterCordon
  node labels after cordon reconcile: map[kmm.node.kubernetes.io/default.kmm-ci-dra.ready: kubernetes.io/hostname:worker-1]
  cordoned node still matches DRA          DaemonSet: true
  cordoned node still matches DevicePlugin DaemonSet: false
--- PASS
=== RUN   TestRepro3_ModuleReconcilerDropsNode
  ModuleReconciler schedulable targets after cordon: 0
  NMC spec modules:   0
  NMC status modules: 1
--- PASS
=== RUN   TestRepro4_UnloaderRunsWhileDRAPodStillMatches
  CreateUnloaderPod called for default/kmm-ci-dra on worker-1
  unloader Pod created:            true
  kmm-ready still on node:         true
  DRA DaemonSet still selects node: true
--- PASS
```

Happy to attach the test file if it's useful; it's around 250 lines and mostly setup.

## Steps to reproduce on a cluster

1. Apply a Module with both `spec.moduleLoader` and `spec.dra`. `ci/module-kmm-ci-dra.yaml` already has that shape.
2. Wait for the kmod to load and the DRA DaemonSet Pod to report ready.
3. `kubectl cordon <node>`.
4. The node leaves the Module's schedulable targets and a `kmm-worker-<node>-<module>` unloader Pod appears, while the DRA DaemonSet Pod is still `Running`.

Whether that ends in a hang or just a narrow window depends on whether the driver holds something that blocks `rmmod`. `ci/prow/e2e-dra` uses `dra-example-driver`, which doesn't, so CI wouldn't catch it today. There's no cordon or drain step anywhere under `ci/` or `.github/` in any case.

## Expected behavior

When a target node becomes unschedulable, the DRA DaemonSet should stop selecting it so the driver Pod terminates, rather than the unloader racing a live driver.

## Proposed change

Mirror what #1274 did for the device plugin:

* a `GetDRATargetNodeLabel` helper next to `GetDevicePluginTargetNodeLabel` in `internal/utils/kmmlabels.go`, producing `kmm.node.kubernetes.io/<ns>.<name>.dra-target`. `pkg/labels` only exposes the `-ready` labels today and `device-plugin-target` was kept internal, so I'd keep this one internal too unless you'd rather have it as public API;
* `handleDRATargetLabels` and `removeDRATargetLabels` on `draReconcilerHelperAPI`, wired into `Reconcile` the same way as the device plugin ones, including cleanup on Module deletion, on `spec.dra` removal, and on `spec.moduleLoader` being removed from an existing Module;
* the DRA DaemonSet nodeSelector asking for both labels when there is a ModuleLoader:

```go
nodeSelector := map[string]string{
	utils.GetKernelModuleReadyNodeLabel(mod.Namespace, mod.Name): "",
	utils.GetDRATargetNodeLabel(mod.Namespace, mod.Name):         "",
}
```

* a Node watch on `DRAReconciler`, which means passing it the `*filter.Filter` that `cmd/manager/main.go:134` already builds and hands to `NewDevicePluginReconciler` on line 150. `NewDRAReconciler` on line 184 currently takes only `client`, `nodeAPI` and `scheme`. `DevicePluginReconcilerNodePredicate` can be reused as is; renaming it to something that reads sensibly for both callers would be nicer but adds diff, so I'd leave that for a follow-up.

Two details worth getting right.

The cleanup function has to list nodes by the target label rather than by `mod.Spec.Selector`, otherwise a selector change orphans labels on nodes that are no longer selected. `removeDevicePluginTargetLabels` already does it that way.

Schedulability has to be evaluated against `mod.Spec.Tolerations`, so that the workaround from #1271 keeps working: a Module that tolerates `node.kubernetes.io/unschedulable` stays targeted on a cordoned node, keeps its target label, and is never unloaded. Nothing here changes when KMM decides to unload, only what happens to the DRA Pod when it does.

DRA-only Modules should be left alone. `dra_reconciler.go:521-523` already falls back to `mod.Spec.Selector` when there's no ModuleLoader, matching `device_plugin_reconciler.go:519-521`, and with no kmod involved there's nothing to sequence against.

For tests I'd cover, in `dra_reconciler_test.go`, the schedulable and unschedulable cases, a node whose taint is tolerated by the Module, `spec.dra` removal, Module deletion, the ModuleLoader-removed transition, error aggregation across several failing nodes, and the generated nodeSelector for both the ModuleLoader and DRA-only shapes. Plus a cordon/uncordon step in `ci/prow/e2e-dra` asserting the DaemonSet's desired count drops to zero and comes back. `mock_dra_reconciler.go` would come from `make generate` rather than by hand.

Happy to open the PR if this looks like the right shape.

## One thing I noticed while reading #1274

Its description says:

> The NMC reconciler defers unloader creation while the `device-plugin-ready` label is present on an unschedulable node, ensuring the unloader only runs after the device plugin pod terminates.

That doesn't appear to have made it into the merged commit. `dcc567e` touches no NMC file, and `device-plugin-ready` has never appeared in `nmc_reconciler.go`. So both paths currently rely on label removal alone, and nothing waits for the Pod to actually terminate before the unloader starts. If that was dropped deliberately then the same reasoning applies to `dra-ready` and I'll leave it out; if it got lost, it may be worth restoring for both at once.

## Environment

`main` at d897991.

thc1006:
DaemonSet Pods tolerate the unschedulable taint, so cordoning a node does not remove the DRA driver Pod. Its nodeSelector only requires the kernel-module-ready label, and that label stays on the node until the unloader succeeds, which it cannot do while the driver still holds the kernel module. So the Pod never leaves and the unloader keeps retrying.

#1274 fixed exactly this for device plugin by introducing a `device-plugin-target` node label, but the DRA reconciler never got the counterpart. This patch adds a `dra-target` node label maintained by the DRA reconciler, which now watches nodes, and the DRA DaemonSet nodeSelector requires it beside the kernel-module-ready label. Removing the label lets the DaemonSet controller delete the DRA Pod, and then the unloader converges on a later retry.

Fixes #1328
Refs #1332

## Why the ResourceClaim part is here

Whether the driver may stop is a different question from whether the node can accept new Pods. Cordoning leaves the running Pods where they are, and the DRA driver is what kubelet calls for `NodeUnprepareResources` on the devices those Pods hold. The upstream guidance is explicit about it:

> When draining a node, drain the DRA driver as late as possible [...] If you implement custom drain logic for nodes, consider checking that there are no allocated/reserved ResourceClaim or ResourceClaimTemplates before terminating the DRA driver itself.

So the label is also kept on any node where a ResourceClaim allocated from this Module's `spec.dra.driverName` is still reserved by a Pod, and this overrides both the selector and the schedulability. The reconciler watches ResourceClaims, so the last reservation going away is what releases the driver. This needs `resource.k8s.io/resourceclaims` with `list;watch`.

Without this part, the naive version of the fix, which is dropping the label as soon as the node is cordoned, is a smaller change but known to be unsafe.

## What it guarantees

A node carries `dra-target` while the Module selects it and it is schedulable, or while a ResourceClaim allocated from the Module's driver is still reserved by a consumer on it. The second overrides the first, which is what keeps a cordoned node's driver up until its consumers are done.

The rest is about not acting on something that has already moved:

- A reservation that cannot be placed on a node, because its Pod is gone and the allocation names no single node, holds every label the pass would otherwise drop. `allocatedNodeName` only names a node when every ORed selector term pins the same one.
- The Module is confirmed against the API server, uncached, before anything in the pass writes. UID, generation and deletion are all compared, since deletion leaves the generation alone. The node labels are then decided in full before any of them is written, so a pass that learns partway through that the Module has moved on leaves every node as it was rather than half moved to a spec that is gone. Nothing cleans up after one: the DRA resources have no finalizer of their own (#1331).
- Dropping a label is confirmed against a second uncached read of the reservations, and the patch itself takes an optimistic lock, so a node uncordoned between the read and the write gets a conflict rather than losing its driver.
- Adding the target selector to a DaemonSet that predates it rolls its Pods, so that migration waits while any claim is reserved or unplaceable, and `handleDRA` and garbage collection wait with it. Correcting a selector whose value is wrong does not wait: that value already selects no node, so it is what took the driver away, and it is patched with a JSON Patch guarded on the value this pass saw, so a key that has since gone or changed again is left to the next pass instead. Adding that key to an existing DaemonSet belongs to the claim-gated pass alone: ordinary convergence carries over whatever the live object holds rather than asserting it, and a pass whose classification is invalidated stops touching DaemonSets altogether, so neither a stale snapshot nor `CreateOrPatch` re-reading the live object can complete a held migration. A rejected patch counts as that guard only once a re-read shows the value really moved, since 422 also carries ordinary validation and the migration's own merge patch has no test operation at all.
- The migration takes its own uncached read rather than reusing the one the label removals used, since a reservation can appear between the two.
- A DaemonSet is looked for uncached before one is created. The list a pass starts from is cached and writing a node label queues this Module again, so a pass can run before the DaemonSet the last one created is visible, and a generated name would produce a second one for the same version rather than a conflict.
- Garbage collection waits for an old version's Pods to be gone, not only for its counts to say none are wanted. Once its selector stops matching, the Pod still on the node is counted in `numberMisscheduled` alone, while desired, current and ready all read zero.
- A held migration and a reservation that cannot be placed are reported as Warning events on the Module, so a drain that will not converge names the claim holding it rather than only showing up in the controller's log.
- A pass that cannot settle on an event asks to come back on a timer: an unplaceable reservation waits on a Pod binding, a node held only by a claim waits on a release event that a failed mapper list would swallow, a held migration waits on the claims, and a node a claim needs that cannot be read waits on the node itself.

## What it does not guarantee

The DRA DaemonSet selects on `kernel-module-ready` as well as `dra-target`, and this PR only owns the second one. `nmc_reconciler.go:112` drops the ready labels for any Module still in a cordoned node's NMC spec, with no claim check at all, so whether the driver survives a cordon comes down to which of `NMCReconciler` and `ModuleReconciler` reaches it first. #1328 describes both orderings; this fixes the one where the unloader runs with the driver still up. The other one needs a hold the NMC side respects, which is #1332. So this stops the DRA reconciler dropping the driver itself; it does not make the drain claim-safe on its own.

- A Module without a `spec.moduleLoader` has no kernel module to sequence against, so it keeps selecting on the Module selector and gets none of this.
- The target label keeps a node in the DaemonSet's selector, it does not make the Pod tolerate anything: an untolerated `NoExecute` taint still evicts the driver, a `NoExecute` toleration with a `tolerationSeconds` still expires, and an untolerated `NoSchedule` one still stops it coming back.
- The guard covers the automatic selector migration. Once the selector is present, an explicit Module update to the image, env, probes or security context still rolls the driver through `handleDRA`, so a driver that cannot be restarted under load needs seamless upgrade support.
- Garbage collection of an older version's DaemonSet waits for its Pods to be gone, not only for the count of wanted ones to reach zero, since deleting the DaemonSet takes whatever is left with it.
- Deleting the Module removes the label from every node without looking for claims. The DRA reconciler has no finalizer of its own, so cleanup can also be cut short: #1331.
- Uncached reads close the informer lag, not the whole race. Nothing puts a ResourceClaim and a Node in one transaction, so a reservation created after the read still lands after the check.
- `spec.dra.driverName` must identify one Module in the cluster. Retention is keyed by it, so two Modules sharing a name hold each other's labels and defer each other's migrations.
- The migration is held for the whole Module, not per node, because a DaemonSet's pod template is cluster-wide. One long-lived claim can therefore keep a legacy DaemonSet on the old selector while an unrelated node is drained.
- Upgrading to this changes the pod template of every DRA DaemonSet the Module already owns, not only the current version's, so each DRA Pod restarts once.

## Testing

`ci/prow/e2e-dra` runs the guard from both sides. First it repairs a missing target selector on the live DaemonSet before a consumer takes a claim, waits for that rollout to settle, and then keeps `dra-consumer` running across a cordon. Once the claim is reserved it checks that the DaemonSet generation does not move, and through the cordon window it asserts the label on `node/minikube` specifically alongside the reservation still naming the consumer, so a claim released early cannot make the label check pass for the wrong reason. The label going away afterwards is tied to the reservation clearing, not just to the consumer disappearing.

Then it does the same thing the other way round. Taking the selector off rolls the DaemonSet itself, so that happens with the controller scaled to zero and is left to settle before any claim is taken. The other order would replace the driver Pod under an active claim, which is the thing the window is meant to prove does not happen. A second consumer then takes a claim, the controller comes back to a migration already pending, and for 45 seconds the selector, the DaemonSet generation, the driver Pod's UID and readiness, and the reservation are all checked to hold. Releasing the claim is what lets the migration through, and that is a normal delete rather than a forced one, so the eviction only finishes if `NodeUnprepareResources` does.

`dra-example-driver` does not hold `kmm_ci_a`, so CI cannot reproduce an `rmmod` that really fails. Making it do so means giving `kmm_ci_a` something a process can hold open, and `ci/kmm-kmod-dockerfile.yaml` builds that module by cloning the default branch and is shared with `e2e-incluster-build` and `e2e-helm-incluster-build`, so it is a fixture change with its own blast radius rather than something to fold in here. As it stands the unload succeeds during the cordon, `removeOrphanedLabels` drops the ready label, and the driver Pod goes with it. That is why the cordon scenario force deletes its consumer rather than draining it, and why it covers the target label's lifecycle rather than a graceful `NodeUnprepareResources`. The migration scenario keeps the driver up throughout, so it deletes its consumer normally.

Unit tests cover selector narrowing, a node losing and regaining a selector label, a stale label on a node which is also cordoned, a converged cluster patching nothing, tolerated and untolerated taints, a node deleted in the middle, the in-use override in both directions, a cordon followed by an uncordon, the DaemonSet migration and its skip conditions, missing versus wrong selector values, a stale and a deleting Module, `spec.dra` immutability, Module deletion, and the new mapper and predicates.

Every line this adds to `dra_reconciler.go` is covered, and every fix has a test that fails without it, checked by mutating the fix and confirming which spec goes red. That is worth saying because more than one of them looked right and was not: swallowing a patch conflict is only safe when something is guaranteed to come back, and reading the node predicate is what showed nothing was.

Since everything here is mock based like the rest of the repo, I also ran the assumptions under it against a real API server with envtest, and did not commit that harness since there is no envtest suite here. It is what confirmed rather than argued the two worst findings: with a real ResourceClaim allocated to two ORed `metadata.name` terms, the old `allocatedNodeName` reports the first node and the reconcile takes the target label off the second one; and a plain merge patch removes a label from a node that was uncordoned after the read, where the optimistic one gets a conflict and leaves it alone. I checked there too that writing Module status leaves `.metadata.generation` alone, which is what the re-read depends on.

For `allocatedNodeName` itself I cross checked against the node affinity matcher in `k8s.io/component-helpers` over every selector buildable from a small pool of terms, asserting that whenever it names a node, no other node in the universe matches the selector. 584 selectors, no violation, and the previous version fails it.

## Not in this PR

Sequencing the unload against the driver Pod's exit is not here. I tried it by making `ModuleReconciler` hold the module in NMC desired state until the driver Pod is gone, and the e2e showed it makes `NMCReconciler` remove the kernel-module-ready label instead, since the module is then still in the spec of an unschedulable node, so the Pod is deleted for the opposite reason. It belongs to `NMCReconciler` where the ready label lives.

The device plugin's own target label hardening, adopting `module.EffectiveTolerations` in the existing call sites, extracting the shared reconciler after both controllers are settled, the generic `FindModulesForNode` fix and some e2e wait hardening are split into separate branches. #1330, #1331 and #1333 stay separated.

#1333 is worth naming because it meets this change head on. A reserved claim keeps the target label up while the node is cordoned, so the unloader keeps failing and retrying, and if the node is uncordoned in that window the label comes straight back, which is covered by a unit test here. What does not come back is `ProcessModuleSpec` cancelling the unloader still on the node, so it can succeed later and take the module off a node that is back in service. That half is NMC's and stays in #1333. #1331 is the other one to watch at release time: this adds more node labels to clean up and the DRA reconciler still has no finalizer of its own, so a Module can go away before the cleanup finishes.

<details>
<summary>What earlier review rounds changed</summary>

Kept for anyone following the history. Everything here is already in the code above.

- `reconcileDRATargetLabel` only visited nodes the selector returns or that already carry the label, so a node still in use, no longer selected and already stripped of its label was never looked at and could not get it back. Those are now fetched by name.
- Garbage collection read `desiredNumberScheduled == 0` from the DaemonSet list taken before the selector migration patched it, so an older DaemonSet could be repaired and deleted in the same reconcile. It now waits for `status.observedGeneration` to catch up.
- A reservation that cannot be placed on a node was read as proof the driver was free. `nodesUsingDRADriver` now resolves the node from the claim's allocation when the consumer cannot answer for one, and reports anything still unplaceable so the pass keeps every label it would otherwise drop.
- Dropping a label deletes the driver Pod, and no later event can call that termination back, so a removal is now confirmed against uncached reads rather than the informer: the Module the decision came from, then the reservations. Both happen once per pass and only when there is something to drop. That closes the informer lag, not the whole race. A reservation created after the read still lands after the check, and there is no transaction spanning a ResourceClaim and a Node.
- `spec.dra` cannot be removed and `spec.dra.driverName` cannot be changed. Retention is keyed by the driver name, and removing `spec.dra` took the cleanup path without any claim check, which is also how a rename would otherwise have got through. Adding `spec.dra` to a Module that never had one is still allowed, since there is no driver there to pull out from under a claim.
- `status.dra.desiredNumber` is now the number of nodes the reconciled target label leaves selected, rather than the raw selector count, so a node held for a reserved claim is counted. A node whose label patch comes back `NotFound` is left out of it, since it no longer exists.
- `allocatedNodeName` returned the first exact `metadata.name` requirement it found, but `NodeSelectorTerms` are ORed, so an allocation naming two nodes was read as naming the first. The second then looked unused and could lose its driver with the claim still reserved. It now answers only when every term pins the same name, and leaves the reservation unresolved otherwise.
- The removal patch carried no resourceVersion precondition, so a node uncordoned between the list and the patch still lost its label from stale state. Removals now take an optimistic lock and skip on conflict, leaving the decision to the reconcile that change queues. The Module is re-read for the same reason: a selector or toleration this pass never saw can make a node it wants to drop eligible again.
- The user and developer docs claimed the driver outlives its consumers and described removing `spec.dra`. They now say what the label controls, what it does not, and that admission rejects that removal.
- A conflict on the removal patch was treated as success on the grounds that the change causing it would come back as an event. It does not: the node predicate takes taints and labels, and a status update is neither. Conflicts now ask for a requeue and leave the node in the desired count until one lands.
- Nothing woke the reconciler when a reservation that could not be placed finally got its Pod scheduled, or when the only thing holding a node was a claim whose release event was lost. Both now requeue on a timer, since the state they wait on is not watched.
- Adding the target selector to a DaemonSet that predates it rolls its Pods, which takes the driver off the node long enough for a retrying unloader to win. That migration is now judged against an uncached read and held while any claim is reserved or unplaceable. `setDRAAsDesired` applies the same selector, so the rest of the DaemonSet reconciliation waits with it.
- `status.dra.nodesMatchingSelectorNumber` was still counted with the Module's raw tolerations while the desired number next to it used the effective ones, so a node under memory pressure could appear in one and not the other.
- Reviewing my own version of the above found two more. The requeue for a node held only by a claim keyed off the node not matching the selector, but the case it exists for is a node that still matches and is cordoned, so it never fired where it mattered, and the test I wrote for it had been shaped to the same mistake. And the requeue was being set on the result before the remaining steps could fail, which hands controller-runtime a result next to an error, something it ignores and warns about.
- CI then found the guard breaking the e2e: it removed the target selector from the live DaemonSet after the claim was reserved and waited two minutes for a repair the guard is there to withhold. The test was doing the thing the guard exists to prevent, so the drift check now runs before a consumer takes a claim, and what happens under an active claim is asserted instead: the DaemonSet generation does not move.
- I had also shared one uncached read between the label removals and the migration to save a call. That is wrong: a reservation can appear in between, and the migration would then roll the driver on the strength of what the removal saw. Each takes its own read again.
- Deletion does not bump `metadata.generation`, so a pass that started before it was still treated as current. It is compared separately now, and a Module that moved on stops the whole pass rather than only the DaemonSet work, since garbage collection reads the version off the object the pass started from. Deferred migration holds garbage collection back for the same reason.
- I had the node predicate watch `spec.unschedulable` as well, to see a cordon without waiting for the node lifecycle controller to turn it into a taint. `IsNodeSchedulable` only reads taints, so all that bought was an earlier reconcile that decides nothing. Making it count would mean teaching the shared helper about it, which the device plugin and NMC also use, so it is out again rather than left as a claim the code does not make good on.
- The e2e asserted that some node carried the target label. The cluster has two, so a stale label on the wrong one could hide a missing one on the right one.
- The freshness check was still conditional: it only ran when a label was about to come off or a DaemonSet needed the selector. A converged pass never reaches it, and a DRA-only Module never enters that pass at all, so an older generation could still write its DaemonSet template back or garbage-collect the version the current one wants. The Module is now confirmed against the API server before anything in the pass writes, and the checks immediately before the destructive steps stay where they are.
- A DaemonSet asking for the wrong target label value was treated like one missing it, so both waited for the claims. That one is the wrong way round: a wrong value already selects no node, which is what took the driver away, and the claim waiting on that driver cannot clear until it is back. Only a missing key waits now; a wrong value is corrected straight away, including on a pass that is holding a migration.
- Finding the Module has moved on was recorded but not acted on: the same branch still corrected wrong-value selectors before handing the result back, and the node loop could apply an addition before it got as far as looking. It is a write barrier now, and nothing after it touches a node or a DaemonSet.
- The correction itself was an unconditional merge patch, so a selector going from wrong to missing between the read and the write turned a recovery into the migration the claims are holding back. It is a JSON Patch now, with a `test` on the value this pass classified. I had it as a bare `replace` first, on the assumption that RFC 6902 makes that fail for a path that is not there; against a real API server it is accepted and simply adds the key, which is the bypass it was meant to close.
- Coverage of the lines this adds went from 13 uncovered to none, which is what turned up the untested paths above: a failed Module confirmation, a correction that no longer applies, a claims read that fails before a migration, and a DaemonSet already being deleted.
- A node kept only by a claim was looked up in the cache and forgotten on `NotFound`. Nothing else can put that label back: the node matches neither the selector nor an existing target label, so no node event maps to this Module. It is an uncached read now, and a node that really is gone leaves a requeue behind.

</details>

---

/cc @yevgeny-shnaidman @ybettan @TomerNewman

### DIFF
--- a/internal/controllers/dra_reconciler.go
+++ b/internal/controllers/dra_reconciler.go
@@ -20,10 +20,16 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"maps"
+	"slices"
+	"strings"
+	"sync"
+	"time"
 
 	kmmv1beta1 "github.com/kubernetes-sigs/kernel-module-management/api/v1beta1"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/constants"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/filter"
+	"github.com/kubernetes-sigs/kernel-module-management/internal/module"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/node"
 	"github.com/kubernetes-sigs/kernel-module-management/internal/utils"
 	appsv1 "k8s.io/api/apps/v1"
@@ -32,6 +38,9 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -57,22 +66,52 @@ const (
 
 type DRAReconciler struct {
 	client         client.Client
+	filter         *filter.Filter
 	reconHelperAPI draReconcilerHelperAPI
 }
 
 func NewDRAReconciler(
 	client client.Client,
+	apiReader client.Reader,
+	filter *filter.Filter,
 	nodeAPI node.Node,
+	recorder record.EventRecorder,
 	scheme *runtime.Scheme,
 ) *DRAReconciler {
-	reconHelperAPI := newDRAReconcilerHelper(client, nodeAPI, scheme)
+	reconHelperAPI := newDRAReconcilerHelper(client, apiReader, nodeAPI, recorder, scheme)
 	return &DRAReconciler{
 		client:         client,
+		filter:         filter,
 		reconHelperAPI: reconHelperAPI,
 	}
 }
 
+// draClaimDriverIndex keeps a pass looking for one driver from walking every claim in the cluster.
+const draClaimDriverIndex = "status.allocation.devices.results.driver"
+
+// resourceClaimDrivers names every driver an allocated ResourceClaim uses, and nothing for one the
+// scheduler has not allocated yet: an unallocated claim holds no driver on any node.
+func resourceClaimDrivers(rawObj client.Object) []string {
+	claim := rawObj.(*resourcev1.ResourceClaim)
+	if claim.Status.Allocation == nil {
+		return nil
+	}
+
+	drivers := make([]string, 0, len(claim.Status.Allocation.Devices.Results))
+	for _, result := range claim.Status.Allocation.Devices.Results {
+		drivers = append(drivers, result.Driver)
+	}
+
+	return drivers
+}
+
 func (r *DRAReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(
+		context.Background(), &resourcev1.ResourceClaim{}, draClaimDriverIndex, resourceClaimDrivers,
+	); err != nil {
+		return err
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&kmmv1beta1.Module{}).
 		Owns(&appsv1.DaemonSet{}).
@@ -80,6 +119,16 @@ func (r *DRAReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&resourcev1.DeviceClass{},
 			handler.EnqueueRequestsFromMapFunc(filter.DeviceClassToModuleReconcileRequest),
 			builder.WithPredicates(filter.HasLabel(constants.ModuleNameLabel)),
+		).
+		Watches(
+			&v1.Node{},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindDRAModulesForNode),
+			builder.WithPredicates(filter.DRAReconcilerNodePredicate()),
+		).
+		Watches(
+			&resourcev1.ResourceClaim{},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindDRAModulesForResourceClaim),
+			builder.WithPredicates(filter.ResourceClaimUsageChanged()),
 		).
 		Named(DRAReconcilerName).
 		Complete(
@@ -102,25 +151,72 @@ func (r *DRAReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Module) (
 		return res, fmt.Errorf("could not get DeviceClasses for module %s, namespace %s: %v", mod.Name, mod.Namespace, err)
 	}
 
+	// Everything past here writes, and a pass working from a spec the Module has moved on from can
+	// undo the current one.
+	current, err := r.reconHelperAPI.confirmCurrentModule(ctx, mod)
+	if err != nil {
+		return res, err
+	}
+
+	if !current {
+		logger.Info("Module changed before this pass wrote anything, leaving the rest to the next one")
+
+		return ctrl.Result{RequeueAfter: draTargetRequeue}, nil
+	}
+
+	// Labels go after the resources selecting on them, so a failed patch strands no DaemonSet.
 	if mod.GetDeletionTimestamp() != nil {
-		return ctrl.Result{}, r.reconHelperAPI.deleteDRAResources(ctx, mod.Name, mod.Namespace)
+		if err = r.reconHelperAPI.deleteDRAResources(ctx, mod.Name, mod.Namespace); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err = r.reconHelperAPI.removeDRATargetLabels(ctx, mod); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not remove dra-target labels on deletion: %v", err)
+		}
+		return ctrl.Result{}, nil
 	}
 
 	if mod.Spec.DRA == nil {
 		if err = r.reconHelperAPI.deleteDRAResources(ctx, mod.Name, mod.Namespace); err != nil {
 			return ctrl.Result{}, err
 		}
+		if err = r.reconHelperAPI.removeDRATargetLabels(ctx, mod); err != nil {
+			return ctrl.Result{}, fmt.Errorf("could not remove dra-target labels: %v", err)
+		}
 		return ctrl.Result{}, r.reconHelperAPI.clearDRAStatus(ctx, mod)
 	}
 
-	err = r.reconHelperAPI.handleDRA(ctx, mod, existingDRADS)
-	if err != nil {
-		return res, fmt.Errorf("could not handle DRA: %v", err)
+	// The label has to exist before the DaemonSet requires it, and there is only something to
+	// sequence against while a kernel module is loaded, so without a loader it is dropped instead.
+	targetResult := draTargetResult{}
+	if mod.Spec.ModuleLoader != nil {
+		if targetResult, err = r.reconHelperAPI.handleDRATargetLabels(ctx, mod, existingDRADS); err != nil {
+			return res, fmt.Errorf("could not reconcile dra-target labels: %v", err)
+		}
+	}
+	// The rest of this pass would write from a spec that is no longer the Module's, and deleting a
+	// DaemonSet the current spec wants back is not something a later pass can undo.
+	if targetResult.stale {
+		logger.Info("Module changed while its target labels were being reconciled, leaving the rest to the next one")
+
+		return ctrl.Result{RequeueAfter: draTargetRequeue}, nil
 	}
 
-	err = r.reconHelperAPI.garbageCollectDRADaemonSets(ctx, mod, existingDRADS)
-	if err != nil {
-		return res, fmt.Errorf("failed to run DRA garbage collection: %v", err)
+	// setDRAAsDesired applies the same selector, and garbage collection reads a replica count the
+	// held migration has not settled yet, so both wait with it.
+	if !targetResult.deferDaemonSets {
+		if err = r.reconHelperAPI.handleDRA(ctx, mod, existingDRADS); err != nil {
+			return res, fmt.Errorf("could not handle DRA: %v", err)
+		}
+
+		if err = r.reconHelperAPI.garbageCollectDRADaemonSets(ctx, mod, existingDRADS); err != nil {
+			return res, fmt.Errorf("failed to run DRA garbage collection: %v", err)
+		}
+	}
+
+	if mod.Spec.ModuleLoader == nil {
+		if err = r.reconHelperAPI.removeDRATargetLabels(ctx, mod); err != nil {
+			return res, fmt.Errorf("could not remove stale dra-target labels: %v", err)
+		}
 	}
 
 	err = r.reconHelperAPI.handleDeviceClasses(ctx, mod, existingDCs)
@@ -128,12 +224,15 @@ func (r *DRAReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Module) (
 		return res, fmt.Errorf("could not handle DeviceClasses: %v", err)
 	}
 
-	err = r.reconHelperAPI.moduleUpdateDRAStatus(ctx, mod, existingDRADS)
+	err = r.reconHelperAPI.moduleUpdateDRAStatus(ctx, mod, existingDRADS, targetResult.targeted)
 	if err != nil {
 		return res, fmt.Errorf("failed to update DRA status of the module: %v", err)
 	}
 
 	logger.Info("DRA reconcile loop finished successfully")
+
+	// Only here: controller-runtime ignores a result handed to it next to an error.
+	res.RequeueAfter = targetResult.requeueAfter
 
 	return res, nil
 }
@@ -143,33 +242,58 @@ func (r *DRAReconciler) Reconcile(ctx context.Context, mod *kmmv1beta1.Module) (
 type draReconcilerHelperAPI interface {
 	getModuleDRADaemonSets(ctx context.Context, name, namespace string) ([]appsv1.DaemonSet, error)
 	handleDRA(ctx context.Context, mod *kmmv1beta1.Module, existingDRADS []appsv1.DaemonSet) error
+	confirmCurrentModule(ctx context.Context, mod *kmmv1beta1.Module) (bool, error)
+	handleDRATargetLabels(ctx context.Context, mod *kmmv1beta1.Module, existingDRADS []appsv1.DaemonSet) (draTargetResult, error)
+	removeDRATargetLabels(ctx context.Context, mod *kmmv1beta1.Module) error
 	garbageCollectDRADaemonSets(ctx context.Context, mod *kmmv1beta1.Module, existingDS []appsv1.DaemonSet) error
 	deleteDRAResources(ctx context.Context, moduleName, moduleNamespace string) error
-	moduleUpdateDRAStatus(ctx context.Context, mod *kmmv1beta1.Module, existingDRADS []appsv1.DaemonSet) error
+	moduleUpdateDRAStatus(ctx context.Context, mod *kmmv1beta1.Module, existingDRADS []appsv1.DaemonSet, targetedNodes int) error
 	clearDRAStatus(ctx context.Context, mod *kmmv1beta1.Module) error
 	getModuleDeviceClasses(ctx context.Context, name, namespace string) ([]resourcev1.DeviceClass, error)
 	handleDeviceClasses(ctx context.Context, mod *kmmv1beta1.Module, existingDCs []resourcev1.DeviceClass) error
 }
 
 type draReconcilerHelper struct {
-	client          client.Client
+	client client.Client
+	// apiReader bypasses the cache: dropping a label deletes the driver Pod, and nothing calls the
+	// termination back, while adding one from a stale read is harmless.
+	apiReader       client.Reader
 	daemonSetHelper draDaemonSetCreator
 	nodeAPI         node.Node
+	recorder        record.EventRecorder
+}
+
+// event is a no-op when the helper was built without a recorder, which is how the unit tests build
+// it. A drain that does not converge is otherwise only visible in the controller's own logs.
+func (drh *draReconcilerHelper) event(mod *kmmv1beta1.Module, reason, format string, args ...any) {
+	if drh.recorder == nil {
+		return
+	}
+
+	drh.recorder.Eventf(mod, v1.EventTypeWarning, reason, format, args...)
 }
 
 func newDRAReconcilerHelper(client client.Client,
+	apiReader client.Reader,
 	nodeAPI node.Node,
+	recorder record.EventRecorder,
 	scheme *runtime.Scheme,
 ) draReconcilerHelperAPI {
 	daemonSetHelper := newDRADaemonSetCreator(scheme)
 	return &draReconcilerHelper{
 		client:          client,
+		apiReader:       apiReader,
 		daemonSetHelper: daemonSetHelper,
 		nodeAPI:         nodeAPI,
+		recorder:        recorder,
 	}
 }
 
 func (drh *draReconcilerHelper) getModuleDRADaemonSets(ctx context.Context, name, namespace string) ([]appsv1.DaemonSet, error) {
+	return listModuleDRADaemonSets(ctx, drh.client, name, namespace)
+}
+
+func listModuleDRADaemonSets(ctx context.Context, reader client.Reader, name, namespace string) ([]appsv1.DaemonSet, error) {
 	dsList := appsv1.DaemonSetList{}
 	opts := []client.ListOption{
 		client.MatchingLabels(map[string]string{
@@ -178,7 +302,7 @@ func (drh *draReconcilerHelper) getModuleDRADaemonSets(ctx context.Context, name
 		}),
 		client.InNamespace(namespace),
 	}
-	if err := drh.client.List(ctx, &dsList, opts...); err != nil {
+	if err := reader.List(ctx, &dsList, opts...); err != nil {
 		return nil, fmt.Errorf("could not list DaemonSets: %v", err)
 	}
 
@@ -193,15 +317,51 @@ func (drh *draReconcilerHelper) handleDRA(ctx context.Context, mod *kmmv1beta1.M
 	logger := log.FromContext(ctx)
 
 	ds, version := getExistingDRADSFromVersion(existingDRADS, mod.Namespace, mod.Name, mod.Spec.ModuleLoader)
+
+	// The list this pass started from is cached, and writing a node label queues the Module again,
+	// so a pass can run before the DaemonSet the last one created is visible. Creating one then
+	// succeeds, because a generated name never conflicts, and leaves two for the same version.
 	if ds == nil {
+		fresh, err := listModuleDRADaemonSets(ctx, drh.apiReader, mod.Name, mod.Namespace)
+		if err != nil {
+			return err
+		}
+
+		ds, version = getExistingDRADSFromVersion(fresh, mod.Namespace, mod.Name, mod.Spec.ModuleLoader)
+	}
+
+	creating := ds == nil
+	if creating {
 		logger.Info("creating new DRA DaemonSet", "version", version)
 		ds = &appsv1.DaemonSet{
 			ObjectMeta: metav1.ObjectMeta{Namespace: mod.Namespace, GenerateName: mod.Name + "-dra-"},
 		}
 	}
 
+	targetLabel := utils.GetDRATargetNodeLabel(mod.Namespace, mod.Name)
+
 	opRes, err := controllerutil.CreateOrPatch(ctx, drh.client, ds, func() error {
-		return drh.daemonSetHelper.setDRAAsDesired(ctx, ds, mod)
+		// Requiring this label narrows what an existing DaemonSet selects, which is the migration
+		// the claims hold back, so only the gated pass may start it. Carry over what the live
+		// object has instead of asserting it, and a stale classification cannot migrate behind it.
+		held, wasSet := ds.Spec.Template.Spec.NodeSelector[targetLabel]
+
+		if err := drh.daemonSetHelper.setDRAAsDesired(ctx, ds, mod); err != nil {
+			return err
+		}
+
+		// Without a loader the node selector is the Module's own map, which must not be edited.
+		if creating || mod.Spec.ModuleLoader == nil {
+			return nil
+		}
+
+		if wasSet {
+			ds.Spec.Template.Spec.NodeSelector[targetLabel] = held
+		} else {
+			delete(ds.Spec.Template.Spec.NodeSelector, targetLabel)
+		}
+
+		return nil
 	})
 
 	if err == nil {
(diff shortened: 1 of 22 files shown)


thc1006:
The e2e caught something I had backwards, so I have dropped one of the two changes I pushed earlier.

I had `ModuleReconciler` hold the module in the NMC desired state until the DRA driver Pod was gone, on the theory that the unloader would otherwise race a driver that is still terminating. What the run showed was the opposite: the target label survived the cordon exactly as intended, but the driver Pod was deleted anyway.

```
++ kubectl get nodes -l kmm.node.kubernetes.io/default.kmm-ci-dra.dra-target -o name
+ out=node/minikube
++ kubectl get pods -l kmm.node.kubernetes.io/role=dra,... -o name
+ out=
```

Keeping the module in the spec of an unschedulable node is the condition `NMCReconciler` uses to take the kernel-module-ready label off, and the DRA DaemonSet selects on both labels. The barrier was therefore removing the Pod it was meant to protect. Without it the module leaves the desired state at cordon time, the ready label is held by the orphaned-status path until the unloader succeeds, and the unloader keeps failing on `EBUSY` for exactly as long as the driver holds the module — which is the behaviour the label change was after in the first place.

The claim-aware retention stays, since that part the e2e confirmed works. A driver Pod slow to terminate can still lose a race with the unloader; that belongs in `NMCReconciler`, where the ready label actually lives, and I would rather propose it separately than smuggle a second mechanism into this PR.

`e2e-hub` and `e2e-helm-hub` in that run had no failing step — they were cancelled when `e2e-dra` failed.

thc1006:
Another round, and the e2e was right about something I had wrong.

I had the test asserting that the target label, the driver Pod **and** the kernel module all survive the cordon while the claim is held. Only the first of those is something this PR controls. `dra-example-driver` does not hold `kmm_ci_a`, so in CI the unloader succeeds while the node is cordoned, `removeOrphanedLabels` drops the ready label once the status entry goes, and the DaemonSet loses its Pod for that reason rather than this one. The run said so plainly:

```
++ kubectl get nodes -l kmm.node.kubernetes.io/default.kmm-ci-dra.dra-target -o name
+ out=node/minikube          <- the label survived, which is the new behaviour
++ kubectl get pods -l kmm.node.kubernetes.io/role=dra,... -o name
+ out=                       <- the Pod did not, for an unrelated reason
```

So the test now checks the label's own lifecycle: present after the cordon while the claim is reserved, gone once the consumer is drained off. Those two states either side of the drain isolate the claim as the cause, which is the thing worth locking down; the Pod's fate additionally depends on a driver that genuinely blocks `rmmod`, and CI has no way to produce one. It also strips the target selector off the live DaemonSet with `kubectl patch` and waits for the reconciler to put it back, which is what a DaemonSet created by an older KMM looks like.

## What the maintainer-review pass changed

Reading back through the inline comments on #1265 through #1322, three things in here were going to get flagged and have been fixed. `DRAReconcilerNodePredicate` and `DevicePluginReconcilerNodePredicate` were two names for one behaviour, so they are now a single `PluginReconcilerNodePredicate` — `Plugin` being the name this repo already uses for what the two have in common. `mod.Spec.ModuleLoader != nil` was being checked in `Reconcile` and again in `handleDRA`, so the DaemonSet-selector step moved into `handleDRATargetLabels`, which is only reached under that condition; both sides of the target label are now reconciled in one place, in the order they have to happen in. And `nodesUsingDRADriver` guarded against an empty `driverName`, which `internal/webhook/module.go:175` already rejects, so that guard is gone.

The RBAC rule is `list;watch` rather than `get;list;watch`: nothing reads a ResourceClaim by name.

## What envtest settled

Since everything here is mock-based, I ran the assumptions underneath it against a real API server. Two of the answers changed the PR.

The first is a correction to something I wrote in the description. I had claimed the desired-state guard is what stops the node watch feeding itself. It is not: an unchanged `MergeFrom` patch serialises to `{}`, is still sent, and comes back without a `resourceVersion` bump, so no event follows and the loop would settle either way. What the guard actually saves is one PATCH per node per reconcile in steady state.

The second is a cost I had not disclosed. Patching `spec.template.spec.nodeSelector` on a DaemonSet takes `.metadata.generation` from 1 to 2, so upgrading to this restarts every DRA and device plugin Pod once. Unavoidable if the selector is to change at all, but it belongs in the description rather than in someone's cluster unannounced.

It also confirmed the parts I was only assuming: `HasLabels` finds a node whose label value is wrong where a value-based list finds none, and the API server stores a claim consumer as `APIGroup ""`, `Resource "pods"` and the Pod's UID, which is what `nodesUsingDRADriver` matches on. I have not committed the harness — there is no envtest suite in this repo and adding one does not belong here.

## Gaps I am not closing here

The device plugin has no e2e coverage anywhere in this repo, so its ordering, tolerations and DaemonSet-selector changes rest on unit tests alone. `e2e-dra` deploys with kustomize, so the Helm role change is not exercised either. Both are real work rather than oversights; happy to take either on separately if you would rather they came first.

Under `ginkgo --randomize-all` four specs in `internal/controllers` fail, and the same four fail on `main` — `prepareSchedulingData` and friends share state across specs. None of the specs added here are order-dependent. CI does not randomise, so nothing is blocked on it, but it seemed worth saying rather than leaving as a surprise.

thc1006:
/hold

This got too large to be reviewed as a single bug fix, so I cut it back to the DRA drain lifecycle only. It is 20 files and +1889/-16 now, it was 27 files and +2277/-132.

The claim-aware retention stays here. Dropping the driver right after the node is cordoned takes it away while kubelet still needs it for `NodeUnprepareResources`.

~~so this PR also closes #1332~~ Correcting that: it does not. This PR only owns `dra-target`. `NMCReconciler` can still drop `kernel-module-ready` first, with no claim check, and that path takes the driver away regardless of what this PR does. #1332 stays open for it, and the description says the same.

The device plugin changes, the toleration cleanup, the shared target-label refactor, the generic `FindModulesForNode` fix and some e2e wait hardening are moved out into separate branches. #1330, #1331 and #1333 stay as they are. No test is deleted, they moved together with the behavior they cover.

Description is updated.

TomerNewman:
Thanks for the work here, and for splitting the findings into separate issues (#1330, #1331, #1333), makes this much easier to review as a focused fix.
I can see it's still on hold, please ping us once you're ready and we will take a look.

TomerNewman:
@thc1006 I see its still on hold, do you think this PR will be merged until 12.9?

thc1006:
Off hold now.

Since you last looked I pushed six fixes from another review pass. The regular DaemonSet path could complete a migration the claims were holding, because `CreateOrPatch` re-reads the live object and rewrites the whole spec; adding the target selector to an existing DaemonSet is now the claim-gated pass's alone. A DaemonSet is looked for uncached before one is created, so a cache miss cannot leave two for the same version. The node-label pass decides the whole batch before it writes any of it, so it writes nothing once it finds the Module has moved on. Garbage collection waits for `numberMisscheduled` as well, since that is where a Pod left on a node the selector no longer matches is counted. A 422 counts as the guarded patch losing its race only after a re-read shows the value moved. The migration e2e now settles the legacy DaemonSet before a claim is taken, so its own setup no longer rolls the driver under the claim it is meant to protect.

Whether it lands by Saturday is down to review time rather than anything on my side. Checks are green and I can turn changes around quickly.

internal/controllers/dra_reconciler.go
```diff
@@ -81,6 +95,16 @@ func (r *DRAReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			handler.EnqueueRequestsFromMapFunc(filter.DeviceClassToModuleReconcileRequest),
 			builder.WithPredicates(filter.HasLabel(constants.ModuleNameLabel)),
 		).
+		Watches(
+			&v1.Node{},
+			handler.EnqueueRequestsFromMapFunc(r.filter.FindDRAModulesForNode),
+			builder.WithPredicates(filter.DRAReconcilerNodePredicate()),
+		).
+		Watches(
+			&resourcev1.ResourceClaim{},
```
TomerNewman: Assuming we run on k8s < v1.34, will it work?

internal/controllers/dra_reconciler.go
```diff
@@ -211,6 +346,623 @@ func (drh *draReconcilerHelper) handleDRA(ctx context.Context, mod *kmmv1beta1.M
...
+func (drh *draReconcilerHelper) nodesUsingDRADriver(
+	ctx context.Context,
+	reader client.Reader,
+	mod *kmmv1beta1.Module,
+) (driverUsage, error) {
+	logger := log.FromContext(ctx)
+	usage := driverUsage{nodes: sets.New[string]()}
+
+	claims := resourcev1.ResourceClaimList{}
+	if err := reader.List(ctx, &claims); err != nil {
```
TomerNewman: In this scenario for each reconcile - it lists every resourceclaim in the cluster and then filters with the driver name, right?
It could be inefficient if we will have a lot of resourceclaims.
I suggest to index the resourceclaim with the driver name and then use client.MatchingFields for the cached list

TomerNewman (commented):
Also, since this is a big PR, wdyt about splitting it to multiple small prs?

thc1006:
Agreed, and it grew back since I cut it down in August. I will split it.

It is +4105 now, 986 of that being the change itself. I already have the ResourceClaim index you asked for, so it goes in the first piece with `nodesUsingDRADriver`. The rest splits by what each part owns: the target label reconcile, the DaemonSet selector migration, and the webhook and e2e.

On the index itself, the API server has no field selector for the driver, so `MatchingFields` only works on the cached read. The uncached recheck still lists everything, otherwise it fails with a BadRequest.

thc1006:
The index is pushed. One thing to note about it: the API server has no field selector for the driver, so `MatchingFields` only narrows the cached read. The uncached recheck still lists everything, otherwise it comes back as a BadRequest.

On splitting, I tried it and did not get far. I would rather show you what stopped me than guess at it again.

The webhook guard is the one piece that comes out cleanly as a file. On main though, nothing outside the webhook reads `spec.dra.driverName`, so on its own it asks to make a field immutable with nothing that breaks today. The claim retention cannot come out: `reconcileDRATargetLabel` decides `wantLabel := byClaim || bySelector`, and a version without that drops the driver on cordon while kubelet still needs it for `NodeUnprepareResources`. I had most hope for the DaemonSet selector migration, since its production side is only 15 lines of `handleDRATargetLabels`, but removing it also removes `deferDaemonSets`, which is what makes `handleDRA` and the garbage collection wait for the migration to settle.

So what is left is one behaviour and the tests for it. 986 of the 4159 lines are the change itself.

There is one more way to make the first PR small, if you would like it. Add the label and the DaemonSet selector first and never remove it, then removal and claim retention on top. Every step is safe on its own, and the cost is that you read `reconcileDRATargetLabel` twice. You also asked here on the 7th whether this would land by 12.9, so I would rather you choose than have me pick for you.

TomerNewman:
@thc1006 I appreciate you digging into this! :)
wdyt about doing this split though?
1. Webhook immutability: just the spec.dra/driverName immutability check. Small, standalone, easy yes.
2. Label groundwork: the dra-target label add logic (selector + schedulability only, no claims, no removal) and the DaemonSet selector change, but only applied to newly-created DaemonSets. No behavior change for anyone upgrading, since nothing removes the label yet.
3. The actual fix: claims (watch/index/retention), removal, and the DaemonSet migration for existing DaemonSets, all together with tests, since that's the part that can't be safely separated.

thc1006:
Yes, let's do that. Splitting the webhook off first is better than where I had it.

Piece 2 is safe the way you describe it. `setDRAAsDesired` only leaves the target label in the selector when it's creating the DaemonSet; for one that already exists it carries over whatever the live object had, so nothing migrates until piece 3 brings that in.

One correction to what I said on the 8th: the ResourceClaim index belongs in piece 3, not piece 1. Nothing calls `nodesUsingDRADriver` until the claims are there, and lint fails on a function with no caller.

1 and 2 don't touch the same files, so I can put both up at once if that suits you better than one after the other. Starting on them now.

PR #1334, opened 2026-08-17
Add optional modprobedDir to ModprobeSpec

NataliShemtov:
Add an optional modprobedDir field so Modules can point at modprobe.d files in the driver image.
Regenerate CRDs and add unit tests.

### DIFF
--- a/api/v1beta1/module_types.go
+++ b/api/v1beta1/module_types.go
@@ -174,6 +174,13 @@ type ModprobeSpec struct {
 	// +optional
 	FirmwarePath string `json:"firmwarePath,omitempty"`
 
+	// ModprobedDir is an absolute path in the driver container image that contains
+	// modprobe.d configuration files (flat directory only; no subdirectories).
+	// When set, KMM copies those files into /etc/modprobe.d/ in the worker pod
+	// before modprobe load/unload.
+	// +optional
+	ModprobedDir string `json:"modprobedDir,omitempty"`
+
 	// ModulesLoadingOrder defines the dependency between kernel modules loading, in case
 	// it was not created by depmod (independent kernel modules).
 	// The list order should be: upmost module, then the module it depends on and so on.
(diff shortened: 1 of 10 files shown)


api/v1beta1/module_types_test.go
```diff
@@ -0,0 +1,53 @@
...
+package v1beta1
+
+import (
+	"encoding/json"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("ModprobeSpec", func() {
```
TomerNewman: Those tests here are basically test the constructor, not sure how much value it gives since there isn't any logic to test (unlike KMM's internal functions/controllers/webhook/etc...)

NataliShemtov: Dropped it.

api/v1beta1/module_types.go
```diff
@@ -174,6 +174,14 @@ type ModprobeSpec struct {
 	// +optional
 	FirmwarePath string `json:"firmwarePath,omitempty"`
 
+	// ModprobedDir is an absolute path in the driver container image that contains
+	// modprobe.d configuration files (regular files only; no subdirectories).
+	// When set, KMM copies those files into /etc/modprobe.d/ in the worker pod
+	// before modprobe load/unload. Mutually exclusive with ModulesLoadingOrder.
```
TomerNewman: another thing, since you've not introduced yet mutually exclusivness with `ModulesLoadingOrder` field, I would omit that from this commit for now.

api/v1beta1/module_types.go
```diff
@@ -174,6 +174,14 @@ type ModprobeSpec struct {
...
+	// Setting this field causes worker pods to run Privileged.
```
TomerNewman: same goes for here

api/v1beta1/module_types.go
NataliShemtov: Ok - will update it later

api/v1beta1/module_types.go
NataliShemtov: same here

api/v1beta1/module_types.go
```diff
@@ -174,6 +174,13 @@ type ModprobeSpec struct {
 	// +optional
 	FirmwarePath string `json:"firmwarePath,omitempty"`
 
+	// ModprobedDir is an absolute path in the driver container image that contains
+	// modprobe.d configuration files (regular files only; no subdirectories).
```
ybettan:
```suggestion
	// modprobe.d configuration files (flat directory only; no subdirectories).
```

PR #1335, opened 2026-08-19
Switch to server-side apply for kubectl install/deploy commands

TomerNewman:
Client-side apply stores the full manifest in the last-applied-configuration annotation, capped at 256KB.
The Module and ManagedClusterModule CRDs haven't hit that limit yet, but the pending configurable SecurityContext change (#1323) will push them over it and break kubectl apply.

Switch to --server-side --force-conflicts preemptively, since it tracks field ownership on the API server instead and isn't subject to this limit.

---

/cc @ybettan

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ KUSTOMIZE_CONFIG_CRD ?= config/crd
 
 .PHONY: install
 install: manifests ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	kubectl apply -k $(KUSTOMIZE_CONFIG_CRD)
+	kubectl apply --server-side --force-conflicts -k $(KUSTOMIZE_CONFIG_CRD)
 
 .PHONY: uninstall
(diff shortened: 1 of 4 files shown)


yevgeny-shnaidman:
@TomerNewman the optional fields (if not set) are not added to the annotations, IIRC.

TomerNewman:
> @TomerNewman the optional fields (if not set) are not added to the annotations, IIRC.

The error is from the CRD object itself 
`The CustomResourceDefinition "modules.kmm.sigs.x-k8s.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes`

yevgeny-shnaidman:
> > @TomerNewman the optional fields (if not set) are not added to the annotations, IIRC.
> 
> The error is from the CRD object itself `The CustomResourceDefinition "modules.kmm.sigs.x-k8s.io" is invalid: metadata.annotations: Too long: may not be more than 262144 bytes`

so we see the error during CRD deployment?

TomerNewman:
> so we see the error during CRD deployment?

yes

PR #1336, opened 2026-08-19
Validate modprobedDir in Module admission webhook

NataliShemtov:
Reject non-absolute or malformed modprobedDir in the webhook. 

A Module may set `modprobedDir` and `modulesLoadingOrder` together.
That combination is not rejected at admission.

### DIFF
--- a/internal/webhook/module.go
+++ b/internal/webhook/module.go
@@ -327,7 +327,24 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
 		}
 	}
 
+	if modprobe.ModprobedDir != "" {
+		if !isWellFormedAbsolutePath(modprobe.ModprobedDir) {
+			return fmt.Errorf("modprobedDir %q must be a well-formed absolute path", modprobe.ModprobedDir)
+		}
+	}
+
 	return nil
+}
+
+func isWellFormedAbsolutePath(p string) bool {
+	if !filepath.IsAbs(p) {
+		return false
+	}
+	trimmed := strings.TrimSuffix(p, "/")
+	if trimmed == "" {
+		trimmed = "/"
+	}
+	return filepath.Clean(trimmed) == trimmed
 }
 
 func validateSignSection(sign *kmmv1beta1.Sign, dirName string) error {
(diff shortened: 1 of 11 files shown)


internal/webhook/module.go
```diff
@@ -327,9 +327,44 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
 		}
 	}
 
+	if modprobe.ModprobedDir != "" {
+		if !isWellFormedAbsolutePath(modprobe.ModprobedDir) {
+			return fmt.Errorf("modprobedDir %q must be a well-formed absolute path", modprobe.ModprobedDir)
+		}
+
+		if len(modprobe.ModulesLoadingOrder) > 0 {
```
TomerNewman: instead of setting the mutual exclusive check here, you can set in modules_type.go as a kubebuilder annotation, we already do it with DRA and Deviceplugin.

NataliShemtov: Alright - Moved mutual exclusion to an XValidation on ModprobeSpec.

internal/webhook/module.go
```diff
@@ -327,9 +327,36 @@ func validateModprobe(modprobe kmmv1beta1.ModprobeSpec) error {
...
+	if modprobe.ModprobedDir != "" {
+		if !isWellFormedAbsolutePath(modprobe.ModprobedDir) {
+			return fmt.Errorf("modprobedDir %q must be a well-formed absolute path", modprobe.ModprobedDir)
+		}
+	}
+
 	return nil
 }
 
+func isWellFormedAbsolutePath(p string) bool {
```
TomerNewman, line 339: IMO we can reuse stdlib instead of manually looping
```golang
func isWellFormedAbsolutePath(p string) bool {
	if !filepath.IsAbs(p) {
		return false
	}
	trimmed := strings.TrimSuffix(p, "/")
	if trimmed == "" {
		trimmed = "/"
	}
	return filepath.Clean(trimmed) == trimmed
} 
```

NataliShemtov: Thanks, applied it.

PR #1339, opened 2026-08-24
Apply worker.runAsUser to privileged firmware loader pods

### LINKED ISSUE #1337: Firmware loading fails on containerd-based clusters due to ignored `runAsUser`
### Summary

KMM is failing to load firmware on vanilla Kubernetes (k8s) clusters because it doesn't apply `runAsUser` to the firmware loader pod, giving the container engine control over which effective capabilities the resulting container gets. CRI-O and containerd differ in how they handle this implementation detail, making this bug reproducible on vanilla k8s but not on OpenShift (OCP).

Linked Jira card: [MGMT-25099](https://redhat.atlassian.net/browse/MGMT-25099)

### Expected behavior

KMM should be able to load firmware for KMDs that ship it when `moduleLoader.container.modprobe.firmwarePath` is set on vanilla K8s.

### Actual behavior

KMM fails to load firmware on vanilla K8s unless the image is rebuilt with `USER 0`.

### Steps to reproduce

1. Deploy KMM with the [default worker image](https://github.com/kubernetes-sigs/kernel-module-management/blob/ed592f28ce204baa836e6a0be77af60f699e244a/Dockerfile.worker) (or any image that sets a non-root `USER`) on containerd.
2. Create a `Module` whose `moduleLoader.container.modprobe.firmwarePath` is set (i.e., the driver ships firmware).
3. Observe the loader worker fail to write to `/sys/module/firmware_class/parameters/path` (permission denied), leaving the kernel unable to find required firmware files.
4. Observe that the failure persists even when `worker.runAsUser: 0` is set in the KMM config.
5. Observe that the issue is fixed after rebuilding the worker image with `USER 0`.

### Additional details

When a `Module` uses firmware loading (i.e., sets `moduleLoader.container.modprobe.firmwarePath`), KMM does not apply `worker.runAsUser` to the privileged loader pod's `SecurityContext`. The worker therefore runs as the worker image's baked-in user. On images where that user is non-root (e.g., the default [Dockerfile.worker](https://github.com/kubernetes-sigs/kernel-module-management/blob/ed592f28ce204baa836e6a0be77af60f699e244a/Dockerfile.worker#L44)) and the container engine doesn't grant the privileged container `CAP_DAC_OVERRIDE` in its effective set (e.g., containerd), the worker cannot write to the root-owned sysfs file `/sys/module/firmware_class/parameters/path`, so firmware loading fails.

`worker.runAsUser` currently only affects the non-privileged containers created by KMM, so setting this field to `0` does not help our privileged firmware loader containers. The only workaround is to rebuild the worker image with `USER 0`.

OpenShift doesn't encounter this bug for 2 reasons:
 - CRI-O (OCP's container engine) gives privileged non-root containers a full effective capability set by default. containerd (the container engine used by some other k8s distros) gives those same containers empty effective capability sets.
 - When installed using OLM, KMM on OCP uses a worker image (`registry.redhat.io/kmm/kernel-module-management-worker-rhel9:2.7.0`) that [doesn't set a `USER`](https://github.com/chr15p/rh-kmm-konflux2/blob/main/release-2.7/Dockerfile.worker) and therefore runs as root (i.e., it'd still work even if CRI-O worked more like containerd)

Note: this is distinct from #737 (read-only `/sys`). In our case, `/sys` is mounted `rw`, but the write  still fails because of the empty effective capability set.

### Proposed fix

In [`setWorkerSecurityContext`](https://github.com/kubernetes-sigs/kernel-module-management/blob/ed592f28ce204baa836e6a0be77af60f699e244a/internal/pod/workerpodmanager.go#L579-L589), apply `RunAsUser` regardless of the privileged flag, like so:

```go
sc := &v1.SecurityContext{}

if privileged {
    sc.Privileged = &privileged
} else {
    sc.Capabilities = &v1.Capabilities{Add: []v1.Capability{"SYS_MODULE"}}
    sc.SELinuxOptions = &v1.SELinuxOptions{Type: workerCfg.SELinuxType}
}

sc.RunAsUser = workerCfg.RunAsUser   // apply on both paths
```

With `worker.runAsUser` defaulting to `0`, the firmware write then succeeds out of the box. Alternatively, we could explicitly run the loader as root whenever `moduleLoader.container.modprobe.firmwarePath` is set.

TomerNewman:
On containerd, a privileged non-root container gets an empty effective capability set, so the firmware loader (which runs privileged) could not write to the root-owned /sys/module/firmware_class/parameters/path, even with worker.runAsUser set to 0.
 CRI-O masked this since it grants privileged non-root containers a full effective capability set.
 
 ---
 
 /cc @yevgeny-shnaidman @abyrne55
 fix #1337

### DIFF
--- a/internal/pod/workerpodmanager.go
+++ b/internal/pod/workerpodmanager.go
@@ -584,9 +584,10 @@ func setWorkerSecurityContext(pod *v1.Pod, workerCfg *config.Worker, privileged 
 		sc.Capabilities = &v1.Capabilities{
 			Add: []v1.Capability{"SYS_MODULE"},
 		}
-		sc.RunAsUser = workerCfg.RunAsUser
 		sc.SELinuxOptions = &v1.SELinuxOptions{Type: workerCfg.SELinuxType}
 	}
+
+	sc.RunAsUser = workerCfg.RunAsUser
 
 	container.SecurityContext = sc
 
--- a/internal/pod/workerpodmanager_test.go
+++ b/internal/pod/workerpodmanager_test.go
@@ -201,6 +201,7 @@ var _ = Describe("CreateLoaderPod", func() {
 			if withFirmwareLoading && firmwareHostPath != nil {
 				container.SecurityContext = &v1.SecurityContext{
 					Privileged: ptr.To(true),
+					RunAsUser:  workerCfg.RunAsUser,
 				}
 			} else {
 				container.SecurityContext = &v1.SecurityContext{


internal/pod/workerpodmanager.go
```diff
@@ -584,10 +584,11 @@ func setWorkerSecurityContext(pod *v1.Pod, workerCfg *config.Worker, privileged
 		sc.Capabilities = &v1.Capabilities{
 			Add: []v1.Capability{"SYS_MODULE"},
 		}
-		sc.RunAsUser = workerCfg.RunAsUser
 		sc.SELinuxOptions = &v1.SELinuxOptions{Type: workerCfg.SELinuxType}
 	}
 
+	sc.RunAsUser = workerCfg.RunAsUser
```
ybettan, line 590: What happens if we need firmware and the user has updated the config to `runAsUser: 100` for example?

TomerNewman: On containerd it should fail then

PR #1340, opened 2026-08-26
Update PRD:Allow modprobedDir together with modulesLoadingOrder

NataliShemtov:
Supersedes the mutual-exclusion decision in #1319. 
The combination must not be rejected.
If a user file collides with KMM's loading-order file, the module fails to load instead of overwriting silently.

### DIFF
--- a/docs/enhancements/0005-modprobed-config.md
+++ b/docs/enhancements/0005-modprobed-config.md
@@ -28,10 +28,9 @@ modules that rely on this pattern cannot be reliably unloaded through KMM.
   Firmware loading capability, with no degradation to either.
 - Users are always informed when their modprobe.d configuration could not
   be applied, rather than experiencing a silent failure.
-- The modprobe.d capability introduces no degradation to the existing
-  `ModulesLoadingOrder` capability — achieved by preventing the two from
-  being enabled together on the same module (see Non-Goals), rather than
-  allowing them to silently conflict.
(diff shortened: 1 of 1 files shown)


docs/enhancements/0005-modprobed-config.md
```diff
@@ -49,14 +48,21 @@ modules that rely on this pattern cannot be reliably unloaded through KMM.
...
-  in the worker Pod, which collides with this capability's use of the
-  same path. Merging both into a single input (for example, generating
-  and mounting the combined configuration from an init container) is a
-  long-term solution that is out of scope for this enhancement. Users who
-  need both softdep-style module ordering and other modprobe.d directives
-  can express the ordering directly in their own modprobe.d files.
+
+### 2.3 Approach
+
+User modprobe.d files are copied from the driver image into
```
yevgeny-shnaidman, line 54: somewhere we should suggest to the user that the modprobe.d file be named as <kernel module name>.conf
This is not a hard requirement from the  modprobe, but it is a very useful convention

NataliShemtov: Yes, from what ive seen [here](https://manpages.debian.org/bookworm/kmod/modprobe.d.5.en.html) modprobe only reads *.conf files.
I’ll add a note that user files should use that suffix.

PR #1342, opened 2026-09-07
Fix DRA and device plugin cleanup finalization

### LINKED ISSUE #1331: DRA cleanup can be skipped entirely because no finalizer holds the Module
DRA cleanup runs from the DRA reconciler, but nothing keeps the Module around long enough for it to finish. I noticed this while working on #1329, which adds a node label to the set of things that need cleaning up, so I am filing it separately.

The only finalizer on a Module is `constants.ModuleFinalizer`, added and removed by the Module reconciler. `finalizeModule` waits for the NMC in-use labels to go away and then removes it, without knowing anything about DRA. Once it is gone the API server deletes the object, and `reconcile.AsReconciler` gets NotFound the next time the DRA reconciler tries, so the deletion branch never runs.

That leaves behind:

- DeviceClasses, which are cluster-scoped. A namespaced Module cannot be a valid owner of a cluster-scoped object, so ownerReferences cannot cover this and the only cleanup path is the reconciler's own `DeleteAllOf`.
- `dra-target` node labels, once #1329 lands.
- Any device-plugin-target labels, which have the same shape today.

It is most likely for a Module with `spec.dra` but no `spec.moduleLoader`: there is no NMC usage to wait for, so the Module finalizer can be removed almost immediately.

A recreated Module with the same namespace and name then finds pre-labeled nodes, so a driver Pod can be scheduled before the new generation has evaluated whether those nodes are schedulable at all.

A DRA-specific finalizer, for example `kmm.sigs.x-k8s.io/dra-cleanup`, would let the reconciler delete the DaemonSets, wait for the Pods to be gone, delete the DeviceClasses, remove the target labels and clear the status before letting the Module go. Adding the Module UID to the managed resources would also stop a new generation from inheriting the previous one's leftovers.

Happy to work on this if it sounds right.

thc1006:
This patch gives the DRA and device plugin reconcilers a finalizer of their own on the Module, so that a Module is not deleted before either of them has finished cleaning up.

The two plugins have the same gap and the fix is the same shape for both, so they are here together.

### Why the cleanup can be skipped today

`finalizeModule` waits on the NMC in-use labels and knows nothing about either plugin. A Module with a plugin spec and no `moduleLoader` has no NMC at all, so both of its lists come back empty and `ModuleFinalizer` is removed on the first pass after the deletion starts.

`reconcile.AsReconciler` ends its Get with `client.IgnoreNotFound(err)`, so once the object is gone the typed `Reconcile` is never called again. Returning an error to be retried does not help, because there is nothing left to retry against.

A DeviceClass is cluster scoped, so a namespaced Module cannot be a valid owner and ownerReferences do not cover it either. Nothing else would collect it.

### The shared finalizer list

`setFinalizerAndStatus` and `finalizeModule` both patch `metadata.finalizers` with `client.MergeFrom`. A merge patch sends that list whole. Computed from a read that predates the plugin finalizer, the removal sends:

```
{"metadata":{"finalizers":null}}
```

which drops every finalizer on the object, not only its own. Both writers now take `client.MergeFromWithOptimisticLock{}` and fail on a stale read instead. This is not optional for the rest of the patch to work.

The Pod node label reconciler removes its own finalizer from the plugin Pods the same way, and the Module now waits behind those Pods, so that write takes the lock too. Its test asserted the payload `{"metadata":{"finalizers":null}}`, which is right while nothing else is holding the Pod and wrong the moment something is.

### Registering the finalizer

Registration goes in before anything the finalizer protects is written, and a Module that has gone between the read and that write is not read as one that no longer needs it. Nothing would be left to run the cleanup, and the pass would carry on to write a DeviceClass, or leave target labels on nodes, with no Module to collect them. Removal still treats `NotFound` as done, since a Module that is gone has nothing left to release.

### Reporting cleanup as done

The API server accepting a DELETE is not the object having gone, and a cache that has not caught up would report an empty cluster either way, so the cleanup reads through `mgr.GetAPIReader()` and reports `done` only when it finds nothing left. A Pod counts while it is still an object, whatever its phase, and only when a DaemonSet owns it, so worker Pods are not waited for. Deletes carry the UID and resourceVersion that were read, so a replacement under the same name is left alone.

Only objects that are not already going are asked for. A repeat foreground delete puts `foregroundDeletion` back after the collector took it off, and both of those writes come back through the DeviceClass watch as another pass: on a held DeviceClass that ran the cleanup ten times a second, and the version-locked patch that releases the hold could never land between two of them. Measured on kind, the same wait now logs one pass every 2.014s rather than every 0.1s. The count still covers everything that was listed, so an object that is still finalizing keeps the Module: counting deletes sent instead would release it early.

Two existing behaviours are kept: DaemonSets and Pods written before the role label existed are still matched, per the `TODO(3.0)` in `getModuleDevicePluginDaemonSets`, and node target labels are now found by key rather than by value, so a node whose label was edited is not missed.

The patch that takes a target label off names the key rather than being diffed off the node. `Labels` is `omitempty`, so a node left with none serialises without the field and the diff becomes `{"metadata":{"labels":null}}`, which would take that node's other labels with it. A node always carries the kubelet's own labels, so this is a property of the helper rather than something reachable on a real node, but what it sends should not rest on that.

### The deadlock this would otherwise create

Waiting for a plugin Pod means waiting for `NodeLabelerFinalizer` to come off it. `deleteLabel` in the Pod node label reconciler treated a missing Node as an error, so a Pod whose node had gone would keep that finalizer for good, and now the Module behind it as well. A node that is not there carries no label, so that case returns nil.

### Not in this patch

Module UID labels on the managed resources. `standardLabels` is used for both `ds.Spec.Selector.MatchLabels` and the pod template, and a DaemonSet selector is immutable once created, so that needs its own migration story.

### Docs

A Module that stays in `Terminating` is now something an operator can be asked about, so `troubleshooting.md` says which finalizer waits for what, why a DaemonSet can still be listed while it is on its way out, and that the log line each pass writes names the DaemonSets, Pods and DeviceClasses it can still see.

### Testing

Unit tests cover both reconcilers symmetrically, and every fix has a test that fails without it, checked by reverting the fix and confirming which spec goes red.

I also ran it on kind with Kubernetes v1.35.0, against this branch and against master for comparison. With a device-plugin Module that has no `moduleLoader`, and its Pod held by a test finalizer: `ModuleFinalizer` is gone within three seconds on both, but master deletes the Module in the same window and leaves the Pod behind, while this branch keeps the Module on `device-plugin-cleanup` until the hold is released, and then everything goes. The same with a DRA Module and its DeviceClass held: the DaemonSet and Pods are already collected, the DeviceClass sits in Terminating, and the Module waits on `dra-cleanup` for it.

That second sequence is now in `ci/prow/e2e-dra`, since a Module with no `moduleLoader` is the case where the generic finalizer leaves early and there is nothing else to fall back on. It waits for the DaemonSets and Pods to have gone and for `dra-cleanup` to be the only finalizer left before it starts watching, so that the held DeviceClass is the only thing the Module can still be waiting for, and it pins the DeviceClass UID so that one which goes early or comes back under the same name fails rather than reads as the one being held.

That ordering is doing work. Against a build where the DeviceClass no longer counts toward the cleanup's remaining work, with one DRA Pod held by a test finalizer, the version without the wait passes: the held Pod keeps the Module there for the whole interval on its own, and the check never reaches the DeviceClass. The version here fails, because the Module never gets to the state the check is about. Against master it fails at the first wait, on a finalizer that is never added.

The device plugin half is in `ci/prow/e2e-incluster-build`, which needs neither DRA nor a second node. It holds one of the plugin's Pods rather than a DeviceClass, and that is the case the DRA test cannot reach: the DaemonSet is deleted with foreground propagation, so it stays until that Pod has gone, and the Pod cannot go until `NodeLabelerFinalizer` comes off it. Waiting for the DaemonSet to disappear would therefore never come true here, so the wait is that everything has been asked to go and only the held Pod cannot finish.

The test's own finalizer goes on beside `NodeLabelerFinalizer`, never over it, and the wait is for the Pod to be left holding only the test's entry: what the test watches is KMM releasing its own. `metadata.finalizers` is a list and a merge patch sends a list whole, so both scripts go through `ci/prow/finalizer-test-lib.sh`, which reads the list back and writes it with only the test's entry added or taken out, under the resourceVersion it read. Neither the release nor the trap clears the list, so a Pod that KMM is still holding stays held rather than being tidied away by the test. Against a build that never removes `NodeLabelerFinalizer`, the test fails. It starts from a Pod that is ready and a node carrying both of the labels KMM sets for it, so that neither going missing later can read as cleanup when it was never set. Those two answer to different things: the ready label is the Pod finalizer's, and has to be off the node before the Pod is left holding only the test's entry, while the target label is the Module cleanup's and is checked at the end. On master it fails at the first wait, on a finalizer that is never added.

Fixes #1331

---

### DIFF
--- a/ci/module-kmm-ci-device-plugin-only.yaml
+++ b/ci/module-kmm-ci-device-plugin-only.yaml
@@ -1,0 +1,11 @@
+---
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: Module
+metadata:
+  name: kmm-ci-device-plugin-only
+spec:
+  devicePlugin:
+    container:
+      image: registry.k8s.io/pause:3.10
+  selector:
+    kubernetes.io/hostname: minikube
--- a/ci/module-kmm-ci-dra-only.yaml
+++ b/ci/module-kmm-ci-dra-only.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: kmm.sigs.x-k8s.io/v1beta1
+kind: Module
+metadata:
+  name: kmm-ci-dra-only
+spec:
+  dra:
+    driverName: cleanup.example.com
+    serviceAccountName: dra-example-driver
+    container:
+      image: registry.k8s.io/dra-example-driver/dra-example-driver:v0.3.0
+      command: ["dra-example-kubeletplugin"]
+      env:
+        - name: DRIVER_NAME
+          value: cleanup.example.com
+    deviceClasses:
+      - name: kmm-ci-cleanup-device
+  selector:
+    kubernetes.io/hostname: minikube
--- a/ci/prow/e2e-dra
+++ b/ci/prow/e2e-dra
@@ -78,3 +78,100 @@ timeout 1m bash -c 'until ! minikube ssh -- lsmod | grep kmm_ci_a; do sleep 3; d
 
 echo "Wait for the Module to be deleted"
 kubectl wait --for delete modules.kmm.sigs.x-k8s.io/kmm-ci-dra
+
+# Without a moduleLoader there is no NMC, so ModuleFinalizer goes on the first pass and only the
+# DRA one holds the Module. A DeviceClass is cluster scoped, so nothing else would collect it.
+source "$(dirname "$0")/finalizer-test-lib.sh"
+
+echo "Deploy a Module with a DRA spec and no moduleLoader"
+kubectl apply -f ci/module-kmm-ci-dra-only.yaml
+
+echo "Wait for the DRA finalizer and the DeviceClass"
+timeout 2m bash -c 'until kubectl get module kmm-ci-dra-only -o json | jq -e ".metadata.finalizers | index(\"kmm.node.kubernetes.io/dra-cleanup\")" > /dev/null; do sleep 3; done'
+timeout 2m bash -c 'until kubectl get deviceclass kmm-ci-cleanup-device; do sleep 3; done'
+
+# Beside anything already there, never over it. The trap lets it go if anything below fails.
+echo "Hold the DeviceClass so its deletion cannot complete"
+hold_with_test_finalizer deviceclass/kmm-ci-cleanup-device
+# Pinned, so a DeviceClass that goes and comes back under the same name fails rather than passes.
+DC_UID=$(kubectl get deviceclass kmm-ci-cleanup-device -o jsonpath='{.metadata.uid}')
+trap 'release_test_finalizer deviceclass/kmm-ci-cleanup-device || true' EXIT
+
+kubectl delete module kmm-ci-dra-only --wait=false
+
+# Everything else has to be gone first, or it would hold the Module by itself and this would pass
+# even with the DeviceClass wait broken. Nothing holds these Pods, so they do go.
+echo "Wait until the held DeviceClass is the only thing the Module waits for"
+DRA_DEADLINE=$(( SECONDS + 120 ))
+while true; do
+  if ! DRA_MODULE_JSON=$(kubectl get module kmm-ci-dra-only --ignore-not-found -o json); then
+    echo "could not read the Module"
+    exit 1
+  fi
+  if [ -z "${DRA_MODULE_JSON}" ]; then
+    echo "the Module was released before the DeviceClass was let go"
+    exit 1
+  fi
+
+  if ! DRA_DC_JSON=$(kubectl get deviceclass kmm-ci-cleanup-device --ignore-not-found -o json); then
+    echo "could not read the DeviceClass"
+    exit 1
+  fi
+  if [ -z "${DRA_DC_JSON}" ]; then
+    echo "the held DeviceClass went while the Module was still here"
+    exit 1
+  fi
+
+  # Read before the test, so that a failed query is a failure rather than an empty cluster.
+  if ! DRA_DS=$(kubectl get daemonset -l kmm.node.kubernetes.io/module.name=kmm-ci-dra-only -o name); then
+    echo "could not list the DRA DaemonSets"
+    exit 1
+  fi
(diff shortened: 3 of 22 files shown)


TomerNewman (commented):
Same goes for here, I think it would be the best if you split this PR as well into multiple smaller ones.

https://github.com/kubernetes-sigs/kernel-module-management/pull/1329#pullrequestreview-5138690317

thc1006:
You are right that it is big. It is +1845, but 396 of that is the change itself, the rest is 1107 lines of tests and mocks and 342 of ci and docs.

I can split it. The smallest pieces I get are the two finalizer-write fixes in `module_reconciler.go` and `pod_node_label_reconciler.go` (+149), the shared cleanup helper (+648), the DRA side (+348), the device plugin side (+356), and the e2e and docs (+342). They have to land in that order though. The optimistic lock is what stops the generic finalizer write from dropping the plugin ones, so it is a prerequisite rather than a separate cleanup, and lint rejects the shared helper until one of the two reconcilers uses it.

That is four rounds of review, and you asked for this by 12.9. Would you rather I split it now and let the date move, or land this one and I open the split as follow-ups? Your call on the date, I am fine either way.

TomerNewman:
@thc1006 Thanks for laying that out so clearly!
This one is with the highest priority for us, so let's go with the split you suggested- we'll do our best to turn reviews around quickly so it doesn't drag the timeline out too much.

thc1006:
Closing this. The split we agreed on is up as #1343 for the two finalizer-write fixes and #1344 for the DRA side, and the device plugin side and the e2e and docs follow as each of those lands.

Leaving #1331 open. It also lists the `dra-target` labels, and those arrive with #1329, so the issue outlives these pieces. The `Fixes #1331` on this PR was wrong.

PR #1343, opened 2026-09-08
Fix the finalizer writes on Modules and their plugin Pods

thc1006:
This is the first of the smaller pieces of #1342, which @TomerNewman asked me to split up. It carries the two finalizer-write fixes the rest of that work rests on, and nothing else.

## What goes wrong

`metadata.finalizers` is a list, and a merge patch sends a list whole. `setFinalizerAndStatus` and `finalizeModule` both compute their patch from the Module they were handed, so a write from a read that predates another controller's entry takes that entry off along with its own. `deleteFinalizer` in the Pod node label reconciler has the same shape when it removes `NodeLabelerFinalizer` from a plugin Pod.

Nothing in tree puts a second finalizer on those objects today, which is why this has not been hit. The DRA and device plugin finalizers in #1342 do, and they are the ones that would go.

## What this changes

The three writes now carry the resourceVersion they read, so a stale one comes back as a conflict and the next pass starts from a fresh read.

`deleteLabel` also reported a Node that had already gone as an error. `NodeLabelerFinalizer` then stayed on the Pod for good, and so did anything waiting for that Pod to go. A Node that is not there carries no label, so that reads as done.

## Testing

Unit tests cover each of the three writes and the missing Node. I checked every one of them by reverting the fix and confirming which spec goes red.

Part of #1331.

---


internal/controllers/module_reconciler.go
```diff
@@ -228,9 +228,11 @@ func (mrh *moduleReconcilerHelper) setFinalizerAndStatus(ctx context.Context, mo
 	logger := log.FromContext(ctx)
 	logger.Info("Adding finalizer", "module name", mod.Name, "module namespace", mod.Namespace)
 
+	// The plugins hold their own finalizers on this object, and a merge patch sends the whole
```
yevgeny-shnaidman:
```suggestion
	//  both DRA and DevicePlugin can also set finalizers, on Module, so we switch to client.MergeFromWithOptimisticLock{} in order to avoid overriding previous operations
```

yevgeny-shnaidman:
@thc1006 just one (nit) very small comment, and then we can push

internal/controllers/module_reconciler.go
thc1006: Applied, over both lines of the old comment since the second one had no subject left on its own. I also took out the extra space after `//` to match the rest of the tree.

PR #1344, opened 2026-09-08
Give the DRA reconciler a finalizer of its own

thc1006:
Second of the smaller pieces of #1342. It gives the DRA reconciler a finalizer of its own, and leaves the device plugin side, the e2e and the docs to the pieces after it.

This one builds on #1343 and should land after it. Without the version lock that #1343 puts on the two writers of `ModuleFinalizer`, a write from a stale read takes the finalizer added here off with it, and the cleanup below never runs.

## Why the cleanup can be skipped today

`finalizeModule` waits on the NMC in-use labels and knows nothing about DRA. A Module with a DRA spec and no `moduleLoader` has no NMC at all, so both of its lists come back empty and `ModuleFinalizer` is removed on the first pass after the deletion starts.

`reconcile.AsReconciler` ends its Get with `client.IgnoreNotFound(err)`, so once the object is gone the typed `Reconcile` is never called again. Returning an error to be retried does not help, because there is nothing left to retry against.

A DeviceClass is cluster scoped, so a namespaced Module cannot be a valid owner and ownerReferences do not cover it either. Nothing else would collect it.

## Registering it

Registration goes in before anything the finalizer protects is written, and a Module that has gone between the read and that write is not read as one that no longer needs it. Nothing would be left to run the cleanup, and the pass would carry on to write a DeviceClass with no Module to collect it. Removal still treats `NotFound` as done, since a Module that is gone has nothing left to release.

## Reporting cleanup as done

The API server accepting a DELETE is not the object having gone, and a cache that has not caught up would report an empty cluster either way, so the cleanup reads through `mgr.GetAPIReader()` and reports `done` only when it finds nothing left. Deletes carry the UID and resourceVersion that were read, so a replacement under the same name is left alone.

An object that is already going is counted rather than asked for again. A repeat foreground delete puts `foregroundDeletion` back after the collector has taken it off, and both of those writes come back through the DeviceClass watch as another pass. On kind that ran the cleanup ten times a second for as long as a DeviceClass was held; with the guard it is one pass every two seconds, which is the requeue the code asks for.

A Pod counts while it is still an object, whatever its phase, and only when a DaemonSet owns it, so worker Pods are not waited for.

A pass reports every failure it hit, not the last one. The DeviceClass read comes after the DaemonSets have been asked to go, and returning only its error would drop a delete that had already failed.

## Creating the DaemonSet only once

Registering the finalizer is a write to the Module, so the pass that does it queues another one. That next pass can reach `handleDRA` before the DaemonSet informer has seen what this one created, and the cached list it works from would still be empty. `CreateOrPatch` looks the object up by name, so an empty list means a second `GenerateName` and a second DaemonSet for the same version. Garbage collection only takes DaemonSets whose version label differs, so nothing collects it.

`handleDRA` now asks the API server before it decides to generate a name. Both reads go through the same selector, so a mistake in one of them cannot make the other look at something else.

## About the shared file

`plugin_cleanup.go` holds the two helpers this uses. The device plugin side in the next piece uses the same two, plus a third for its node labels, which is why they are not inlined here.

## Testing

Unit tests cover the registration and release paths, both callers of the cleanup, the terminating-object case, and the scope of each list the cleanup makes. That last one matters because a DeviceClass is cluster scoped, so the namespace label on it is all that keeps this Module away from a same-named one somewhere else. Every fix has a test that fails without it, checked by reverting the fix and confirming which spec goes red.

I also ran it on kind with Kubernetes v1.35.0. With a DRA Module that has no `moduleLoader` and its DeviceClass held by a test finalizer, `ModuleFinalizer` is off the Module about a tenth of a second after the delete, leaving `dra-cleanup` as the only one left; the Module stays there until the hold is released, and then everything goes. On main the Module is deleted in the same window and the DeviceClass is left behind.

Part of #1331.

---

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -181,7 +181,7 @@ func main() {
 	}
 
 	if kubeVersion.AtLeast(constants.MinKubeMajorForDRA, constants.MinKubeMinorForDRA) {
-		if err = controllers.NewDRAReconciler(client, nodeAPI, scheme).SetupWithManager(mgr); err != nil {
+		if err = controllers.NewDRAReconciler(client, mgr.GetAPIReader(), nodeAPI, scheme).SetupWithManager(mgr); err != nil {
 			cmd.FatalError(setupLogger, err, "unable to create controller", "name", controllers.DRAReconcilerName)
 		}
 	} else {
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -19,6 +19,10 @@ const (
 	ModuleFinalizer   = "kmm.node.kubernetes.io/module-finalizer"
 	JobEventFinalizer = "kmm.node.kubernetes.io/job-event-finalizer"
 
+	// The DRA reconciler holds the Module itself, because ModuleFinalizer is released as soon as no
+	// NMC lists the Module as in use, which says nothing about whether the cleanup has finished.
+	DRAFinalizer = "kmm.node.kubernetes.io/dra-cleanup"
+
 	ManagedClusterModuleNameLabel  = "kmm.node.kubernetes.io/managedclustermodule.name"
 	KernelVersionsClusterClaimName = "kernel-versions.kmm.node.kubernetes.io"
 	DockerfileCMKey                = "dockerfile"
--- a/internal/controllers/dra_reconciler.go
+++ b/internal/controllers/dra_reconciler.go
@@ -62,10 +62,11 @@ type DRAReconciler struct {
 
 func NewDRAReconciler(
 	client client.Client,
+	apiReader client.Reader,
 	nodeAPI node.Node,
 	scheme *runtime.Scheme,
 ) *DRAReconciler {
-	reconHelperAPI := newDRAReconcilerHelper(client, nodeAPI, scheme)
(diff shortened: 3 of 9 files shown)


thc1006:
Pushed a test-only change. One table entry said it covered a Pod owned by something other than a DaemonSet, but it built a Pod with no owner at all, so `owner.Kind != "DaemonSet"` was never exercised; it is two entries now. I also added a check that the DeviceClass list is selected by both Module labels rather than by namespace, since that one is cluster scoped. No production change.

thc1006:
The e2e run on this head did not reach the tests. `e2e-helm` failed in "Download container images" with `Unable to download artifact(s): Failed to ListArtifacts ... (403) Forbidden`, and `e2e-dra` was then cancelled while creating minikube, so its Run test step was skipped rather than failing. The same workflow passed a couple of hours later on #1347, so it looks like a one-off on the artifact service.

I don't have permission to rerun it. Could you restart run 34254544495 when you get a chance?

thc1006:
Rebased onto main, which picks up #1347 and the SecurityContext change and gives this a run against the base it would actually merge into. The rerun of 34254544495 I asked about earlier is not needed now; that run was against a base two commits behind, so it would not have told us much either way.

Nothing else changed here.

PR #1345, opened 2026-09-08
Stage modprobedDir in worker pods and set Privileged

NataliShemtov:
When `modprobedDir` is set, stage regular files from the driver image into the worker pod's shared tmp volume and run loader and unloader pods Privileged.

The init container fails if the path is missing, not a directory, empty, or contains a subdirectory.
When `modulesLoadingOrder` is set, KMM always mounts `softdep.conf` as a file (`subPath`) so `/etc/modprobe.d` stays writable.


TomerNewman:
> When modprobedDir is set, stage regular files from the driver image into the worker pod's shared tmp volume and run loader and unloader pods Privileged.

but modeprobe searches under /etc/modprobe.d/*.conf no? so why it is being mounted to /tmp?

internal/pod/workerpodmanager.go
```diff
@@ -510,6 +526,10 @@ func setWorkerSofdepConfig(pod *v1.Pod, modulesLoadingOrder []string) error {
 		ReadOnly:  true,
 		MountPath: "/etc/modprobe.d",
 	}
+	if fileMount {
```
yevgeny-shnaidman: Why do we need to change it only in case of modprobe config defined? Why not make it mounted with subPath always? That way we can remove the "if" clause and we have no more need for fileMount input parameter

NataliShemtov: You're right - simpler this way.
I dropped fileMount. setWorkerSofdepConfig always mounts softdep.conf at /etc/modprobe.d/softdep.conf.

For Modules that only set modulesLoadingOrder, /etc/modprobe.d is no longer a read-only overlay, so worker-image conf files stay visible and the directory stays writable for later modprobedDir copies.

Enhancement (PRD) updated to match it.

internal/pod/workerpodmanager.go
```diff
@@ -536,6 +552,50 @@ cp -R %s %s;
...
+  exit 1
+fi
+mkdir -p "$dst" || exit 1
+for p in "$src"/*; do
+  if [ -f "$p" ]; then
+    cp "$p" "$dst/" || exit 1
+  fi
+done
+`
+	copyCommand := fmt.Sprintf(template, strconv.Quote(modprobedDir), strconv.Quote(dst))
```
TomerNewman: Lets say the path contains something like `/opt/driver/$(touch /tmp/marker)`, will execute `touch /tmp/marker` or treats it as a plain text?

NataliShemtov: Fixed it, good catch.
With double quotes, $() would still run in /bin/sh -c. 
I now wrap the path in single quotes so it is treated as plain text.

internal/pod/workerpodmanager.go
```diff
@@ -536,6 +551,64 @@ cp -R %s %s;
...
 }
 
+func addModprobedDirCopyCommand(pod *v1.Pod, modprobedDir string) error {
+	container, _ := podcmd.FindContainerByName(pod, initContainerName)
+	if container == nil {
+		return errors.New("could not find the init container")
+	}
+
+	dst := filepath.Join(sharedFilesDir, modprobedDir)
+	const template = `
```
yevgeny-shnaidman, line 561: since it is const, lets pick one of the options:
1. move it to the the variable at the beginning of the file
or
2. set it in a different file under scripts and use "//go:embed". You can see the example under internal/mcfg/scripts and in internal/mcfg/mcfg.go

NataliShemtov: Moved the script to internal/pod/scripts/copy-modprobed-dir.sh and load it with //go:embed.

TomerNewman (approved):

PR #1353, opened 2026-09-16
Apply staged modprobedDir files before modprobe

NataliShemtov:
Copy regular files from shared tmp into /etc/modprobe.d/ before load and unload.
Fail if modulesLoadingOrder is set and the user staged softdep.conf.


internal/worker/worker.go
```diff
@@ -105,6 +107,10 @@ func (w *worker) LoadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, fir
 		args = append(args, cfg.Modprobe.Parameters...)
 	}
 
+	if err := w.applyModprobedDir(cfg); err != nil {
+		return err
```
TomerNewman:
```suggestion
		return fmt.Errorf("failed to apply modprobe.d files: %w", err)
```

internal/worker/worker.go
```diff
@@ -155,6 +161,10 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, f
 		}
 	}
 
+	if err := w.applyModprobedDir(cfg); err != nil {
+		return err
```
TomerNewman:
```suggestion
		return fmt.Errorf("failed to apply modprobe.d files: %w", err)
```

internal/worker/worker.go
NataliShemtov: Done.

internal/worker/worker.go
NataliShemtov: Done.

internal/worker/worker.go
```diff
@@ -37,6 +37,8 @@ func NewWorker(mr ModprobeRunner, fh utils.FSHelper, logger logr.Logger) Worker
 
 const sharedFilesDir = "/tmp"
 
+var workerModprobeDDir = "/etc/modprobe.d"
```
TomerNewman, line 40: btw why not const?

NataliShemtov: Tests override this to a temp dir so we don't write into /etc/modprobe.d.
(Same pattern as firmwareClassPathLocation).

internal/worker/worker.go
```diff
@@ -172,3 +182,51 @@ func (w *worker) UnloadKmod(ctx context.Context, cfg *kmmv1beta1.ModuleConfig, f
...
+		return nil
+	}
+
+	src := filepath.Join(sharedFilesDir, cfg.Modprobe.ModprobedDir)
+	entries, err := os.ReadDir(src)
+	if err != nil {
+		return fmt.Errorf("failed to read staged modprobedDir %s: %v", src, err)
+	}
+
+	var files []os.DirEntry
```
yevgeny-shnaidman, line 197: IIRC we have already verified(somewhere) that this directory does not contain any subdirectories, no?

NataliShemtov: Yes. The init script already rejects subdirectories before copying into /tmp, so the worker should not see any in the normal pod path.
I can drop this check if you prefer.

NataliShemtov: Dropped the extra check here.

PR #1358, opened 2026-09-22
Emit Module Warning Events when a worker pod fails

NataliShemtov:
When a loader or unloader pod reaches Failed, record ModuleLoadFailed or ModuleUnloadFailed on the Module, including the terminated container message when present. Keep deleting the pod.


internal/controllers/nmc_reconciler.go
```diff
@@ -602,6 +603,40 @@ func (h *nmcReconcilerHelperImpl) SyncStatus(ctx context.Context, nmcObj *kmmv1b
 	return errors.Join(errs...)
 }
 
+func (h *nmcReconcilerHelperImpl) recordWorkerPodFailure(ctx context.Context, p *v1.Pod) {
+	logger := ctrl.LoggerFrom(ctx)
+
+	mod := &kmmv1beta1.Module{}
+	nsn := types.NamespacedName{Namespace: p.Namespace, Name: p.Labels[constants.ModuleNameLabel]}
+	if err := h.client.Get(ctx, nsn, mod); err != nil {
```
TomerNewman: Why not propagating Module to the func? for every failture now it does a API call

NataliShemtov: you are right - that's an extra API call. Passed the Module in and dropped it.

